PR #30, opened 2024-10-25
Updating GreenHEART Examples

kbrunik:
# Clean up GreenHEART Examples

- Added a README to the examples folder. Addresses #28.
- Removed unsupported/outdated example.
- Fixed import in reference design.
- Incorporated plot from @elenya-grant for PEM model example.


johnjasa (approved):
Looks great, I pushed some extremely minor changes directly to the PR.

All four of the example notebooks have some quasi-repetitive explanations in the first half of the notebooks. Is this on purpose for repetition, or is each notebook supposed to stand on its own? In the future for docs I'd suggest removing redundancies if not needed, but that's a later consideration.

kbrunik:
> Looks great, I pushed some extremely minor changes directly to the PR.
> 
> All four of the example notebooks have some quasi-repetitive explanations in the first half of the notebooks. Is this on purpose for repetition, or is each notebook supposed to stand on its own? In the future for docs I'd suggest removing redundancies if not needed, but that's a later consideration.

Yeah there's definitely some repetition, would love to have a large discussion about how the examples should work i.e. if they should be stand-alone or not. I feel 50/50 on it myself.

PR #31, opened 2024-10-25
Add a Conda `environment.yml` for easy setup

RHammond2:
# Simply the conda setup

This PR adds an `environment.yml` conda specification file to make setting up and working with GreenHEART easier for first time users. The primary issue this resolves is the additional package installations that must happen prior to installing GreenHEART itself.

## Related issue

N/A

- `README.md`: Updates the installation instructions
- `environment.yml`: Adds the conda environment specification file

N/A

N/A


johnjasa:
The changes to the readme make it seem like you installing via the conda environment doesn't require you to set the NREL API keys. I say this because the first header has three steps and the next header has eight and the two installation options seem mutually exclusive.

Let me know if I'm misinterpreting, but maybe clarifying you can install either way and then still need to set API keys or anything else necessary would be helpful.

johnjasa (approved):

PR #72, opened 2025-01-17
Add initial new GreenHEART (ODIES) documentation

johnjasa:
# Add initial new GreenHEART (ODIES) documentation

This PR adds the first bit of documentation focused on the development of new new GreenHEART (ODIES).
Rather than having this info off in my sandbox repo, I moved the initial docs to here so development can more easily happen in one place.

## PR Checklist

- [x] Documentation

## Related issues

N/A

N/A

N/A

N/A


.github/workflows/ci.yml
```diff
@@ -5,22 +5,16 @@ on: [ push, pull_request ]
 jobs:
   build:
     runs-on: ubuntu-latest
-    defaults:
```
RHammond2, line 8: Any reason why we're abandoning the environment settings? I don't have a strong preference either way, but was curious.

docs/_toc.yml
```diff
@@ -11,6 +11,13 @@ parts:
   - file: markdown-notebooks
   - file: install
   - file: CONTRIBUTING
+- caption: New GreenHEART
+  chapters:
+  - file: why_make_new_greenheart
+  - file: features_not_currently_implemented
```
RHammond2, line 17: Love that your voice is even included in the file naming!

docs/class_structure.md
```diff
@@ -0,0 +1,23 @@
+# Class structure in ODIES
+
+A major focus of ODIES is modularizing the components and system architecture so it's easier to construct and analyze complex hybrid power plants producing commodities for a variety of uses.
```
RHammond2, line 3: I know line length isn't enforced in GreenHEART, but it would be nice if we avoided very long lines to be kinder to our future selves.

docs/describing_whats_in_classes.md
```diff
@@ -0,0 +1,36 @@
...
+Specifically, this system must output `CapEx` and `OpEx` values for the technology.
+These values are later used in financial modeled and cost breakdowns.
+
+## Financial model (optional)
+
+Each technology class can define `get_financial_model()`, which returns an OpenMDAO system containing the financial model.
+This would override any plant level financial model, and is useful for technologies that have unique financial considerations.
+This will also be more relevant as we develop non-single-owner capabilities.
+
+## Control model (optional, unused)
```
RHammond2, line 27: A general comment is to watch out for defining undefined functionality that can come back to haunt us.

docs/development_guidelines.md
```diff
@@ -0,0 +1,45 @@
...
+Additionally, to help users create subsystems with the correct methods and attributes, we should have tests that users can run to check their technology models.
+
+First, we will define different types of tests that we have in ODIES:
+- Unit tests: tests for individual functions or methods
+- Integration tests: tests for the interaction between different functions or methods
+- Regression tests: tests for ensuring that new changes do not break existing functionality
+
+### Test location
+
+Each folder within the `ODIES` package should have a corresponding `tests` folder.
```
RHammond2, line 23: General comment that may not apply here: Should the unit, integration, and regression get separate folders within `tests/` that then roughly or directly mimic the code structure within? This then allows for easily setting up a workflow for running the each individually, such as `pytest --unit` (https://github.com/NREL/OpenOA/blob/main/test/conftest.py#L15-L37). 

I could certainly see `unit/` following this guidance on mimicking the structure, but regression and integration probably might be more focused on workflows than smaller units of testing.

docs/why_make_new_greenheart.md
```diff
@@ -0,0 +1,81 @@
...
+
+The end-goal of ODIES is to remove this call to HOPP as a monolith and instead break out the individual technologies so they are all exposed equally to ODIES.
+This would entail reworking the dispatch implementation so it is controlled by ODIES and not by HOPP, which is a non-trivial task.
+HOPP would still exist as a standalone tool, but major framework development would be done in ODIES.
+
+## Where should code I develop and implement live?
+
+Historically, ODIES has been a sort of hybrid itself, where it contains both the tool itself as well as project-specific code.
+Additionally, HOPP has been a separate tool that ODIES calls to obtain electricity production data and has been developed alongside ODIES.
+This has led to a somewhat disjoint codebase that is difficult to maintain and understand.
```
RHammond2, line 75: "disjointed"

tests/greenheart/test_hydrogen/test_h2_transport/test_h2_compressor.py
```diff
@@ -69,7 +69,7 @@ def test_system_power_report(self):
         )
         comp.compressor_power()
         _motor_rating, total_system_power = comp.compressor_system_power()
-        assert total_system_power == approx(3627.3907562149357)
+        assert total_system_power == 3627.3907562149357
```
RHammond2, line 72: Now, who's got the errant change! I'm not sure  how this one snuck in here too.

RHammond2 (commented):
Thanks for compiling all the thoughts on ODIES/new GreenHEART! I left a couple of comments on typos and small things like that, and some questions that may be out of scope for this PR that were at least worth raising.

.github/workflows/ci.yml
johnjasa: Huh, I think this snuck in like the approx change below. I'll revert this!

docs/class_structure.md
johnjasa: I get what you're saying in general, especially for code. I'm a fan of single sentence lines for docs regardless of length, though I do generally have a pretty wide monitor. Are you thinking we should have a ~100 char length line limit for docs too?

docs/describing_whats_in_classes.md
johnjasa: Definitely, could be close to gold plating or worse!

docs/development_guidelines.md
johnjasa: I think I prefer having the tests for each subpackage of the code living within that subpackage. So the tests wouldn't be off in their own folder with a mimicked structure, but instead right within the structure. Here's an example of that: https://github.com/OpenMDAO/OpenMDAO/tree/master/openmdao/core the core folder has files and then a `tests` folder that has unit tests for each of those files. What do you think about that kind of setup instead?

I agree that regression and integration would be separate in a higher level test folder.

PR #84, opened 2025-02-12
Plots

jaredthomas68:
# Include plotting functionality for greenheart output
This PR provides a few plotting options based on the `energy_flows.csv` data output from greenheart to visualize the electricity and hydrogen dispatch/flow/production independently and together. There are certainly better ways this could be done, but I'm not sure how far to go given the impending changes.

## PR Checklist

- [x] Documentation

- `examples/01-green-hydrogen.ipynb`: add example of using the new plots
- `greenheart/tools/eco/utilities.py`
  - `visualize_plant`: adjustments for slightly more automatic scaling and better looking plots for onshore and offshore
- `greenheart/tools/plot.py`
  - `get_hour_from_datetime`
  - `plot_hydrogen_flows`: new function
  - `plot_energy_flows`: new function
  - `plot_energy`: new function


examples/01-green-hydrogen.ipynb
kbrunik, line 1: @johnjasa can you remind me, did you say something about not pushing results in Jupyters?

greenheart/tools/eco/utilities.py
```diff
@@ -419,7 +419,7 @@ def visualize_plant(
     onshorex = 50
     onshorey = 50
 
-    wind_buffer = np.min(turbine_x) - (onshorey + 2 * rotor_diameter + electrolyzer_side)
+    wind_buffer = np.min(turbine_x) - (onshorey + 3 * rotor_diameter + electrolyzer_side)
```
kbrunik, line 422: Why did this change?

greenheart/tools/eco/utilities.py
```diff
@@ -889,10 +892,10 @@ def add_turbines(ax, turbine_x, turbine_y, radius, color):
         i = 0
         for x, y in zip(turbine_x, turbine_y):
             if i == 0:
-                elable = "Electrolyzer"
+                elabel = "Electrolyzer"
```
kbrunik, line 895: I'm wondering if we want to change this to "H$_2$ Electrolyzer" or "PEM Electrolyzer" since we are working on additional types of electrolysis (iron, alkaline, etc)

greenheart/tools/eco/utilities.py
```diff
@@ -1123,23 +1126,25 @@ def add_turbines(ax, turbine_x, turbine_y, radius, color):
...
-            solar_patch = patches.Rectangle(
-                (solarx, solary),
-                solar_side_x,
-                solar_side_y,
-                color=solar_color,
-                fill=None,
-                label="Solar Array",
-                hatch=solar_hatch,
-            )
+            if design_scenario["wind_location"] != "offshore":
```
kbrunik, line 1135: Should this be an if statement for "pv_location"?

examples/01-green-hydrogen.ipynb
jaredthomas68: I believe you are right @kbrunik. @RHammond2 I pushed these results since there were new figures in the example. What is the correct/accepted approach on this?

greenheart/tools/eco/utilities.py
jaredthomas68: Good point. I've changed it to "H$_2$ Electrolyzer" in all instances for now but am open to further changes.

greenheart/tools/eco/utilities.py
jaredthomas68: No, there are different plots to add solar to depending on if wind is onshore or offshore. Convoluted and should be fixed, but I think that making this plot more clear/robust should wait for another PR.

greenheart/tools/eco/utilities.py
```diff
@@ -1206,51 +1211,67 @@ def add_turbines(ax, turbine_x, turbine_y, radius, color):
...
+                round(np.min(turbine_y) - 0.3 * point_range_y, ndigits=roundto),
+                round(np.max(turbine_y) + 0.3 * point_range_y, ndigits=roundto),
+            ],
+        )
+        # ax[ax_index_wind_plant].autoscale()
+        ax[ax_index_wind_plant].set(aspect="equal")
+        # ax[ax_index_wind_plant].xaxis.set_major_locator(ticker.\
+        #   MultipleLocator(np.round(point_range_x*0.5, decimals=-3)))
+        # ax[ax_index_wind_plant].yaxis.set_major_locator(ticker.\
+        #   MultipleLocator(np.round(point_range_y*0.5, device_spacing=-3)))
```
johnjasa, line 1255: Here and in a few spots below here, do you intend to keep these commented lines on purpose?

johnjasa:
Looking at the generated plots, I think there's a small error in the `Electricity Output`; it should probably be divided by 1e6 instead of multiplied by 1e6 somewhere
image

johnjasa:
Non-blocking nitpick: at hour 0 there's a vertical line for net dispatch; is this meaningful or sort of a plotting artifact? is it easy to remove if it's spurious?
image

examples/01-green-hydrogen.ipynb
johnjasa: You're right @kbrunik we discussed it, though for now we should keep the jupyter notebooks as-is and push results with them. In the future, we might avoid keeping results in the repo; see (and add to!) the discussion here: https://github.com/NREL/GreenHEART/issues/75

jaredthomas68:
> Looking at the generated plots, I think there's a small error in the `Electricity Output`; it should probably be divided by 1e6 instead of multiplied by 1e6 somewhere. Okay, I've pushed up a fix directly to this branch, so no action needed for this point. The plots will be updated the next time you update the jupyter notebook itself.
> 
> Also, is the battery power in the correct units here? Seeing tens of GW spikes is throwing me off.
> 
> image

The spikes are due to having a 100 MW/100 MWh battery. During the simulation, one of the spikes actually charge in about 4 hours and then discharges in one hour. The x scale makes that hard to see, but here is the figure zoomed in a little in the x axis: 
image

johnjasa:
> > Looking at the generated plots, I think there's a small error in the `Electricity Output`; it should probably be divided by 1e6 instead of multiplied by 1e6 somewhere. Okay, I've pushed up a fix directly to this branch, so no action needed for this point. The plots will be updated the next time you update the jupyter notebook itself.
> > Also, is the battery power in the correct units here? Seeing tens of GW spikes is throwing me off.
> > image
> 
> The spikes are due to having a 100 MW/100 MWh battery. During the simulation, one of the spikes actually charge in about 4 hours and then discharges in one hour. The x scale makes that hard to see, but here is the figure zoomed in a little in the x axis: image

Thanks for looking into this, Jared! This makes sense, though I do have a question about the scale of the battery power. In the top right subplot of the figure you shared, it looks like battery power is in the 10s of GW as it charges and discharges. Is this correct for a 100MW/100MWh battery?

johnjasa (approved):

jaredthomas68:
> > > Looking at the generated plots, I think there's a small error in the `Electricity Output`; it should probably be divided by 1e6 instead of multiplied by 1e6 somewhere. Okay, I've pushed up a fix directly to this branch, so no action needed for this point. The plots will be updated the next time you update the jupyter notebook itself.
> > > Also, is the battery power in the correct units here? Seeing tens of GW spikes is throwing me off.
> > > image
> > 
> > 
> > The spikes are due to having a 100 MW/100 MWh battery. During the simulation, one of the spikes actually charge in about 4 hours and then discharges in one hour. The x scale makes that hard to see, but here is the figure zoomed in a little in the x axis: image
> 
> Thanks for looking into this, Jared! This makes sense, though I do have a question about the scale of the battery power. In the top right subplot of the figure you shared, it looks like battery power is in the 10s of GW as it charges and discharges. Is this correct for a 100MW/100MWh battery?

Fixed

PR #101, opened 2025-03-22
Fixes for HOPP v3.2.0, update PySAM version to 6.0+

johnjasa:
# Fixes for HOPP v3.2.0, update PySAM version to 6.0+

Now that HOPP v3.2.0 is up on pypi we can update to use the newer pySAM.
However, using HOPP v3.2.0 introduced some issues, specifically around the wind turbine spec tests.

In some cases, we have the rated power of the turbine set at e.g. 6MW, but the floris power curve tops out at e.g. 6123 kW.
HOPP v3.2.0 throws an error in this case.
I've updated the power curve files for those floris files to match the rated turbine power, which in turn caused many test values to change.
I've updated those accordingly.

As part of this, I've changed the CI to use HOPP develop to be in closer lockstep with the repo.
This will help catch issues like this prior to releases.
We can pin released versions of HOPP and H2I together moving forward.

I've also removed the defunct conda build (it was not used).

## PR Checklist

- [-] Documentation

## Related issues

N/A

Test values and examples updated

N/A

Test values updated due to power curve changes


jaredthomas68:
My initial reaction is that we should not be enforcing rated power to match the maximum turbine power. Depending on controls, turbines may operate above rated on the knee at the top of the power curve. I would lean towards a very loose check (maybe within 0.3 relative error or something) if any check like this.

examples/inputs/turbines/lbw_6MW.yaml
jaredthomas68, line 1: I don't think we should be changing the power curves.

tests/greenheart/input_files/turbines/osw_18MW.yaml
jaredthomas68, line 1: Again, I don't think we should change power curves. I think the check is wrong.

tests/greenheart/test_greensteel/input/turbines/lbw_6MW.yaml
jaredthomas68, line 1: Again, let's not change power curves

jaredthomas68 (changes requested):
I ignored the changed tests. I'd like to address the wind turbine rated power check first. I don't think it makes sense to change the power curves, I think the check should be looser to allow for turbines that operate slightly above rated. This is pretty common.

PR #128, opened 2025-05-08
Add custom model capabilities and an example

johnjasa:
# Add custom model capabilities and an example

Users might want to add their own performance, cost, or financial models that live outside of H2I.
These models might be proprietary or simply from another tool.

This PR introduces the capability for fully user-defined models to be used in an integrated way with existing wrapped models.
We showcase these capabilities with an extremely simple paper mill model.

## PR Checklist

- [x] Documentation


examples/06_custom_tech/tech_config.yaml
```diff
@@ -0,0 +1,57 @@
...
+        turbine_group: "onshore"
+        override_wind_resource_height: False
+        adjust_air_density_for_elevation: True
+      cost_parameters:
+        cost_per_kw: 1500.0
+  paper_mill:
+    performance_model:
+      model: "paper_mill_performance"
+      model_class_name: "PaperMillPerformance"
+      model_location: "paper_mill.py"
```
jmartin4nrel, line 42: This is great, I tested it out and paper_mill.py can be moved to any directory, relative or absolute, and it still works. Might be worth enclosing it in a subfolder called "custom_location" or something like that just to demonstrate that capability.

jmartin4nrel (approved):
This look like it should work perfectly to allow custom models (and models hosted on private repos) to get smoothly imported into H2I

examples/06_custom_tech/tech_config.yaml
johnjasa: Good idea, done and pushed up!

kbrunik (commented):
I really like it! I added a docs page for custom models, would be great if you could modify it for clarity and accuracy.

PR #131, opened 2025-05-12
MCH hydrogen storage cost model

elenya-grant:
# TOL/MCH H2 Storage Cost Model

TOL/MCH hydrogen storage cost model based on the article [linked here](https://www.nature.com/articles/s41467-024-53189-2)

## PR Checklist

- [x] Documentation

- `h2integrate/converters/hydrogen/eco_tools_pem_electrolyzer.py`
    - `ECOElectrolyzerPerformanceModel` added output `"rated_h2_production_kg_pr_hr"` (impacted `setup()` and `compute()` methods)
- `h2integrate/storage/hydrogen/eco_storage.py`
    - `H2Storage` component added input `"rated_h2_production_kg_pr_hr"` to `setup()` method
    - `H2Storage` component added call to MCH cost model in `compute()` method
    - added `MCHStorage` cost model import at top of file
- new folder: `h2integrate/simulation/technologies/hydrogen/h2_storage/mch/mch_cost.py`
    - `MCHStorage` class with MCH hydrogen storage cost model

### DIFF
--- a/h2integrate/simulation/technologies/hydrogen/h2_storage/mch/mch_cost.py
+++ b/h2integrate/simulation/technologies/hydrogen/h2_storage/mch/mch_cost.py
@@ -1,0 +1,94 @@
+from attrs import field, define
+
+from h2integrate.core.utilities import BaseConfig
+
+
+def calc_value(sizes, b0, b1, b2, b3, b4):
+    Hc, Dc, Ms, As = sizes
+    cost = b0 + (b1 * Hc) + (b2 * Dc) + (b3 * Ms) + b4 * As
+    return cost
+
+
+@define
+class MCHStorage(BaseConfig):
+    """
+
+    Sources:
+        - "Achieving gigawatt-scale green hydrogen production and
+        seasonal storage at industrial locations across the U.S" which is
+        [linked here](https://www.nature.com/articles/s41467-024-53189-2)
+
+    Args:
+        max_H2_production_kg_pr_hr (float): Maximum amount of hydrogen that may be
+            used to fill storage in kg/hr.
+        hydrogen_storage_capacity_kg (float): Hydrogen storage capacity in kilograms.
+        hydrogen_demand_kg_pr_hr (float):  Hydrogen demand in kg/hr.
+        annual_hydrogen_stored_kg_pr_yr (float): Sum of hydrogen used to fill storage
+            in kg/year.
+    """
+
+    max_H2_production_kg_pr_hr: float
+    hydrogen_storage_capacity_kg: float
+    hydrogen_demand_kg_pr_hr: float
+    annual_hydrogen_stored_kg_pr_yr: float
+
+    # dehydrogenation capacity [metric tonnes/day]
+    Dc: float = field(init=False)
+
+    # hydrogenation capacity [metric tonnes/day]
+    Hc: float = field(init=False)
+
+    # maximum storage capacity [metric tonnes]
+    Ms: float = field(init=False)
+
+    # annual hydrogen into storage [metric tonnes]
+    As: float = field(init=False)
+
+    # overnight capital cost coefficients
+    occ_coeff = (54706639.43, 147074.25, 588779.05, 20825.39, 10.31)
+    # fixed O&M cost coefficients
+    foc_coeff = (3419384.73, 3542.79, 13827.02, 61.22, 0.0)
+    # variable O&M cost coefficients
+    voc_coeff = (711326.78, 1698.76, 6844.86, 36.04, 376.31)
+
+    # cost year associated with the costs in this model
+    cost_year = 2024
+
+    def __attrs_post_init__(self):
+        # Equation (3): DC = P_avg
+        self.Dc = self.hydrogen_demand_kg_pr_hr * 24 / 1e3
+
+        # Equation (2): HC = P_nameplate - P_avg
+        P_nameplate = self.max_H2_production_kg_pr_hr * 24 / 1e3
+        self.Hc = P_nameplate - self.Dc
+
+        # Equation (1): AS = sum(curtailed_h2)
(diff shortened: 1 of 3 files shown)


tests/h2integrate/test_hydrogen/test_tol_mch_storage.py
```diff
@@ -0,0 +1,72 @@
...
+# Test values are based on Supplementary Table 3 of https://doi.org/10.1038/s41467-024-53189-2
+Dc_tpd = 304
+Hc_tpd = 304
+As_tpy = 35000
+Ms_tpy = 16200
+
+# Supplementary Table 3
+toc_actual = 639375591
+foc_actual = 10239180
+voc_actual = 17332229
```
johnjasa, line 15: I love that you used test values from the paper and made it clear!

tests/h2integrate/test_hydrogen/test_tol_mch_storage.py
```diff
@@ -0,0 +1,72 @@
...
+
+max_cost_error_rel = 0.06
```
johnjasa, line 17: Is this 6% difference just due to the fits not exactly interpolating the actual results? I'm guessing this is expected and not cause for concern but I still want to check.

h2integrate/storage/hydrogen/eco_storage.py
```diff
@@ -146,10 +154,32 @@ def compute(self, inputs, outputs):
...
+                [inputs["rated_h2_production_kg_pr_hr"][0], storage_max_fill_rate]
+            )
+            h2_charge_discharge = np.diff(hydrogen_storage_soc, prepend=False)
+            h2_charged_idx = np.argwhere(h2_charge_discharge > 0).flatten()
+            annual_h2_stored = sum([h2_charge_discharge[i] for i in h2_charged_idx])
+
+            h2_storage = MCHStorage(
+                max_H2_production_kg_pr_hr=max_rated_h2,
+                hydrogen_storage_capacity_kg=h2_storage_capacity_kg,
+                hydrogen_demand_kg_pr_hr=hydrogen_demand_kgphr,
```
johnjasa, line 169: Because this assumes that `hydrogen_demand_kgphr` exists, but that only gets instantiated if `size_capacity_from_demand` is True, do we need to account for that logic here? Specifically, what if a user wants to use the MCH cost model without sizing capacity from demand? I think we could either instantiate that variable even if capacity sizing isn't done or purposefully say this is not a supported capability -- I'm not sure if that's your intent, but that's why I'm checking.

h2integrate/simulation/technologies/hydrogen/h2_storage/mch/mch_cost.py
```diff
@@ -0,0 +1,98 @@
+from attrs import field, define
+
+from h2integrate.core.utilities import BaseConfig
+
+
+@define
+class MCHStorage(BaseConfig):
```
johnjasa, line 7: This threw me off for a bit because it's not being used in the same way as other `BaseConfig`-inheriting classes, i.e. it's not a `tech_config` object in the normal H2I sense. I don't think this is a problem, though the other storage models are just "normal" classes that don't inherit from `BaseConfig`.

It feels like it might be confusing to somebody looking at the config classes and seeing that most of them are just data classes (e.g. [the electrolyzer](https://github.com/NREL/H2Integrate/blob/develop/h2integrate/converters/hydrogen/eco_tools_pem_electrolyzer.py#L20)) but this one is not used directly from a `tech_config` and also that it does computations (i.e. it isn't just a data class; it does more).

So, all this to say, maybe I would prefer to have this be a regular class instead of inheriting from `BaseConfig`? Is there a specific reason or functionality afforded by this being a `BaseConfig`? Let me know if any of this doesn't make sense, I'm happy to talk through it!

johnjasa (changes requested):
Thanks for this, Elenya! I love the beautiful simplicity of the model and its implementation, as well as the fact that you included tests.

I pushed some changes directly to further simplify the MCH model and also left a few comments and questions here!

tests/h2integrate/test_hydrogen/test_tol_mch_storage.py
elenya-grant: yes - I don't think its cause for concern. I should note that I got this value from estimating the error between the LCOS (equation) from Supplementary Table 3 (LCOS = 2.05) for the same capacities compared to the value I got using Equation 7 from the paper with the coefficients listed under that equation (LCOS = 2.0175). That being said - I cannot figure out what resulted in different LCOS numbers and there is a possibility that it may be due to some error in my implementation. I have a test that shows this (currently fails) if that'd be useful.

h2integrate/storage/hydrogen/eco_storage.py
elenya-grant: I am not sure how to use the MCH cost model without specifying hydrogen demand because the user would then have to specify the dehydrogenation capacity somehow. I will make it throw an error for the time being if the `size_capacity_from_demand` is False and the storage type is 'mch'.

h2integrate/simulation/technologies/hydrogen/h2_storage/mch/mch_cost.py
elenya-grant: removed the BaseConfig inheritance, just a regular class now!

tests/h2integrate/test_hydrogen/test_tol_mch_storage.py
johnjasa: Thanks for this and the additional context! I see that the difference could be from the actual paper _or_ from your implementation, but it's not cause for concern.

johnjasa (approved):
Thanks for this, Elenya! I pushed up a docstring for the LCOS test you introduced to make it clear why it's added. Good to go from my end!

PR #163, opened 2025-06-17
Revamping ammonia production models

johnjasa:
# Revamping ammonia production models

As part of the ongoing work to model generation-following ammonia, we need to update our ammonia production models to be more capable and modular.
This PR takes the existing ammonia performance and cost models, removes the prior version (function-based), and replaces them with a native H2I version now dubbed the simple ammonia model.
We also introduce another ammonia production model that has hydrogen, nitrogen, and electricity as inputs instead of being handled as feedstocks.
This is the first step towards a more modular and detailed ammonia production model.

I also removed the `_autosummary` docs files because they're generated automatically during the doc creation process.

## PR Checklist

- [x] Documentation


h2integrate/converters/ammonia/test/test_ammonia_symloop_model.py
jmartin4nrel, line 1: Filename should be "synloop" not "symloop"
Line 12 should be MWh per kg not MW

jmartin4nrel (approved):
It seems in `test_ammonia_symloop.py` like we're setting H2, N2, and electricity in to all be at the perfect ratio needed for the ammonia. And then the next step is to build a solver or using the existing optimizer for the case of when we know the ammonia output we want but don't necessarily know how much H2/N2/electricity we need

PR #165, opened 2025-06-19
Add DOC

kbrunik:
# Add Direct Ocean Capture Technology

Add a direct ocean capture (DOC) technology to the H2Integrate framework. The technology model is located in the [MarineCarbonManagement Github Repository](https://github.com/NREL/MarineCarbonManagement/tree/main) and is not pip installable, as such the technology is integrated but isn't usable out of the box. A user would have to pip install the specific Github repository into their environment to use the DOC technology in a system.

Updated H2Integrate to allow all technologies to be able to access the `driver_config.yaml`.

Updated test suite to allow for installation of optional dependencies like [MarineCarbonManagement Github Repository](https://github.com/NREL/MarineCarbonManagement/tree/main) for CI testing on Github.

## PR Checklist

- [x] Documentation

- `h2integrate/converters/marine_carbon_capture/marine_carbon_capture_baseclass.py`
  - performance, cost, and finance base model classes for marine carbon capture models. 
- `h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py`
  - performance, cost, and finance model classes for DOC, analogous to electrolyzer_baseclass.py for renewable H2. This requires the mcm repository for the methods to run.
- `h2integrate/core/h2integrate_model.py and /supported_models.py`
  - added DOC to the supported models for performance, cost and finance.
- `examples/09_wind_wave_doc/`
  - an example of the direct ocean capture model with wind and wave energy offshore
- `.github/workflows/ci.yml`
  - updated workflow to install the MarineCarbonManagement Github repository for CI testing.
- `h2integrate/converters/marine_carbon_capture/test/test_doc.py`
  - added tests for DOC model testing both performance and that mcm is installed in environment.
- `pyproject.toml`
  - added testpaths
- All technologies
  - made `driver_config.yaml` an input for all technologies

Test values are the same as values derived when running echem_mcc model in [MarineCarbonManagement Github Repository](https://github.com/NREL/MarineCarbonManagement/tree/main).


h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
```diff
@@ -0,0 +1,228 @@
...
+            pump_config=echem_mcc.PumpInputs(),
+            seawater_config=echem_mcc.SeaWaterInputs(
+                sal=self.config.sal,
+                tempC=self.config.temp_C,
+                dic_i=self.config.dic_i,
+                pH_i=self.config.pH_i,
+            ),
+            save_outputs=True,
+            save_plots=True,
+            output_dir="./output/",  # TODO: how do I change this to be based on folder_output in driver_config.yaml?
```
kbrunik, line 104: @johnjasa Any suggestions on what to do about `save_outputs`, `save_plots`? I can set to false and they won't output for now but I wasn't sure if it would make sense to have them save to the outputs dir?

kbrunik:
Added an issue  #167 to improve financial model for the DOC. I think that can happen in a later PR and it's ready for review now.

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
johnjasa: I changed all technologies so they have access to driver_config and can use the folder for outputs. In the process, I realized we could streamline this more but probably outside of this PR so I made an issue here: https://github.com/NREL/H2Integrate/issues/166

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # TODO: figure out how to share this input between finance and cost model. it's an output of performance model
+        self.add_input(
+            "total_tank_volume_m3",
+            val=0.0,
+            units="m**3",
+        )
+
+    def compute(self, inputs, outputs):
+        ED_inputs = echem_mcc.ElectrodialysisInputs(
```
jmartin4nrel, line 172: So this is probably not a necessary thing to change if it doesn't slow things down... but it does seem like a shame that we're calling `echem_mcc.ElectrodialysisInputs()` repeatedly in both the performance and cost model `compute` functions, when we could just be preserving the output from the first call in the performance model and loading it back up in the cost model. @johnjasa have you been toying around with ways to save non-scalar, non-vector results of `compute` - i.e. things that don't fit nicely into a `self.add_output()`? My first instinct was to take a look at `hopp_wrapper`, in which we seem to have devised a method of caching outputs in a .pkl and loading them back up within the bounds of a `compute` function.

examples/09_wind_wave_doc/tech_inputs/hopp_config.yaml
```diff
@@ -0,0 +1,96 @@
...
+    system_capacity_kw: 50000
+    minimum_SOC: 20.0
+    maximum_SOC: 100.0
+    initial_SOC: 90.0
+    fin_model: !include default_fin_config.yaml
+  grid:
+    interconnect_kw: 362129.538
+    fin_model: !include default_fin_config.yaml
+
+config:
```
jmartin4nrel, line 83: I need to add a block here to get the example to run. HOPP version conflict? I am on hopp 3.3.0
```
  simulation_options:
    cache: true
    wind:
      skip_financial: false
```

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
```diff
@@ -0,0 +1,247 @@
...
+    def setup(self):
+        super().setup()
+        self.performance_config = DOCPerformanceConfig.from_dict(
+            merge_shared_inputs(self.options["tech_config"]["model_inputs"], "performance")
+        )
+        self.config = DOCCostModelConfig.from_dict(
+            merge_shared_inputs(self.options["tech_config"]["model_inputs"], "cost")
+        )
+
+        # TODO: figure out how to share this input between finance and cost model. it's an output of performance model
```
jmartin4nrel, line 164: To do this is pretty simple - just copy and paste the same `self.add_input()` into the `DOCFinance` class. I tried it out and it works just fine: 
image
image
image

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
```diff
@@ -0,0 +1,247 @@
...
+                "pip install git+https://github.com/NREL/MarineCarbonManagement.git"
+            )
+
+    def setup(self):
+        self.config = DOCPerformanceConfig.from_dict(
+            merge_shared_inputs(self.options["tech_config"]["model_inputs"], "performance")
+        )
+        super().setup()
+        self.add_output(
+            "plant_mCC_capacity_mtph",
```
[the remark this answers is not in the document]
johnjasa: Good call; this is fixed for this specific instance now but I've made an issue for future related work here: https://github.com/NREL/H2Integrate/issues/173

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
johnjasa: Great comments all around. I talked this through with Kaitlin and landed on this:
- In this case, I made a function that is called in both the performance and cost models that instantiates the inputs object. Because the object is not computationally expensive to create, I think that's the best and most clear option here.
- I added to the doc page with tips for how to approach cases when the performance and cost models are coupled, like this one, and also in the case when the shared object is more computationally expensive to create that you should use a pickle file like `hopp_wrapper` does as you suggested.

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
johnjasa: Thanks for doing that, @jmartin4nrel! And yes, you can have any amount of inputs connected to a single output. @kbrunik do you want to add logic to the finance model compute that uses the tank volume as part of this PR? If so, you should be able to do that by adding and using that input; let me know if I can help!

johnjasa:
Thanks, @jmartin4nrel! I've changed the PR based on your review comments which were very helpful.

h2integrate/converters/marine_carbon_capture/direct_ocean_capture.py
johnjasa: I just talked this through with Kaitlin, we can't remember why Kaitlin made this comment as it's not necessary to have the tank volume in the financial model. We've also reworked the financial model to use the existing H2I core instead of a special model.

jmartin4nrel (approved):
The next thing I would add (next PR maybe) is to pass cost inputs along from the `tech_config/cost_parameters` - of which there is currently only one into the `echem_mcc.ElectrodialysisCostInputs()`. Looking into the imported `mcm` module, we could make an option to feed it the `user_costs` option and these "user-provided costs": 
Screenshot 2025-07-15 140250

PR #188, opened 2025-07-18
Add an electricity splitter

johnjasa:
# Add an electricity splitter

Here we add an electricity splitter that can take in one stream of electricity and output two.
This is the backwards mirror to the existing combiner.
The splitter is more complex than just adding streams together because we must choose the breakdown of electricity going to the two out streams.

There are two ways to control the split of electricity:
- `prescribed_electricity` where you specify the electricity profile you want to send to the first connected technology. This can be a scalar value set in the config or an array set by an upstream technology.
- `fraction` where a value between 0 and 1 is the relative fraction of electricity input that is sent to the first technology.

I've added:
- `test_power_splitter.py` to test the splitter technology under different conditions
- `12_splitter_wind_doc_h2` to showcase the splitter in a joint end-use system
- `connecting_technologies.md` with more detail about connecting technologies and using combiners and splitters

Minor note: the LCOC and LCOH values in the example are relatively high as they consider the costs of the entire energy system, not just those technologies involved in producing that commodity. Once #176 we can change this behavior so only those technologies are included in the financial stackup.

## PR Checklist

- [x] Documentation


jmartin4nrel (changes requested):
Here are the only things I would change right now:
- Keep naming consistent: if `electricity` is the commodity that it is splitting, call it an `electricity_splitter`, not a `power_splitter`
- The `fraction_of_electricity_to_first_tech`  and `prescribed_electricity_to_first_tech` parameters of `SplitterPerformanceConfig` are a little vague. You have to jump back to the `plant_config` to remember which one is your `first_tech`. I would instead define another parameter, `priority_tech` - a str that gets validated to match the names of one of the techs going into the splitter. Then have that determine which tech gets the fraction or prescribed profile.

Here are things I would love to see in future improvements, not in this PR:
- A generic `splitter` that can split any defined `commodity_type`, not just `electricity`
- Handling of more than two splits so you don't have to daisy-chain. Maybe this could be achieved by turning my suggested `priority_tech` into `tech_priority_order`, with multiple fractions and/or prescribed profiles defined as a dict with the tech names as keys.
- Apply the same above two bullet points to the `combiner`
- Roll the `splitter` and `combiner` together into a `bus`, so we can have >2 electricity sources (e.g. wind, solar, nuclear) , and >2 electricity sinks (e.g. electrolysis, deferrable_load, non_deferrable_load)

jmartin4nrel (approved):
This looks great, don't let me stop you from merging once conflicts are resolved

PR #178, opened 2025-07-15
ProFAST params optional input

elenya-grant:
# Option to provide ProFAST params in finance parameters

In the `plant_config`, the user can provide a dictionary of ProFAST parameters to be used instead of populating ProFAST parameters from hard-coded defaults and some other user-provided finance parameters. To utilize this functionality, the finance parameters specified in the `plant_config.yaml` file required (and used) are:
```yaml
finance_parameters:
  pf_params: !include "profast_params.yaml"
  costing_general_inflation: 0.025 # used to adjust modeled costs to cost_year
  depreciation_method: "MACRS" 
  depreciation_period: 5 
  discount_years:
```
where the `profast_params.yaml` file should be formatted as follows (did not include all profast params here):
```yaml
params:
  long term utilization: 1 #capacity factor
  analysis start year: 2032
  operating life: 30 
  installation months: 36 
  demand rampup: 0 
  ...

```

## PR Checklist

- [x] `CHANGELOG.md` has been updated to describe the changes made in this PR
- [-] Documentation
  - [-] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [-] Examples have been updated
- [x] Tests pass (If not, and this is expected, please elaborate in the tests section)
- [x] PR description thoroughly describes the new feature, bug fix, etc.

## Related issues

Standardize definition of analysis start year Issue #99 

## Impacted areas of the software

- `h2integrate/core/finances.py`
  - `ProFastComp.compute()`: updated setting profast parameters (the financial ones) if user provides `pf_params` in `finance_parameters`
- `docs/user_guide/specifying_finance_parameters.md`: new documentation file to give some brief information on financial parameters.
- `docs/_toc.yml`: updated to include `user_guide/specifying_finance_parameters.md` file.
- `h2integrate/core/test/test_finances.py`: added test named `test_profast_config_provided` to test changes from this PR, equivalent to `test_electrolyzer_refurb_results` test but has finance parameters specified as `pf_params`.

PR #210, opened 2025-08-11
PV example and LCOE Test

elenya-grant:
# PV example and LCOE Test

New example and LCOE test for utility-scale PV using cost and performance data from the [2024 ATB workbook](https://data.openei.org/files/6006/2024_v3_Workbook.xlsx). The tests allow for more than usual error to account for possible differences in the pv model parameters, LCOE financial parameters (compared to the ATB), PV capacity factor due to differences in the PV model, differences in reflecting lifetime performance (from H2I model compared to how it was used in ATB), etc.

The new example utilizes the following new models/features:
- uses "pysam_solar_plant_performance" as the performance model (PR #193)
- uses "atb_utility_pv_cost" cost model (PR #187)
- inputs financial parameters with profast parameter input file (PR #178)

## PR Checklist

- [x] Documentation

- `tests/h2integrate/test_all_examples.py`
  - `test_pv_example()`: test LCOE and capacity factor for Example 15. The output values are tested against values from the 2024 ATB Workbook for Utility-PV - Class 2, using 2030 as the year and "Moderate" technology advancement scenario. 
- `examples/15_solar_pv/`
  - `plant_config.yaml`: plant file that uses the same solar resource as Example 11
  - `driver_config.yaml`: driver config file, nothing too exciting here
  - `tech_config.yaml`: technology config file that aims to replicate the representative technology for utility-scale PV described in the ATB but somewhat modified given that this example is based on 2030 technology and the ATB gives more detail on the [2035 technology](https://atb.nrel.gov/electricity/2024/utility-scale_pv#scenario_descriptions)
  - `profast_params.yaml`: profast params input file (alternative method to specify finance parameters)
  -  `run_solar_pv.py`: python run script for example


johnjasa (changes requested):
I like the idea of this PR -- comparing the results we're getting and the ATB seems worthwhile! I want to get a better idea of your goals with this and that can help shape my further feedback.

Is your main goal to show that we're relatively close to the ATB results, but they're slightly different? If so, I'd suggest a doc page that goes through the results, shows the differences, and links to the ATB numbers. This way users could see and understand the similarities and differences more clearly. I would suggest removing the example test then.

If the main goal is to ensure that results do not change in the solar cost model, then I don't think the example test is necessarily capturing that because of the large tolerance on the asserts. There's also already the `test_pysam_with_atb_costs.py` file ensure values do not change coming out of the model.

In summary, thank you for this! But I think some iteration on the goal and how we're achieving that would be helpful.

kbrunik:
> I like the idea of this PR -- comparing the results we're getting and the ATB seems worthwhile! I want to get a better idea of your goals with this and that can help shape my further feedback.
> 
> Is your main goal to show that we're relatively close to the ATB results, but they're slightly different? If so, I'd suggest a doc page that goes through the results, shows the differences, and links to the ATB numbers. This way users could see and understand the similarities and differences more clearly. I would suggest removing the example test then.
> 
> If the main goal is to ensure that results do not change in the solar cost model, then I don't think the example test is necessarily capturing that because of the large tolerance on the asserts. There's also already the `test_pysam_with_atb_costs.py` file ensure values do not change coming out of the model.
> 
> In summary, thank you for this! But I think some iteration on the goal and how we're achieving that would be helpful.

Since Jared mentioned that almost all financial models will return different results I've been trying to ponder the best way to approach validating our financial methods. I'm not sure that I have cracked the code on the best way to approach this. I think I agree with John, given the large tolerances on the tests and differences that aggregate from the performance modeling and financial models I'm not sure this PR makes it clearer that everything is implemented correctly.

I appreciate you being willing to work on this cryptic issue but I'm wondering if we move this back into the rumination phase on how to approach the issue.

elenya-grant:
will go back to the drawing board on this one and will open a new PR in the future once it's determined what this example should accomplish.

Issue #204, opened 2025-08-07
Allow variable time step number and length


jaredthomas68:
# Allow variable time step number and length
We are limited to 8760 hourly time steps. It would be helpful to allow a flexible number of time steps and time step duration for a range of reasons, including: testing short simulations, running simulations with different data for multiple years, adjust temporal fidelity, etc.

## Proposed solution
For the number of time steps, I think we could start with putting in an `n_time_steps` parameter at the top level config that will then flow down to all levels of the simulation.

For the time step duration, it is more complicated. Many models assume hourly time steps. For those models we may need to still run hourly and then up- or down-sample after that model runs so the output can be provided in the desired number of steps and step duration.

## Additional context
Several technologies use a default hard-coded time step number, as does control. With so much dynamic sizing, it is difficult to determine the size at model setup time if needed, as in demand rule-based control (#195).

johnjasa:
Related to https://github.com/NREL/HOPP/issues/420

elenya-grant:
My thoughts are:

The plant_config could either include these under 'plant' or under a new section like 'simulation'
```yaml
plant:
   plant_life: 30
   dt: 3600 # number of seconds per timestep
   n_timesteps: 8760 # number of time steps per simulation 
   simulation_start_time: 01/01/2022 01:30:00 # should be formatted as mm/dd/yy HH:MM:SS 
   simulation_timezone: -5 # should be UTC offset, timezone corresponding to simulation_start_time
```

simulation_start_time is formatted in datetime package as `'%m/%d/%Y %H:%M:%S'`
OR

```yaml
plant:
   plant_life: 30
simulation:
   dt: 3600 # number of seconds per timestep
   n_timesteps: 8760 # number of time steps per simulation 
   start_time: 01/01/2022 01:30:00 # should be formatted as mm/dd/yy HH:MM:SS 
   timezone: -5 # should be UTC offset, timezone corresponding to simulation_start_time
```

Then models can access `dt` and `n_timesteps` from (for example) `self.options['plant_config']['simulation']['dt']` and `self.options['plant_config']['simulation']['n_timesteps']`

johnjasa:
Other thoughts from Elenya:

Considerations beyond simply updating all models to handle varying time-steps and simulation lengths

[...]

PR #219, opened 2025-08-19
Time info in `plant_config`

elenya-grant:
# Added simulation time info to `plant_config`

The purpose of this PR is to make simulation information, such as simulation length (`n_timesteps`) and time interval (`dt`) accessible to models through the `plant_config` in the future rather than models using hard-coded values. Hopefully, allowing these variables to be accessed from the `plant_config` would encourage development future models to be flexible for varying simulation length and time step interval. **This is not intended to update existing models to handle varying simulation lengths and timesteps, but to make these time variables accessible for future development.**

```yaml
plant:
   plant_life: 30
   simulation:
      dt: 3600 # number of seconds per timestep, required and defaults to 3600
      n_timesteps: 8760 # number of time steps per simulation,  required and defaults to 8760
      start_time: 01/01 00:30:00 # should be formatted as mm/dd HH:MM:SS or mm/dd/yyyy HH:MM:SS, optional. Could be used in simulations that are not one year
      timezone: -5 # UTC offset in hours, timezone corresponding to start_time. Could be used to align and adjust timeseries data to the same timezone and start-time.
```

## PR Checklist

- [-] Documentation

## Related issues

#204 

- `h2integrate/core/inputs/plant_schema.yaml`
  - `site`: removed `elevation` and `timezone` from required properties - these aren't used anywhere at the moment
  - `plant`: added `simulation` section with variables `dt`, `n_timesteps`, `start_time`, `timezone` as possible properties.
- `h2integrate/core/inputs/validation.py`: added checks in `load_plant_yaml()` for non-supported simulation duration or time-step interval
- `h2integrate/core/test/test_framework.py`: added test `test_unsupported_simulation_parameters` to ensure that errors are thrown if an invalid simulation length is input or invalid time step.
- replaced `n_timesteps = 8760` with `n_timesteps = self.options['plant_config']['plant']['simulation']['n_timesteps']` in the following files:
  - `h2integrate/converters/solar/solar_baseclass.py`
  - `h2integrate/converters/water_power/hydro_plant_baseclass.py`
  - `h2integrate/converters/water_power/hydro_plant_run_of_river.py`
  - `h2integrate/converters/wind/wind_plant_baseclass.py`

### Note on updating all hard-coded instances of 8760
Some models do not have 'plant_config' as an option so it's not a simple find and replace `n_timesteps` with the value provided in `plant_config`. Some of these models include:
- `h2integrate/core/resource_summer.py`
- `h2integrate/resource/river.py`

Similarly, some models may only work/function as expected for hourly timesteps and for one-year of simulation. It would be nice to document which models are not flexible to varying time-steps or have limitations on what time-steps or simulation lengths they could handle.


h2integrate/core/inputs/plant_schema.yaml
```diff
@@ -58,7 +58,30 @@ properties:
...
+            minimum: 1
+          start_time:
+            type: string
+            description: simulation start time formatted as mm/dd/yyyy HH:MM:SS or mm/dd HH:MM:SS
+            default: "01/01 00:30:00"
+          timezone:
+            type: number
+            description: timezone as UTC offset corresponding to start_time
+            minimum: -12
+            maximum: 14
```
johnjasa, line 82: I didn't realize it was a 26 hour time spread available via UTC offsets, that's fun!

h2integrate/core/utilities.py
```diff
@@ -244,6 +245,50 @@ def check_plant_config_and_profast_params(
         raise ValueError(msg)
 
 
+def make_time_profile(
```
johnjasa, line 248: What is this function used for? Or what's the vision for it? I don't see it used or tested anywhere.

johnjasa (changes requested):
Great addition! Thanks for handling the schema updates. I'll update some models that are hardcoded as part of this PR. One blocking question about the new utilities function.

h2integrate/core/inputs/plant_schema.yaml
elenya-grant: these are the same bounds as was there for the site timezone. The datetime package says that UTC offset just has to be +/- 24 (see the documentation [here](https://docs.python.org/3/library/datetime.html#datetime.tzinfo.utcoffset))

h2integrate/core/utilities.py
elenya-grant: unused as the moment but - it could be helpful for aligning wind and solar resource data to the same time of year. I can make a test for it if that's the bigger concern, or I can remove it for now and include it in a PR that requires it.

bayc (changes requested):
Seems fine to me. Could you add a check that throws an error if `dt` or `n_timesteps` if not equal to 3600 or 8760? I would say it would be safer to enforce that from the get go since all models do not support changing the number of timesteps. What do you think?

Issue #228, opened 2025-08-22
Add variable OpEx


johnjasa:
# Add variable OpEx

Elenya's PR #222 contained a neat introduction to variable OpEx.
This is especially useful for cases where costs vary year-to-year, such as feedstock consumption.

## Proposed solution

We currently only consider OpEx as a singular entity and do not differentiate between fixed and variable OpEx.
We could separate these ideas out and change the existing OpEx to total OpEx (or similar) once we introduce the subtypes of OpEx.

To correctly handle variable OpEx that varies year-to-year, we will need to change the way the financial calculations are done as they assume in some places that what happens for one year happens across the project lifetime.

PR #222, opened 2025-08-20
WIP: feedstock cost model and variable o&m

elenya-grant:
# Feedstock Cost Model

**Note: this depends on work done in PR #219 and will require further additions to work done in PR #218 before this can be finalized**

Added feedstock cost model that is flexible to handle different types of cost profiles. Costs can be provided as:
-  a scalar, which means that costs are constant across timesteps and years
-  a list that has a length equal to `plant_life` which means that costs are constant during a year but may change year-to-year
- a list that has a length equal to `n_timesteps` which means that costs change per timestep.

## PR Checklist

- [ ] `CHANGELOG.md` has been updated to describe the changes made in this PR
- [ ] Documentation
  - [ ] Docstrings are up-to-date
  - [ ] Related `docs/` files are up-to-date, or added when necessary
  - [ ] Documentation has been rebuilt successfully
  - [ ] Examples have been updated
- [ ] Tests pass (If not, and this is expected, please elaborate in the tests section)
- [ ] PR description thoroughly describes the new feature, bug fix, etc.

## Impacted areas of the software

- `h2integrate/core/feedstocks.py`
  - `FeedstockCostConfig`: configuration file for feedstock costs
  - `FeedstockCostModel`: cost model for some generic feedstock
- `h2integrate/core/model_baseclasses.py`: added `VarOpEx` output
- `h2integrate/core/test/test_feedstock.py`: new file for testing feedstock component. 

Remaining to-dos
- add connection to feedstock cost component in `h2integrate/core/h2integrate_model.py`
- update both models in `h2integrate/core/finances.py` to include variable o&m (this is waiting on profast refactor pull-request to be merged in)
- Also requires PR on time information to be merged in

PR #235, opened 2025-08-28
Added variable o&m

elenya-grant:
# Added variable o&m

Added variable O&M costs into cost models and financial models to enable future-usage of this variable in technology cost models. This functionality would likely be best suited for feedstock costs (PR #222 and PR #221) and any technology that has variable operation costs.

This will be followed-up with a PR with an updated feedstock cost model (updated from PR #221 with functionality that exists in PR #222) that outputs variable o&m.

## PR Checklist

- [x] Documentation

## Related issues
Issue #228

- `h2integrate/core/h2integrate_model.py`: in `connect_technologies()` connected VarOpEx to financial group
-  `h2integrate/core/profast_financial.py`
  - `ProFASTDefaultVariableCost`: default config for variable cost items
  - `ProFastComp.setup()`: added input for adjusted varopex and initialization of self.variable_cost_settings which uses the`ProFASTDefaultVariableCost` config
  - `ProFastComp.compute()`: added variable o&m into populating the profast dictionary workflow.
- `h2integrate/core/model_baseclasses.py`
  - `CostModelBaseClass`: added `VarOpEx` as an output.
- `h2integrate/core/finances.py`: 
  - `AdjustedCapexOpexComp`: added input and corresponding output to adjust variable o&m costs.
- `tests/h2integrate/test_finance.py`: added two tests to show how costs change with added variable o&m costs
  - `test_variable_om_no_escalation()`: tests variable o&m values input into ProFAST as either scalar or dictionary with zero escalation.
  - `test_variable_om_with_escalation()`: tests variable o&m values input into ProFAST as either scalar or dictionary with 0.025 escalation.
- `h2integrate/core/feedstocks.py`: updated `compute()` method to output the cost pear year as `VarOpEx`
- `tests/h2integrate/test_all_examples.py::test_natural_gas_example`: updated accessing opex variables for tests, did not change test values.
Other files with minor changes:
- `h2integrate/converters/hydrogen/wombat_model.py`: added VarOpEx as an output, defaults to 0.0
- `h2integrate/converters/ammonia/test/test_simple_ammonia_model.py`: added plant config as input to cost models (used to be empty dictionary)
- `h2integrate/converters/solar/test/test_pysam_with_atb_costs.py`: added plant config as input to cost models (used to be empty dictionary)
- `h2integrate/core/test/test_ro_desalination.py`:  added plant config as input to cost models (used to be empty dictionary)


h2integrate/core/finances.py
```diff
@@ -17,32 +18,43 @@ def setup(self):
...
+        total_varopex_adjusted = np.zeros(self.plant_life)
         for tech in self.options["tech_config"]:
             capex = float(inputs[f"capex_{tech}"][0])
             opex = float(inputs[f"opex_{tech}"][0])
+            varopex = inputs[f"varopex_{tech}"][0]
             cost_year = int(discrete_inputs[f"cost_year_{tech}"])
             periods = self.target_dollar_year - cost_year
             adjusted_capex = -npf.fv(self.inflation_rate, periods, 0.0, capex)
             adjusted_opex = -npf.fv(self.inflation_rate, periods, 0.0, opex)
+            adjusted_varopex = -npf.fv(self.inflation_rate, periods, 0.0, varopex)
```
jaredthomas68, line 50: I'm not sure how varopex provided as a list and including escalation is handled in ProFAST. I think it is important to add some tests to make sure the expected behavior is happening with the following: varopex as a scalar, varopex as an array, the previous two with zero and non-zero excalation.

elenya-grant: so - I made some tests and the this is the situation:
- zero escalation: LCO is the same regardless of whether feedstocks are provided as scalar or array
- nonzero escalation: LCO is lower with costs provided as list than as scalar in ProFAST.
   - if cost is provided as a scalar, then ProFAST uses the same cost for every year
   - if cost is provided into profast as a dictionary (with all the values being equal), then in ProFAST, the cost reduces every following year. 
   - This difference is because the "escalation" parameter of the variable o&m inputs is nonzero. This does not have to do with the "general inflation rate" parameter.

Part of the motivation to allow for costs to differ across years is partially because this is the only way to reflect varying costs across the years of the plant life due to either a) different "usage" per year or b) different costs per year. Different "usage" per year would already be handled *if* we set the `long term utilization` to the actual long-term utilization (as a capacity factor per year) and set the `capacity` to the actual rated capacity. 

I could also add logic into this so that if the variable o&m cost is the same for each year, to input the variable o&m as a scalar value rather than a dictionary into ProFAST.

h2integrate/finances/profast_financial.py
```diff
@@ -339,6 +339,30 @@ def create_dict(self):
...
+
+    Attributes:
+        escalation (float | int, optional): annual escalation of price.
+            Defaults to 0.
+        unit (str): unit of the cost, only used for reporting. The cost should be input in
+            USD/unit of commodity.
+        usage (float, optional): Usage of feedstock per unit of commondity.
+            Defaults to 1.0.
+
+
```
johnjasa, line 348:
```suggestion
```

h2integrate/finances/finances.py
```diff
@@ -44,32 +45,43 @@ def setup(self):
...
+        self.add_output("total_varopex_adjusted", val=0.0, shape=plant_life, units="USD/year")
 
     def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
         total_capex_adjusted = 0.0
         total_opex_adjusted = 0.0
+        total_varopex_adjusted = np.zeros(self.plant_life)
         for tech in self.options["tech_configs"]:
             capex = float(inputs[f"capex_{tech}"][0])
             opex = float(inputs[f"opex_{tech}"][0])
+            varopex = inputs[f"varopex_{tech}"][0]
```
johnjasa, line 72: This `[0]` means that we're only grabbing the first entry in this array that's length `plant_life`. Is this intentional? This means that all entries will have the same value in the array and won't vary year-to-year.

h2integrate/finances/profast_financial.py
```diff
@@ -339,6 +339,30 @@ def create_dict(self):
...
+class ProFASTDefaultVariableCost(BaseConfig):
+    """Configuration class of default settings for ProFAST variable costs.
+    The total cost is calculated as ``usage*cost``.
+
+    Attributes:
+        escalation (float | int, optional): annual escalation of price.
+            Defaults to 0.
+        unit (str): unit of the cost, only used for reporting. The cost should be input in
+            USD/unit of commodity.
+        usage (float, optional): Usage of feedstock per unit of commondity.
```
johnjasa, line 345:
```suggestion
        usage (float, optional): Usage of feedstock per unit of commodity.
```

h2integrate/finances/finances.py
johnjasa: In general, I'd suggest not using the `[0]` indexing when dealing OM variables and only doing so when we absolutely need a `float` value and not an `array`. All variables (inputs and outputs) are represented as numpy arrays even if it's just an array of length 1. This is slick because all the normal numpy operations work regardless of if it's length 1, or size (360, 120, 4) (for example).

h2integrate/finances/profast_financial.py
```diff
@@ -582,9 +622,28 @@ def compute(self, inputs, outputs):
...
+                self.tech_config[tech]["model_inputs"]
+                .get("financial_parameters", {})
+                .get("variable_costs", {})
+            )
+
+            # add CapEx cost to tech-specific variable cost entry
+            varom_cost_per_unit_commodity = inputs[f"varopex_adjusted_{tech}"] / total_production
+            if not all(v == 0.0 for v in varom_cost_per_unit_commodity):
+                varom_dict = dict(zip(years_of_operation, varom_cost_per_unit_commodity))
+                tech_varopex_info.update({"cost": varom_dict})
```
johnjasa, line 633:
```suggestion
            # add CapEx cost to tech-specific variable cost entry
            varopex_adjusted_tech = inputs[f"varopex_adjusted_{tech}"]
            if np.any(varopex_adjusted_tech) > 0:
                varopex_cost_per_unit_commodity = varopex_adjusted_tech / total_production
                varopex_dict = dict(zip(years_of_operation, varopex_cost_per_unit_commodity))
                tech_varopex_info.update({"cost": varopex_dict})
```

Two main changes here:
- using numpy array logic for the if statement to make it easier to read
- using `varopex` instead of `varom` to be consistent naming-wise

johnjasa (changes requested):
Great PR! I love that you already addressed Jared's comments/questions so well with added tests.

I pushed some minor things up directly -- please take a look to see what I did and that you're on board for the changes. I also have some small comments here, nothing too major but I would like to double-check a few things.

Lastly, could you please add a sentence or few in the docs about the addition of variable OpEx? Maybe in the "adding a new tech", "cost year of cost models", and "expected methods and attributes of classes" (and anywhere else you see fit)? Thank you!

PR #239, opened 2025-09-04
Integrating methanol with Direct Ocean Capture CO2

jmartin4nrel:
# Integrating methanol with Direct Ocean Capture CO2

This sets up a new example in which the existing direct ocean capture tech is used to send CO2 to a CO2 hydrogenation methanol production plant located offshore on the Gulf. This plant also uses hydrogen from electrolysis powered by offshore wind with H2 storage and dispatch, tying a bunch of tech nicely together. The current example uses 312 MW of electrolysis (~ the size found feasible in the ECO project) combined with 2000 tons of H2 storage to keep a 100,000 tpy methanol plant fullyk supplied with the H2 it needs for most of the year, with some shortfall in the summer months when wind is lower. Currently there is not a CO2 storage and dispatch (at least not a working one - @kbrunik is there something we could implement from the DOC model?), so the methanol plant does not operate at full capacity.

## Type of Contribution

- [X] Feature Enhancement
  - [ ] New Technology Model
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## General PR Checklist

- [x] `CHANGELOG.md` has been updated to describe the changes made in this PR
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [x] Examples have been updated (if applicable)
- [x] Tests pass (If not, and this is expected, please elaborate in the tests section)
- [x] Added tests for new functionality or bug fixes
- [x] PR description thoroughly describes the new feature, bug fix, etc.

## Impacted areas of the software

- `examples/03_methanol/co2_hydrogenation_doc`: Added this example of DOC integrating with methanol, particular files of interest:
  - `examples/03_methanol/co2_hydrogenation_doc/run_co2h_methanol.py`: Script to run this example
  - `examples/03_methanol/co2_hydrogenation_doc/plot_co2h_methanol.py`: A plotting function imported into the above script which draws out all the material flows going into the methanol plant
- `converters/co2/marine/direct_ocean_capture.py`: Fleshed out docstrings, added a `co2_out` output for interconnection
- `converters/co2/marine/marine_carbon_capture_baseclass.py`: Added a `co2_out` output for interconnection
- `converters/methanol/methanol_baseclass.py`: Added a `total_methanol_produced` output for ProFast finance calculations
[...]

PR #241, opened 2025-09-04
Making pipe and cable substance-agnostic

jmartin4nrel:
# Making pipe and cable substance-agnostic

As it is, the `pipe` and `cable` can only carry one substance each, `hydrogen` and `electricity` respectively, and this is hard-coded into the transporters. Since these are meant to be general lossless transporters they should be able to carry other substances which I call `transport_item`, and I need them to for the purposes of linking other tech modules - in particular CO2 capture and methanol production.

## Type of Contribution

- [X] Feature Enhancement
  - [ ] New Technology Model
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

## General PR Checklist

- [x] Documentation

- `h2integrate/core/h2integrate_model.py`: Added one line which passes a new `transport_item` parameter to `pipe` and `cable`
- `h2integrate/transporters/pipe.py`: Allows a `pipe` to carry any of a defined list of substances, not just `hydrogen`
- `h2integrate/transporters/cable.py`: Allows a `cable` to carry any of a defined list of substances, not just `electricity`, although `electricity` is currently the only defined substance. In the future I imagine we may want to further delineate ac/dc and voltage levels.

This was originally wrapped up in PR #239 , but I realized it should be split off into its own PR since it is a change to a core function that affects other tech. It is needed for that PR to carry CO2 from ocean capture to methanol production.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@
 - Updated option to pass variables in technology interconnections to allow for different variable names from source to destination in the format `[source_tech, dest_tech, (source_tech_variable, dest_tech_variable)]`
 - Added `simulation` section under `plant_config['plant']` that has information such as number of timesteps in the simulation, time step interval in seconds, simulation start time, and time zone.
 - Added `"custom_electrolyzer_cost"` model, an electrolyzer cost model that allows for user-defined capex and opex values
+- Made `pipe` and `cable` substance-agnostic rather than hard-coded for `hydrogen` and `electricity`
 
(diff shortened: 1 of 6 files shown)


johnjasa (approved):
Looking great, good change, thanks Jonathan!

I modified the way the `transport_type` is set so it uses the OpenMDAO options inputs. If we later have more complex transport components with more inputs, we should probably use a config setup like other components. That's a later issue.

I also added a simple test for the pipe case to make sure unallowed options raise an informative error.

PR #433, opened 2026-01-02
Iron: Example with independently sized mine and iron/steel plant

jmartin4u:
# Example with independently sized mine and iron/steel plant

This PR changes the iron example (#21) so that the `iron_mine` can be realistically large and produce a large `iron_ore_out` stream that is not entirely consumed by the downstream `iron_plant`. This allows the `price_iron_ore` to be calculated separately in its own financial subgroup, then passed to the `iron_plant` via a feedstock. To achieve this, I exposed the `price` config variable in the feedstock model as an OpenMDAO input.

This is still a draft PR because there is one issue that needs to be resolved: When `iron_transport` is connected to an `iron_ore` feedstock, it calculates its cost based on the `iron_ore_in` connected to the `iron_ore_out` from the feedstock, which is calculated in the feedstock performance model as the full `rated_capacity`. The `iron_transport` needs its `iron_ore_in` to instead be connected to the `iron_ore_consumed` from the `iron_plant`, but I'm not sure how to do this. I think it is only an issue when transporters themselves are calculating cost per unit of the feedstock transported - it doesn't crop up when we are just using simple pass-through `pipe` and `cable` transporters. I think @elenya-grant is probably the best person for this and tagged her in the TODO list

At the same time I am working in a doc page for `iron_mine` which was missing. If it's not cramming too much into one PR, it would be great if @kbrunik could write a doc page for `iron_dri_plant` too.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] @elenya-grant I need to resolve how to calculate accurate `iron_transport` costs using a feedstock. 
- [ ] @kbrunik Can you write a doc page `iron_plant.md` to go with the `iron_mine.md` page that's already in this PR?

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [ ] Documentation

### Section 4.1: New Files
- `docs\technology_models\iron_mine.md`: Rectifying the missing technology doc page for `iron_mine`

- `examples\21_iron_mn_to_il\`: Updated example to show the breakdown of ore, iron, and steel costs, removed comparisons with old code now that we are fully running on H2I framework.
  - `tech_config.yaml` - Added a `processed_ore_feedstock` that represents the subset of the `iron_mine` ore output that is passed to `iron_plant`, so that these two converters can be sized independently
- `examples\test\test_all_examples.py`: Slight differences in test values due to changes in the example
- `h2integrate\core\feedstocks.py`: Changed `price` from a config variable to an input

When I run `run_iron.py`, the last two cases *should* produce the same levelized costs, but they are not. This is because the `iron_transport` transporter is calculating its costs based on the `rated_capacity`. I want to get this running so that when I change the `rated_capacity` of `processed_iron_ore_feedstocks`, the levelized costs are not affected (as long as the capacity is above what the `iron_plant` needs to consume).

The three subtests in `test_all_examples.test_21_iron_dri_eaf_example` all had to have their values modified:

In the first test, iron ore LCO went down. This is expected because the `mine_electricity_feedstock` price was set to zero. I made this change because, even though we added an `electricity_in` input to the `iron_mine`, the cost of the electricity these mines use was already accounted for in the `processing_cost`. Putting a price on the feedstock would thus effectively double-count the cost of electricity. In a revised ore processing model from NRRI, we should try to get the `processing_cost` broken out more so that we can properly account for different feedstock electricity costs.

In the second and third tests, iron and steel LCO went up. This was not expected - apart from the above change in the ore cost to prevent double-counting electricity (which should have brought costs down), there was another change made: the ore cost was treated as a feedstock cost, rather the capex/opex of the mine being rolled into and adjusted total capex/opex for the whole system. If we were doing a simple LCO model with a fixed charge rate, this wouldn't make a difference, but with ProFast it is making a difference. I'd need to dig more into how the ProFast LCO calculation is happening to understand what causes these differences. But bottom line, this simulation is meant to treat the iron/steel producers as separate entities from the mines, so they should be treating iron ore as a feedstock cost and not be considering capital costs of the mines themselves. None of the individual component capex/opex values have changed, so I'm comfortable sticking with the new test values.

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/transporters/generic_splitter.py
```diff
@@ -138,10 +158,19 @@ def compute(self, inputs, outputs):
             requested_amount = np.maximum(0.0, prescribed_to_priority)
             commodity_to_priority = np.minimum(requested_amount, available_commodity)
             commodity_to_other = commodity_in - commodity_to_priority
+        total_priority = np.sum(commodity_to_priority)
+        total_other = np.sum(commodity_to_other)
+        # Need to do this in case _out is different units than _producced
+        frac_priority = total_priority / (total_priority + total_other)
+        total_produced = inputs[f"total_{self.config.commodity}_produced"]
+        total_priority = total_produced * frac_priority
+        total_other = total_produced * (1 - frac_priority)
```
johnjasa: May be able to refactor this to not require `total_iron_ore_produced` and instead get the sum in the splitter directly. This simplifies the tech interconnections. Suggested by @dakotaramos

elenya-grant (approved):
I think this looks good! Thanks for updating this example, adding a doc page, and removing the old files. And thanks for updating the feedstock component to be more flexible! This PR looks good to me but I'd want your review of the changes I made before it gets merged in! 

I made the change you requested, but would like your feedback on it (many of my comments are pointing out where I made changes). Its a bit of a workaround at the moment but I don't think we should make any large changes to the H2I framework to accommodate the specific use-case (using the consumed amount of something for a transport model) until we have other transport models and better understand whether this use-case is common enough to require a more polished integration.

If you haven't done so already, could you put a quick blurb in the PR description to explain why the test values changed (increased) for the  example 21 integration test?

PR #407, opened 2025-12-09
Enable load following optimization dispatch with Pyomo

genevievestarke:
#  Bring Pyomo load following dispatch capabilities into H2I

This feature would enable dispatch optimization using one storage technology and multiple generation technologies (using the combiner) to determine optimal storage operation to follow a load while minimizing costs. This feature adds the optimization problem formulation, framework, and solver for the optimization in Pyomo. To do this, a `hybrid_rule.py` file was added to aggregate the storage and production variables to create one problem for the optimizer. The added files are 
- `OptimizedDispatchController` class in `pyomo_controllers.py`
- The `hybrid_rule.py` file: Aggregates optimization pieces from the various components and formulates the model for the optimization. This file collects the objective function, initializes model parameters, and updates time series parameters.
- The `generic_converter_opt.py` file: Houses the variables, parameters, constraints, and objective function for the generator technology in the optimization model.
- The  `pyomo_storage_rule_min_operating_cost.py` file: houses the storage and system variables, parameters, constraints, and objective function. Ideally the system parameters would be in a separate file, perhaps, but have been combined for efficiency in the first implementation.
- The `controller_opt_problem_state.py` file: Formulates the optimization model for the solver

The below is an example of how and where the dispatch is run using a battery for electricity storage:
image

### What's changed?
For the optimized dispatch, the control and dispatch functions are no longer separate (see figure above). This also means that the controller needs access to the dispatch parameters. Now all dispatch and controller classes have access to the `shared_parameters` section in the input yaml, meaning that parameters like `commodity_name` that were defined in multiple places before should now be defined once in the `shared_parameters` section. This affects the pyomo heuristic dispatch, as well. Most control parameters are then defined in the `shared_parameters`, as well, because there is a large overlap with the performance variables. Control/dispatch specific variables can be defined in the `control_parameters` section. This include:
- `cost_per_charge`: in $/kW, cost to charge the storage (note that charging is incentivized)
- `cost_per_discharge`: in $/kW, cost to discharge the storage
- `commodity_met_value`: in $/kW, penalty for not meeting the desired load demand
- `cost_per_production`: in $/kW, cost to use the incoming produced commodity (i.e. electricity from wind)
- `time_weighting_factor`:  This parameter discounts each subsequent time step incrementally in the future in the horizon window by the defined amount (defaults to 0.995)
- `tech_name`: Name that the technology will be defined as in the dispatch

[...]

PR #440, opened 2026-01-08
Add csv post process for time series variables

genevievestarke:
# Add csv post process for time series variables

This pull requests adds a function to save specified time series variables to a csv file for post processing. The function is meant to only save time series variables from a single H2I run. It loads the data from an sql file, so the outputs of the H2I run need to be saved with a recorder. You can save all the time series variables from a run by not specifying specific variables, or you can enter a dictionary or list of the time series variables you would like to save. You can also provide units and/or alternative names for the variables. The output is a dictionary of the time series variables, and a csv file if the `save_to_file` input is set to `True`.

Additionally, there is a minor bug fix to return the recorder path.

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [x] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Documentation

### Section 4.1: New Files
- `h2integrate\postprocess\sql_timeseries_to_csv.py`: tests for all possible different use cases of the function `save_case_timeseries_as_csv`
  - `check_get_units_for_var`: Check the units for a variable within a case and returns the variable value in the specified units and the specified units.
  - `save_case_timeseries_as_csv`: Summarize timeseries data from a case within an sql recorder file to a DataFrame
    and save to csv file if `save_to_file` is True.
- `h2integrate\postprocess\test\test_sql_timeseries_to_csv.py`

- `h2integrate/core/pose_optimization.py`
   - `set_recorders()`: minor bugfix to return recorder path if `recorder_attachment=="model"`
- `examples/02_texas_ammonia/driver_config.yaml`: updated to user recorder
- `examples/02_texas_ammonia/run_texas_ammonia_plant.py`: updated to showcase saving timeseries functionality

### DIFF
--- a/h2integrate/postprocess/test/test_sql_timeseries_to_csv.py
+++ b/h2integrate/postprocess/test/test_sql_timeseries_to_csv.py
@@ -29,7 +29,7 @@ def test_save_csv_all_results(subtests, run_example_02_sql_fpath):
     res = save_case_timeseries_as_csv(run_example_02_sql_fpath, save_to_file=True)
 
     with subtests.test("Check number of columns"):
-        assert len(res.columns.to_list()) == 26
+        assert len(res.columns.to_list()) == 35
 
     with subtests.test("Check number of rows"):
         assert len(res) == 8760
@@ -41,7 +41,7 @@ def test_save_csv_all_results(subtests, run_example_02_sql_fpath):
 def test_make_df_from_varname_list(subtests, run_example_02_sql_fpath):
     vars_to_save = [
         "electrolyzer.hydrogen_out",
-        "hopp.electricity_out",
+        "combiner.electricity_out",
         "ammonia.ammonia_out",
         "h2_storage.hydrogen_out",
     ]
@@ -88,7 +88,7 @@ def test_make_df_from_varname_unit_dict(subtests, run_example_02_sql_fpath):
 def test_alternative_column_names(subtests, run_example_02_sql_fpath):
     vars_to_save = {
         "electrolyzer.hydrogen_out": {"alternative_name": "Electrolyzer Hydrogen Output"},
-        "hopp.electricity_out": {"units": "kW", "alternative_name": "Plant Electricity Output"},
+        "combiner.electricity_out": {"units": "kW", "alternative_name": "Plant Electricity Output"},
         "ammonia.ammonia_out": {"alternative_name": None},
         "h2_storage.hydrogen_out": {"alternative_name": "H2 Storage Hydrogen Output"},
     }


h2integrate/postprocess/timeseries_to_csv.py
```diff
@@ -0,0 +1,69 @@
...
+    # set the path for the recorder from stuff specified in the driver_config.yaml
+    fpath = Path.cwd() / "outputs" / "cases.sql"
+
+    # load the cases
+    cr = om.CaseReader(fpath)
+
+    # get the cases as a list
+    cases = list(cr.get_cases())
+
+    variable_list = [
```
elenya-grant: I think that "variable_list" and "units_list" may be easier to read if it was a dictionary. Like:
```python
variable_to_units_dict = {'battery.SOC': 'percent', 'battery.electricity_in': 'MW'} 
```

elenya-grant (changes requested):
This seems like a really great tool to save timeseries profiles! Thank you! I have a few suggestions and would be happy to chat more about other possible considerations. For example, 
- The input to the function could be the .sql filepath and the case index then the filename could be output as the same as the sql filename but includes the case number:
```python
output_fpath = sql_fpath.parent / f"{sql_fpath.stem}_Case{case_index}.csv"
```
- Would there be a case where someone would want to save all the timeseries data? We could make this function also able to save all the timeseries variables if a specified name list isn't input.

h2integrate/postprocess/sql_timeseries_to_csv.py
```diff
@@ -0,0 +1,228 @@
...
+        if var in var_to_values:
+            # don't duplicate data
+            continue
+
+        # get the value
+        val = case.get_val(var)
+
+        # Skip costs that are per year of plant life
+        if "varopex" in var.lower() or "annual_fixed_costs" in var.lower():
+            continue
```
johnjasa: Add a note about why we're skipping these to the comment? Not immediately clear why this logic is needed

h2integrate/postprocess/sql_timeseries_to_csv.py
```diff
@@ -0,0 +1,228 @@
...
+    var_list += [v["prom_name"] for v in input_var_dict.values()]
+    var_list.sort()
+
+    # if vars_to_save is not empty, then only include the variables in var_list
+    if bool(vars_to_save):
+        if isinstance(vars_to_save, dict):
+            varnames_to_save = list(vars_to_save.keys())
+            var_list = [v for v in var_list if v in varnames_to_save]
+        if isinstance(vars_to_save, list):
+            var_list = [v for v in var_list if v in vars_to_save]
```
johnjasa: This logic is pretty expansive in terms of which variables are included and I want to see if it's intentional. Let me explain what I mean.

With this in example 08:
```
# Get a subset of timeseries data and don't save it to a csv file
vars_to_save = [
    "electrolyzer.hydrogen_out",
    "wind.electricity_out",
]
timeseries_data = save_case_timeseries_as_csv(
    h2i_model.recorder_path, vars_to_save=vars_to_save, save_to_file=False
)
```

I get this as the csv:
```
electrolyzer.electricity_in (MW),electrolyzer.hydrogen_out (kg/h),finance_subgroup_electricity_custom.electricity_sum.electricity_wind (MW),finance_subgroup_electricity_profast.electricity_sum.electricity_wind (MW),wind.electricity_out (MW),wind_to_electrolyzer_cable.electricity_in (MW),wind_to_electrolyzer_cable.electricity_out (MW)
628.6673401626845,10196.570622701698,628.6673401626845,628.6673401626845,628.6673401626845,628.6673401626845,628.6673401626845
624.8427474743152,10196.554054672835,624.8427474743152,624.8427474743152,624.8427474743152,624.8427474743152,624.8427474743152
```
To be very explicit, because the wind electricity is an input to many sources (financial streams, electrolyzer, etc), redundant data appears in the .csv. This could be on purpose, but it might get confusing for users, especially in bigger models.

Could you please comment on if this is intended behavior and if so, motivate it a little bit? Thanks!

johnjasa (changes requested):
Fun PR! I'm really happy to see you, @genevievestarke, contributing to the code in nice, digestible chunks as you work through a bigger behemoth in #407.

I have some comments and questions that I've marked. The main quasi-blocking one is on if the number of outputs saved is intentional. I left more info in a bigger comment with an example output showing what I mean.

I made some small edits directly to the branch, too.

My other suggestion is to add at least a mention of this functionality to the docs (`postprocessing_results.md`). A bigger inclusion could show using it within an .ipynb file but that isn't strictly necessary.

Lastly, let me acknowledge that this is a more in-depth review than is purely needed here, but really because I want to give good feedback and make sure this is optimally useful for users. I love the addition and I think it helps answer the question "so I just ran H2I... now what?" so that's very helpful! Thank you!

examples/02_texas_ammonia/run_texas_ammonia_plant.py
```diff
@@ -1,10 +1,25 @@
...
+h2i.run()
 
-model.post_process()
+h2i.post_process()
+
+# Save all timeseries data to a csv file
+timeseries_data = save_case_timeseries_as_csv(h2i.recorder_path, save_to_file=True)
+
+# Get a subset of timeseries data and don't save it to a csv file
+vars_to_save = [
```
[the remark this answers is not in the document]
elenya-grant: Do you think we should change the variable `vars_to_save` to `vars_to_track`? Normally it'll save a csv file but a user could also just have it return the dataframe without saving it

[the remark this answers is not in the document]
johnjasa: Ah, I see the purpose of this call now. I think we can make this more clear. I'd suggest `output_vars` or something similar.

But, the larger issue is that the function `save_case_timeseries_as_csv` doesn't always save to a .csv file (as we're showing in this example call). That's a tough thing to accept because "save" and "csv" are in the name so as a user I'd expect it to save to a .csv. I'd perhaps suggest `save_case_timeseries` or `process_case_timeseries` if you want to retain that functionality but make the name always true. Or maybe it could be as simple as adding some comments to this example to say "In the event you don't to save it to a .csv but you're still interested in the timeseries dataframe, here's how to get it".

h2integrate/postprocess/sql_timeseries_to_csv.py
johnjasa: We're chatting right now! This is resolved and I like the current behavior.

johnjasa (approved):
Looks great, thanks for this meaningful quality-of-life improvement to H2I, @genevievestarke! I really like the functionality this adds and I appreciate the effort you put into iterating on it.

I pushed up some doc changes based on Elenya's last non-block suggestion. Great to come in and I'll merge it after tests pass!

Issue #485, opened 2026-01-29
Update ProFAST finance models to use utilization instead of capacity (1 of 3)


elenya-grant:
This issue is intended to capture a follow-on PR to PR #463

Currently, the ProFAST finances models represent the annual commodity production by setting the 'capacity' parameter based on the actual commodity production and the long term utilization as 1. For models that account for variable commodity production over the lifetime (like the electrolyzer model), this is not accurately represented in the finance results. This can only be represented in the finance results by setting the long term utilization to the capacity factor of the tech and setting the capacity as the actual tech's rated production capacity.

**Timeline for this**: 3-10 days after PR #463 is merged. 

Note: May result in changed test values depending on whether the electrolyzer model outputs the life average capacity factor vs the capacity factor per year of the plant life (see PR #491)

### Benefits of this development:
- Allow for lifetime performance to be reflected in financials
- Reduction in amount of hard-coded tech-specific and/or commodity-specific logic in `H2IntegrateModel`
- Lays the groundwork to potentially fully remove hard-coded tech-specific and/or commodity-specific logic in `H2IntegrateModel`. This would also allow for more consistent logic when commodity streams are not defined and not rely on the use of summer components.
- Removal of tech-specific hard-coded logic in the finance models (all technologies in the finance subgroup would pass the replacement schedule to the finance model)
- Allows more flexibility in what commodities and/or commodity streams may be connected to the finance models

### Background
Currently, the only enforced naming convention in H2I is the use of `"commodity_in"` and `"commodity_out"` for models. The `commodity_out` stream is connected to a `summer` model, which sums the timeseries profile over the length of the simulation. The outputs of these summers are only accurate if a simulation is one year and has an hourly timestep. The output of the summer is connected to the finance models as `total_commodity_produced`, which sets the following values ProFAST:
```python
capacity = inputs[f"total_{self.options['commodity_type']}_produced"][0] / 365.0
profast_params["capacity"] = capacity
profast_params["long term utilization"] = 1 
```

The `capacity` parameter cannot be set as an array to account for varying production over the lifetime of the plant. The `long term utilization` parameter (equivalent to capacity factor) can account for varying production over the lifetime of the plant. To allow for accurate accounting of annual variation of commodity production that some technology models are capable of estimating, we would have to move towards a framework where the ProFAST parameters are set as:
```python
# create years of operation list
years_of_operation = create_years_of_operation(
            self.params.plant_life,
            self.params.analysis_start_year,
[...]

PR #523, opened 2026-02-11
Standardize Feedstock Outputs (follow-on to 463)

elenya-grant:
# Standardize Feedstock Outputs (follow-on to 463)

Updated feedstock component to have standard outputs (like those output from the `PerformanceModelBaseClass`). Feedstock model now outputs `capacity_factor`, `annual_commodity_consumed`, and `total_commodity_consumed`. Note that these are output from the FeedstockCostModel, which is due to the special handling of feedstock components.

This PR was re-opened because it seemed relevant for PR #605 so that non-electricity commodities could also be used with the updated dispatch method that is introduced in PR #605.

**This PR is ready for high-level feedback, the questions for reviewers are noted in Section 2!**

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [ ] depends on reviewer feedback.
- [x] update feedstocks.md
- [x] add test of feedstock component integrated with finance model (by adding it to an example)

**Structural feedback**:
- Is there any reason why the feedstock standard outputs be following the "produced" naming instead of "consumed"? I think that it makes sense for feedstock standard outputs to be "consumed". 

**Implementation feedback**:

**Other feedback**:

- [ ] Documentation

## Section 3: Related Issues

Made Issue #659 when creating this PR, this issue is not resolved by this PR

Resolves part of Issue #485 about updates feedstock to have standardized performance

### Section 4.1: New Files
None

- `h2integrate/core/feedstocks.py`:
  - `FeedstockCostModel`: 
     - added _input_ of `commodity_out`, which is used as the denominator in the capacity factor calculation
     - added outputs of `capacity_factor`, `annual_commodity_consumed`, and `total_commodity_consumed`
-  `examples/test/test_all_examples.py::test_natural_gas_example` added subtest for feedstock capacity factor
- `h2integrate/core/test/test_feedstocks.py`: added new test `test_feedstock_standard_outputs`
   - `H2IntegrateModel.connect_technologies()`: added connection to connect `commodity_out` output of the feedstock performance model to the input  `commodity_out` of the feedstock cost model (this was needed because  the `FeedstockCostModel` and `FeedstockPerformanceModel` are not part of the same sub-system (they don't have shared inputs/outputs - they're connected using connections, not promoted inputs/outputs))
- `examples/16_natural_gas/tech_config.yaml`: added some comments for natural gas feedstock
- `h2integrate/tools/profast_tools.py`: 
   - `make_price_breakdown()`: updated so it can handle cases with a zero total capex.
- `examples/12_ammonia_synloop/`
   - `tech_config.yaml`: updated `n2_feedstock` price to be $5.0/t (instead of 0) or to prevent errors when trying to run profast with zero costs
   - `plant_config.yaml`: added `n2` finance subgroup for integration testing purposes. Removed `n2_feedstock` from the `nh3` technology group so that LCOA doesn't change in tests (before the n2_feedstock had zero cost, but the update to the n2_feedstock price would give a larger LCOA if included). Could instead update the test value if desired.
- `pytest examples/test/test_all_examples.py::test_ammonia_synloop_example`: added subtest for LCON 
 
N/A


h2integrate/core/feedstocks.py
```diff
@@ -33,39 +33,55 @@ def setup(self):
...
             additional_cls_name=self.__class__.__name__,
         )
-        n_timesteps = self.options["plant_config"]["plant"]["simulation"]["n_timesteps"]
-        feedstock_type = self.config.feedstock_type
+        self.n_timesteps = int(self.options["plant_config"]["plant"]["simulation"]["n_timesteps"])
+        self.commodity = self.config.commodity
+        self.commodity_rate_units = self.config.commodity_rate_units
 
-        self.add_output(f"{feedstock_type}_out", shape=n_timesteps, units=self.config.units)
+        # NOTE: should below be renamed to f"{self.commodity}_capacity"?
```
johnjasa: I'd support that change, it seems more clear

johnjasa:
> Is there any reason why the feedstock standard outputs be following the "produced" naming instead of "consumed"? I think that it makes sense for feedstock standard outputs to be "consumed"

Stick with consumed imo. That's more clear given how we're handling feedstock quantities.

h2integrate/core/feedstocks.py
```diff
@@ -74,23 +89,92 @@ def setup(self):
...
+        )
+
+        # Estimate annual consumption based on consumption over the simulation
+        # NOTE: once we standardize feedstock consumption outputs in models, this should
+        # be updated to handle consumption that varies over years of operation
+        outputs[f"annual_{self.commodity}_consumed"] = outputs[
+            f"total_{self.commodity}_consumed"
+        ] * (1 / self.fraction_of_year_simulated)
+
+        outputs[f"rated_{self.commodity}_production"] = inputs[f"{self.commodity}_out"].max()
```
johnjasa: I know I brought this up on a call, but I forget where we landed. Will this assumption (that the rated production is the max of the commodity stream) be limiting in any way? Would this go against users' intuition at some point?

h2integrate/core/feedstocks.py
```diff
@@ -74,23 +89,92 @@ def setup(self):
...
+            val=0,
+            units=self.commodity_rate_units,
         )
 
     def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
-        feedstock_type = self.config.feedstock_type
-        price = self.config.price
-        hourly_consumption = inputs[f"{feedstock_type}_consumed"]
+        # Calculate performance based on consumption
+
```
johnjasa: Suggest removing comment as it isn't necessarily helpful

johnjasa (changes requested):
Thanks for this, Elenya! I like the thought you put into the feedstocks setup and how it could be more standard to match other components. I also appreciate you fixing the NG units to be consistently correct.

My blocking request is for you to update `feedstocks.md` as well. Could you please change that doc so it matches what you've done here, then also add a few words to explain the CF calculation and anything else you introduce here?

h2integrate/core/feedstocks.py
elenya-grant: I don't think so - the inputs[f"commodity_out"] is the output from the performance model, which outputs the commodity_out as an array of all the same values. Aka - the max should equal the mean which should equal the "commodity_capacity" input to the performance model.

h2integrate/core/feedstocks.py
elenya-grant: done

h2integrate/core/test/test_feedstocks.py
```diff
@@ -8,10 +8,76 @@
...
+            "simulation": {
+                "n_timesteps": 8760,
+                "dt": 3600,
+            },
+        },
+    }
+
+
+@fixture
+def ng_tech_input_config():
```
kbrunik: Could you change this to `ng_feedstock_input_config`? I just got a little confused when I thought it was maybe for a natural gas plant

h2integrate/core/test/test_feedstocks.py
```diff
@@ -8,10 +8,76 @@
...
+
+    prob.setup()
+    # Set some consumption values
+    consumption = np.full(8760, 50.0)  # 50 MMBtu/hour
+    prob.set_val("ng_feedstock.natural_gas_consumed", consumption)
+    prob.run_model()
+    with subtests.test("Check feedstock capacity factor"):
+        ng_cf = prob.get_val("ng_feedstock.capacity_factor", units="unitless").mean()
+        assert pytest.approx(ng_cf, rel=1e-6) == 0.5
+    # TODO: add subtests for rated_natural_gas_production, total_natural_gas_consumed,
```
kbrunik: Finish TODO?

kbrunik (changes requested):
I agree with @johnjasa, it would be good to explain the CF calculation, and the other output performance values (and maybe that it's called consumed instead of produced) in the feedstocks docpage. Otherwise I think this looks really good! Thanks for doing this update, it really helps add continuity throughout the models.

Would you also be able to update the changelog?

h2integrate/core/test/test_feedstocks.py
elenya-grant: done!

h2integrate/core/test/test_feedstocks.py
elenya-grant: done!

elenya-grant:
@kbrunik - I updated the doc page but would love your feedback on it! I also updated the changeling and addressed your other comments!

kbrunik (approved):
Looks great to me! Thanks for adding an explanation about the outputs :)

johnjasa (approved):
Nice, I love this generalization and improvement to the feedstock handling! I think especially as we get to more complex multi-commodity systems, this will become even more important. Thank you for also updating example 12 so that there's an LCON computed; it made the capabilities afforded through this PR very clear!

PR #549, opened 2026-02-25
Split out Pyomo Controllers to separate files

elenya-grant:
# Split out Pyomo Controllers to separate files

Split out the controllers in `h2integrate/control/control_strategies/pyomo_controllers.py` to separate files and removed some inheritance so that the logic is easier to follow. Also refactored the `HeuristicLoadFollowingController` to no longer inherit the `SimpleBatteryControllerHeuristic` class (the `SimpleBatteryControllerHeuristic` has been removed and the functionality now exists in the `HeuristicLoadFollowingController`).

These changes are intended to make further development of the pyomo controllers easier by making the individual control strategies easier to understand through the removal of inheritance and being able to look at a single file containing the entire control strategy of interest (the individual files also make it easier to determine which attributes are being used and which ones aren't).

Removed `n_horizon_window` from the `PyomoControllerBaseConfig` and `PySAMBatteryPerformanceModelConfig` because it was unused.

Follow-on work may look at refactoring the `PyomoControllerBaseConfig`, `HeuristicLoadFollowingControllerConfig`, and `OptimizedDispatchControllerConfig` to remove unused attributes. 

## Section 1: Type of Contribution

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [x] Other (please describe): clean-up

## Section 2: Draft PR Checklist

- [ ] Open draft PR
- [ ] Describe the feature that will be added
- [ ] Fill out TODO list steps
- [ ] Describe requested feedback from reviewers on draft PR
- [ ] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [ ] Step 1
- [ ] Step 2

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Added tests for new functionality or bug fixes
- [x] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [x] Examples have been updated (if applicable)
[...]

Issue #521, opened 2026-02-10
Reconsider storage performance model outputs


elenya-grant:
# Reconsider storage performance model outputs
This came up somewhat during PR #493 

This is somewhat related to Issue #498, and was briefly mentioned in [this comment](https://github.com/NatLabRockies/H2Integrate/issues/485#issuecomment-3844107510) on Issue #485 which says:
> the battery capacity factor and rated capacity will not result in the same "battery production" as using "battery.electricity_out" because "battery.electricity_out" is equal to: `electricity_in - electricity_for_charging + electricity_discharged - excess_electricity` whereas the battery capacity factor and rated capacity are based only on electricity_discharged. This would only be a concern where the battery is specified as the commodity stream (for a finance subgroup). 

I'll stick to using the battery for this example, but this applies to any storage component. _Control strategies for storage components_ (currently, this is only the `DemandOpenLoopStorageController` model) have inputs of:
- `commodity_in`
- `commodity_demand`

and outputs of:
- `commodity_unmet_demand`
- `commodity_unused_commodity`
- `commodity_set_point`
- `total_commodity_unmet_demand`

and the storage _performance_ models have an input of `commodity_set_point` and output the standard outputs of the `PerformanceModelBaseClass` such as
- `commodity_out`
- `rated_commodity_production`
- `total_commodity_produced`
- `annual_commodity_produced`
- `capacity_factor`

In the case of using the `DemandOpenLoopStorageController` with the `PerformanceModelBaseClass` (which is done in 6 examples) and perhaps instances of using Pyomo dispatch controllers with the PySAM battery, there will be "issues" when using the storage component as the `commodity_stream` in a finance subgroup (done in Example 14, 18, 19, and 29) because the battery capacity factor and rated capacity are not as "tied" to the "electricity_out" of the battery as is the case for converter performance models. The battery capacity factor and rated capacity are based on the power discharged from the battery. The `electricity_out` of the battery is the `electricity_in - battery_charged + battery_discharged - excess_electricity`. 

For most cases where a converter is defined as the `commodity_stream`, the finance subgroup results should not change if we use `capacity_factor` and `rated_commodity_production` instead of `sum_comp.total_commodity_produced`. This is not the case for storage, because the capacity factor and rated production are calculated on the commodity discharged not the `commodity_out`.

I think that we need to:
1. standardize what's passed from storage control strategies to storage performance models (Issue #498)
[...]

PR #557, opened 2026-02-26
Update to PySAM Battery so size is changed in compute()

elenya-grant:
# Update to PySAM Battery so size is changed in compute()

Updated the `PySAMBatteryPerformanceModel` so that the input charge rate and storage capacity are used in `compute()` and `simulate()` instead of accessing those values from the config. Other changes include:
- removed the use of `BatteryOutputs`, removed `P_chargeable` and `P_dischargeable` from outputs (they don't appear to be used)
- Added outputs of `battery_charge` and `battery_discharge`: (these are planned to be used in updated tests)
- Removed unused or unnecessary attributes of the `PySAMBatteryPerformanceModel`
- Removed hard-coded numbers and replaced with attributes in the `PySAMBatteryPerformanceModelConfig`

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR
- [-] Describe requested feedback from reviewers on draft PR

### TODO:
- [x] Add test to check that battery capacity is changed after `setup()` (one test with a pyomo controller, one test without a controller)
   - [x] without controller
   - [x] with controller (may requires updates to `HeuristicLoadFollowingController` and `PyomoRuleStorageBaseclass`, wait unit PR #549 is merged)
- [x] Remove commented out lines of code in `pysam_battery.py`
- [x] Update docstrings in `pysam_battery.py`

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation

## Section 3: Related Issues

Related to #521 
This addresses the first step outlined in Issue #564

### Section 4.1: New Files
N/A

- `h2integrate/storage/battery/pysam_battery.py`
  - `PySAMBatteryPerformanceModelConfig`: What and why something was changed in one sentence or less.
  - `PySAMBatteryPerformanceModel`
     - `setup()`: removed `P_chargeable` and `P_dischargeable` outputs, added `battery_charge` and `battery_discharge` outputs.
     - `compute()`: removed usage of `BatteryOutputs`, updated to use the inputs of charge rate and storage capacity rather than accessing those values from the config.
     - `simulate()`: removed usage of `BatteryOutputs`, updated to use the inputs of charge rate and storage capacity rather than accessing those values from the config.
- `h2integrate/storage/battery/battery_baseclass.py`: updated the input/output setting to use self.n_timesteps instead of copy_shape
- `h2integrate/storage/battery/test/test_pysam_battery.py`: added a new test (`test_pysam_battery_no_controller_change_capacity`) to check that charge rate is changed.
- `h2integrate/control/test/test_heuristic_controllers.py`: updated to use pytest fixtures and added new test `test_heuristic_load_following_battery_dispatch_change_capacities` to test that the results are as-expected if the battery capacity is changed from the value in the tech config. 

The below two files were changed so that the storage capacity is updated in the heuristic controller in the compute() method
- `h2integrate/control/control_strategies/heuristic_pyomo_controller.py`: added `storage_capacity` as an input, added `inputs` and as an input to the `initialize_parameters()` method.
- `h2integrate/control/control_strategies/pyomo_controller_baseclass.py`: updated call to `initialize_parameters()` to input the `inputs`.

The below two files were changed to have some minor clean-ups and removing hard-coded numbers:
- `h2integrate/control/control_rules/pyomo_rule_baseclass.py`: removed the initialization of the config attribute in the `setup()` method
- `h2integrate/control/control_rules/storage/pyomo_storage_rule_baseclass.py`: added a `PyomoStorageRuleBaseConfig` which includes attributes that were originally hard-coded values. Now some of the Pyomo parameters set in the `_create_parameters()` method use values from the config rather than hard-coded values. The attributes of the config have default values corresponding to the previous hard-coded values to avoid having to modify more files.
- `h2integrate/control/control_rules/converters/generic_converter.py`: updated to import the `PyomoRuleBaseConfig` and create it in the `setup()` method

Added a new test `test_heuristic_load_following_battery_dispatch_change_capacities` which was failing. The failure was because the storage charge rate and storage capacity are not inputs to the `HeuristicLoadFollowingController` and `PyomoRuleStorageBaseclass`. These were updated to take in the capacities as inputs once PR #549 was merged in.


h2integrate/storage/battery/battery_baseclass.py
```diff
@@ -14,23 +14,23 @@ def setup(self):
...
+            shape=self.n_timesteps,
             units="percent",
             desc="State of charge of Battery",
         )
 
         self.add_output(
             "battery_electricity_discharge",
             val=0.0,
-            copy_shape="electricity_in",
+            shape=self.n_timesteps,
```
jaredthomas68: I think the shape should still come from "electricity_in" to make it clear that all these need to be the same length.

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -396,42 +341,47 @@ def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
...
-                for i in range(n_time_steps)
+                for i in range(self.n_timesteps)
             ]
-            unmet_demand = self.outputs.unmet_demand
 
         outputs["unmet_electricity_demand_out"] = unmet_demand
         outputs["unused_electricity_out"] = unused_commodity
         outputs["battery_electricity_discharge"] = battery_power_out
+
+        outputs["battery_charge"] = np.where(battery_power_out < 0, battery_power_out, 0)
```
jaredthomas68: battery power out is enforced positive at https://github.com/NatLabRockies/H2Integrate/pull/557/changes#diff-e26e294957707bea692f2ade6ab862d02dde994cabd452fc0f590c996d46db6cL405

I think you meant to use battery_power here

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -506,29 +459,19 @@ def simulate(
...
             # according to simulation
             max_chargeable_1 = np.maximum(0, -self.system_model.value("P_chargeable"))
             # according to soc
-            max_chargeable_2 = np.maximum(
-                0, (soc_max - soc) * self.config.max_capacity / self.dt_hr
-            )
+            max_chargeable_2 = np.maximum(0, (soc_max - soc) * storage_capacity / self.dt_hr)
             # compare all versions of max_chargeable
-            max_chargeable = np.min([max_chargeable_0, max_chargeable_1, max_chargeable_2])
+            max_chargeable = np.min([charge_rate, max_chargeable_1, max_chargeable_2])
```
jaredthomas68: I would prefer to leave this as it was for clarity, that we are checking specifications, simulation, and soc. I know it looks redundant, but I think it helps the reader.

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -506,29 +459,19 @@ def simulate(
...
             # according to soc
-            max_dischargeable_2 = np.maximum(
-                0, (soc - soc_min) * self.config.max_capacity / self.dt_hr
-            )
+            max_dischargeable_2 = np.maximum(0, (soc - soc_min) * storage_capacity / self.dt_hr)
             # compare all versions of max_dischargeable
-            max_dischargeable = np.min(
-                [max_dischargeable_0, max_dischargeable_1, max_dischargeable_2]
-            )
+            max_dischargeable = np.min([discharge_rate, max_dischargeable_1, max_dischargeable_2])
```
jaredthomas68: same comment as above. The way it was is more clear to me.

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -549,17 +482,6 @@ def simulate(
...
-            # scraped from the PySAM model modules `StatePack` and `StateCell`.
-            for attr in self.outputs.stateful_attributes:
-                if hasattr(self.system_model.StatePack, attr) or hasattr(
-                    self.system_model.StateCell, attr
-                ):
-                    getattr(self.outputs, attr)[sim_start_index + t] = self.system_model.value(attr)
-
-            for attr in self.outputs.component_attributes:
-                getattr(self.outputs, attr)[sim_start_index + t] = getattr(self, attr)
-
```
jaredthomas68, line 559: @bayc I'm curious what you think of removing this and the related data class. It is not being used right now and was originally inserted to match what was done in HOPP, but I don't think we actually need in H2I. However, if we were interested in using more information about the battery at some point then we would need something like this.

jaredthomas68 (changes requested):
I think there are a few things that need discussion with others, specifically @bayc.

h2integrate/storage/battery/pysam_battery.py
elenya-grant: this is a new output that I've added. `battery_power_out` is not enforced to be positive, this is highlighted in `h2integrate/control/test/test_heuristic_controllers.py::test_heuristic_load_following_battery_dispatch`, the `expected_battery_electricity_discharge` variable has negative values and this array is testing the battery output `battery_electricity_discharge`, which is set equal to `battery_power_out`. `battery_power_out` is negative when the battery is charged, and positive when the battery is discharged. 

you're right that the `battery_power_out` does not have negative values (saturated at a minimum value of 0) **_when_** `"pyomo_dispatch_solver"` is **NOT** in `discrete_inputs`, meaning that the battery never charges in this case, which seems wrong to me (I didn't write that bit of code, just updated it to use arrays instead of list comprehension)

Aka - _with a controller_, the `battery_power_out` is not enforced to be positive, but _without a controller_, the logic does force the `battery_power_out` to be positive which indicates that without a controller, we never charge the battery.

h2integrate/storage/battery/pysam_battery.py
elenya-grant: okey dokey - I changed it back

h2integrate/storage/battery/pysam_battery.py
elenya-grant: okey dokey - I changed it back

h2integrate/storage/battery/pysam_battery.py
elenya-grant: Edit - I now understand what you're saying.

I think that the below line (which is in the case without a controller) should be deleted:

```python
battery_power_out = np.max(
                [np.array(battery_power), np.zeros(self.n_timesteps)], axis=0
            )
```

 and the output of the `self.simulate()` function should be changed to `battery_power_out` instead of `battery_power` (also in the case without a controller).

This change would likely change some test values in `h2integrate/storage/battery/test/test_pysam_battery.py`

elenya-grant: I just made the change I proposed above and would like your thoughts on it

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -382,53 +315,63 @@ def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
...
+            unmet_demand = [
                 np.max([0, inputs["electricity_demand"][i] - combined_power_out[i]])
-                for i in range(n_time_steps)
+                for i in range(self.n_timesteps)
             ]
-            unmet_demand = self.outputs.unmet_demand
 
         outputs["unmet_electricity_demand_out"] = unmet_demand
         outputs["unused_electricity_out"] = unused_commodity
         outputs["battery_electricity_discharge"] = battery_power_out
```
jaredthomas68: The intent of this output was to have it be the strictly positive, output only, from the battery, the same as the added "battery_discharge". Our typical approach in H2I is to specify the commodity, so I would prefer to have "electricity" in the name still.

h2integrate/storage/battery/pysam_battery.py
jaredthomas68: The reason for the original names, I think, was that `battery_power was positive and negative for discharge and charge, while `battery_power_out` was just positive and just discharge. I think using `battery_power_out` with positive and negative is confusing. I'm mostly ok with the changes you made, but let's use the name `battery_power` or `battery_electricity` and drop the `out` if it is both positive and negative now.

jaredthomas68: I also think it would be more clear to name the outputs `battery_electricity_discharge` and `battery_electricity_charge` for now, though I would like to change these again once we get the control off the storage tech.

elenya-grant: Okay so are you wanting the output `battery_electricity_discharge` to now be named something else?

like - is this what you're wanting:
```python
outputs["battery_electricity_discharge"] = np.where(battery_power_out > 0, battery_power_out, 0)
outputs["battery_electricity_charge"] = np.where(battery_power_out < 0, battery_power_out, 0)
outputs["battery_electricity_out"] = battery_power_out 
```

or a different name than `battery_electricity_out` (cause you said to drop the "out"). So - like output it as `battery_electricity`?

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -382,53 +315,63 @@ def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
...
+                storage_capacity=inputs["storage_capacity"][0],
                 control_variable=self.config.control_variable,
             )
-            n_time_steps = len(inputs["electricity_demand"])
 
-            # determine battery discharge
-            self.outputs.P = battery_power
-            battery_power_out = [np.max([0, battery_power[i]]) for i in range(n_time_steps)]
+            # battery_power_out is positive when the battery is discharged
+            # and positive when the battery is charged
```
genevievestarke: These two are the same, I think?

h2integrate/storage/battery/pysam_battery.py
genevievestarke: I think we can remove this and document this code in an issue so that we can add it back, if needed. Alternatively, we could explicitly call out that this is not used in the actual calculations (similar to the optimization metrics tracking code).

elenya-grant: made issue #565

h2integrate/storage/battery/pysam_battery.py
elenya-grant: I've responded to this comment (or a similar one) [here](https://github.com/NatLabRockies/H2Integrate/pull/557#discussion_r2880533754)

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -50,10 +57,14 @@ def setup(self):
         if self.config.discharge_efficiency is not None:
             self.discharge_efficiency = self.config.discharge_efficiency
 
-    def initialize_parameters(self):
+    def initialize_parameters(self, inputs):
         """Initializes parameters."""
+
         self.minimum_storage = 0.0
-        self.maximum_storage = self.config.max_capacity
+        if "storage_capacity" in inputs:
```
jaredthomas68: Why the check here, it looks like storage_capacity is enforced to be in inputs in `setup()`

h2integrate/control/test/test_heuristic_controllers.py
```diff
@@ -11,64 +12,66 @@
...
-        "plant_life": 30,
-        "grid_connection": False,
-        "ppa_price": 0.025,
-        "hybrid_electricity_estimated_cf": 0.492,
-        "simulation": {
-            "dt": 3600,
-            "n_timesteps": 8760,
+@fixture
+def plant_config():
+    plant_config = {
```
jaredthomas68: Can we use different names for the methods and return values? like `get_plant_config` or something?

h2integrate/control/test/test_heuristic_controllers.py
```diff
@@ -11,64 +12,66 @@
...
+        "tech_to_dispatch_connections": [
+            ["battery", "battery"],
+        ],
+    }
+    return plant_config
+
+
+@fixture
+def tech_config():
+    tech_config = {
```
jaredthomas68: Same naming concern as on plant_config

h2integrate/control/test/test_heuristic_controllers.py
```diff
@@ -404,3 +408,256 @@ def test_heuristic_load_following_battery_dispatch(subtests):
...
+    with subtests.test("Check unmet_demand"):
+        assert (
+            pytest.approx(expected_unmet_demand_out, abs=1e-4)
+            == prob.get_val("battery.unmet_electricity_demand_out", units="kW")[0:24]
+        )
+
+    with subtests.test("Check unused_electricity_out"):
+        assert (
+            pytest.approx(expected_unused_commodity_out)
+            == prob.get_val("battery.unused_electricity_out", units="kW")[0:24]
```
jaredthomas68, line 662: Thanks for the additional tests

h2integrate/storage/battery/test/inputs/tech_config.yaml
```diff
@@ -19,7 +19,7 @@ technologies:
         max_charge_percent: 0.9
         min_charge_percent: 0.1
       performance_parameters:
-        system_model_source: pysam
+        # system_model_source: pysam
```
jaredthomas68: I think we can drop this line now. It was a holdover from HOPP swapping between pysam and hopp-owned LDES. The pysam battery just does pysam I think, so this should probably be dropped.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -50,10 +57,14 @@ def setup(self):
...
+            self.maximum_storage = inputs["storage_capacity"][0]
+        else:
+            self.maximum_storage = self.config.max_capacity
```
johnjasa:
```suggestion
        self.maximum_storage = inputs["storage_capacity"][0]
```

Because `storage_capacity` is added as an input above, this logic would _always_ be true. I'm suggesting this change which shouldn't change the behavior at all.

h2integrate/storage/battery/test/inputs/tech_config.yaml
johnjasa: Remove? or necessary and helpful to retain this as a commented line?

jaredthomas68: @bayc and @genevievestarke is there still interest in bringing over LDES from HOPP? If so, I think it should probably be a separate model at this point. What do you think?

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -11,63 +8,6 @@
 from h2integrate.storage.battery.battery_baseclass import BatteryPerformanceBaseClass
 
 
-@dataclass
-class BatteryOutputs:
```
johnjasa, line 15: Love this simplification and removal here! Makes the pysam battery setup much more H2I-esque.

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -382,53 +315,63 @@ def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
...
+                for i in range(self.n_timesteps)
             ]
-            unused_commodity = self.outputs.unused_commodity
 
             # determine how much demand was not met
-            self.outputs.unmet_demand = [
+            unmet_demand = [
                 np.max([0, inputs["electricity_demand"][i] - combined_power_out[i]])
-                for i in range(n_time_steps)
+                for i in range(self.n_timesteps)
```
johnjasa:
```suggestion
            # determine how much of the inflow electricity was unused
            unused_commodity = np.maximum(0, combined_power_out - inputs["electricity_demand"])

            # determine how much demand was not met
            unmet_demand = np.maximum(0, inputs["electricity_demand"] - combined_power_out)
```

h2integrate/storage/battery/pysam_battery.py
johnjasa: Big fan of this removal! Thanks Elenya for doing and others for weighing in.

johnjasa (changes requested):
I like this PR! Thank you for it. Really nice simplifications throughout and expansion of capabilities to optimize battery capacity.

Please see some small code suggestions I have an also please add to the changelog. Lastly, in the PR body there are some more TODOs. Have these been addressed in this PR already, are you still working on changes for this PR, or are those future tasks? Please check off those items or make issues for future tasks appropriately. Thanks!

h2integrate/storage/battery/pysam_battery.py
```diff
@@ -382,53 +315,63 @@ def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
...
             # and requests charge when input exceeds demand
 
+            # update the control window to be the number of timesteps in the simulation
+            self.config.n_control_window = self.n_timesteps
+
             # estimate required dispatch commands
             pseudo_commands = inputs["electricity_demand"] - inputs["electricity_in"]
 
-            battery_power, soc = self.simulate(
+            battery_power_out, soc = self.simulate(
```
jaredthomas68: Because this is power in and out, I think we should drop the `out`

h2integrate/storage/battery/pysam_battery.py
jaredthomas68: Close. I think I would do either:

```python
outputs["battery_electricity_discharge"] = np.where(battery_power_out > 0, battery_power_out, 0)
outputs["battery_electricity_charge"] = np.where(battery_power_out < 0, battery_power_out, 0)
```

or

```python
outputs["battery_electricity"] = battery_power_out 
```

But not both, since the two sets carry exactly the same information.

jaredthomas68 (changes requested):
Very close, just a few minor changes I'd like to see before this is brought in.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
elenya-grant: well - this is only used with the pysam battery at the moment so - yes. But I think we should remove this logic once I've created a storage model baseclass (outlined in Issue #564). I can remove it now though - I don't have strong feelings about it either way

h2integrate/storage/battery/test/inputs/tech_config.yaml
elenya-grant: I removed it!

h2integrate/storage/battery/pysam_battery.py
elenya-grant: updated it!

h2integrate/control/test/test_heuristic_controllers.py
elenya-grant: the fixture name is what's used in the test, so if I changed this to `def get_plant_config` then it'd look like this in the test: 
```python
HeuristicLoadFollowingController(plant_config=get_plant_config)
```

I can rename the plant config dictionary within the plant config fixture to have a different name though.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
elenya-grant: I updated to it your suggestion because @jaredthomas68 had a similar comment

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
elenya-grant: done - removed check.

elenya-grant:
@jaredthomas68 

Responding to your comment [here](https://github.com/NatLabRockies/H2Integrate/pull/557#discussion_r2886525462)

I think it's good to output both for now, I am about to open a PR for a generic storage model that can be used with the Pyomo controllers and I have a lot of subtests that check all 3 of these profiles.

I think that the naming doesn't need the technology name, but I can't think of a better way around it. I can change it to your suggestion but know that it may be likely to change in the future (as resolving Issue #564 progresses). I think further discussion on the output naming of storage components can be noted in Issue #521. I will have a PR up shortly that introduces a generic storage model and I think in that case, we will have a better idea of how outputs should be named across the storage models.

Update: I took your naming suggestion for the time being but included all 3 outputs. This can be easily changed in future PRs but I'd like to hold onto these until we're further along in resolving the steps laid out in Issue #564

h2integrate/control/test/test_heuristic_controllers.py
elenya-grant: done!

h2integrate/control/test/test_heuristic_controllers.py
elenya-grant: I renamed the dictionary to be `plant_config_dict`

h2integrate/storage/battery/pysam_battery.py
elenya-grant: fixed!

johnjasa (approved):
Thanks for the changes! I'll let Jared and others continue to weigh in. Make sure to add to the changelog too! :O

jaredthomas68 (approved):
I think this looks ready to come in. Thank you for your iterations on this @elenya-grant

jaredthomas68:
> @jaredthomas68
> 
> Responding to your comment [here](https://github.com/NatLabRockies/H2Integrate/pull/557#discussion_r2886525462)
> 
> I think it's good to output both for now, I am about to open a PR for a generic storage model that can be used with the Pyomo controllers and I have a lot of subtests that check all 3 of these profiles.
> 
> I think that the naming doesn't need the technology name, but I can't think of a better way around it. I can change it to your suggestion but know that it may be likely to change in the future (as resolving Issue #564 progresses). I think further discussion on the output naming of storage components can be noted in Issue #521. I will have a PR up shortly that introduces a generic storage model and I think in that case, we will have a better idea of how outputs should be named across the storage models.
> 
> Update: I took your naming suggestion for the time being but included all 3 outputs. This can be easily changed in future PRs but I'd like to hold onto these until we're further along in resolving the steps laid out in Issue #564

Got it. I'm ok with keeping the duplicate output information for now. Keeping the battery name in the charge/discharge outputs is important to me given the other `electricity_out` output. If we get to where that is no longer an issue, then it should be fine to drop the `battery` part.

Issue #441, opened 2026-01-09
Feature request: Add optimized dispatch option to generic storage model


genevievestarke:
#Add optimized dispatch option to generic storage model

Adding the option for pyomo optimized storage to the generic storage model would enable the optimized dispatch of other forms of storage beyond electricity, such as hydrogen. This would require adding in a pyomo solver option to the compute function of the generic storage model. This might change the inputs and outputs of the generic storage model, so it may be better to add a separate model to handle this?

Issue #564, opened 2026-03-03
Planned steps for refactoring existing dispatch and storage models


elenya-grant:
# Steps towards refactoring storage dispatch

The over-all goal of this work is **to allow for all storage performance models to be usable with all control strategies (pyomo or open-loop)**. The supplementary goals are:
- sync the open-loop storage controllers with the pyomo controllers
- create a storage model performance baseclass so that the storage model is compatible with open-loop controllers and pyomo controllers
- Ensure that all storage and control models can have their charge rate and capacity optimized

Some of the steps of this are detailed across many issues, but I will summarize them here

- [x] 1. Initial clean-up PySAM battery model so that charge rate and capacity can be optimized (PR #557)
- [x] 2. Introduce a generic storage model that is compatible with the Pyomo controllers (Issue #441)
   - PR #571 
- [x] 3. Split out some of the logic in the `PyomoControllerBaseClass` that is only used for the `HeuristicLoadFollowingController` but not needed for the `OptimizedDispatchController` (Issue #443)
   - PR #587
- [x] 4. Refactor the open-loop controllers (and their corresponding storage performance models) to minimize the use of pass-through components and sync-up the open-loop controllers (and their corresponding storage performance models) to better align with the Pyomo controller outputs. The issues that this encompasses are:
   - [x] #566
   - [x] #498
   - [x] PR #589 (dropped)
   - [x] PR #608
   - [x] PR #612 
- [x] 5. Integrate the use of the open-loop controllers into the PySAM battery and generic storage model created in Step 2. (this is a baby step towards Step 7)
   - [x] #567 
   - [x] PR #613
   - [/] PR #615 (follow-on bugfix to 613)
- [ ] 6. Integrate the use of the Pyomo controllers into the `StorageAutoSizingModel` (this is a baby step towards Step 7)
   - [x] #568 
- [ ] 7. Introduce a storage performance model baseclass and update the pysam battery and generic storage model to inherit this. This depends on steps 1, 2, and 4 being complete.
   - [ ] #521
- [ ] 8. Update the `StorageAutoSizingModel` and `SimpleGenericStorage` (or equivalent/whichever is relevant at this point) to inherit the storage model baseclass created in Step 7
- [ ] 9. At this point, all controllers should be compatible with all storage performance models. In one of the above steps, its expected that there may be some updates to existing baseclasses and/or base configuration classes to ensure that storage design parameters have the same naming convention across controllers and storage performance models.

- Clean-ups
   - [x] #569
[...]

Issue #223, opened 2025-08-20
Standardize naming conventions


elenya-grant:
# Standardized naming conventions

Have a standardized and documented naming convention for model inputs and outputs given that a lot of logic relies on strings. This would include naming outputs in a specific way that indicate whether the output is a timeseries profile, an aggregated value, or a lifetime estimated value. Also would include naming conventions for not just the primary flow variables, but also the secondary inputs/outputs that may not be 'primary'. For example, the electrolyzer model has a primary timeseries input of an electricity profile and primary timeseries output of hydrogen. The electrolyzer consumes electricity (although may not consume all the electricity input) and consumes water. The consumption of water is not output from the model and not included in any financial calculations. The electrolyzer model could output results for the plant lifetime but does not at the moment.

Also - have a more standardized/automated method of understanding what technologies are connected to what commodities. 

### Standardized naming convention
Standardized model inputs/outputs could look like introducing a base class for different technology types (converters, storage, transporters, etc). At the very least, this should include documentation.  Below is a draft of some possible naming conventions:

**Converters (commodity to commodity)**
- inputs: 
    - timeseries: 
        - `{primary_feedstock}_in`
        - `{secondary_feedstock}_in` (optional)
    - annual:
    - lifetime:
- outputs: 
    - timeseries: 
        - `{commodity}_out`: timeseries of commodity production
        - `{primary_feedstock}_consumed`: timeseries of primary feedstock consumed
        - `{secondary_feedstock}_consumed`: if used, timeseries of secondary feedstock consumed (only required for hourly prices for secondary feedstock)
    - annual:
        - `max_{commodity}_production_possible`: maximum amount of commodity that could be produced over the simulation if the converter was running at it's rated capacity for the entire simulation. Should be representative of the converter's rated capacity.
        - `total_{commodity}_produced`: sum of `{commodity}_out`
        - `capacity_factor`: equal to `total_{commodity}_produced/max_{commodity}_production_possible`. Should be a value between 0 and 1.
        - `total_{primary_feedstock}_consumed`: sum of `{primary_feedstock}_consumed`
        - `total_{secondary_feedstock}_consumed`: sum of `{secondary_feedstock}_consumed`
    - lifetime (over plant life): 
        - `total_life_{secondary_feedstock}_consumed` (if relevant)
    
**storage**
- inputs: 
    - timeseries: 
        - `{commodity}_in`
    - annual:
    - lifetime:
- outputs: 
    - timeseries: 
[...]

PR #571, opened 2026-03-04
Generic Storage Model that's Compatible with Pyomo Controllers

elenya-grant:
# Generic Storage Model that's Compatible with Pyomo Controllers

Added a generic storage model that can be used with the Pyomo Controllers.

Before this PR, only the PySAM battery was compatible with the pyomo controllers (the heuristic load following dispatch controller and the optimized dispatch controller). This PR enables the ability to use the Pyomo controllers for any commodity (most relevant may be hydrogen storage). 

This PR addresses a step outlined in Issue #564, and would resolve issue #441.

- [ ] Feature Enhancement
  - [ ] Framework
  - [x] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Clean-up docstrings
- [x] Clean-up in-line comments
- [x] Add additional regression tests
- [x] revert changes to pysam battery model
- [x] update `simulate` method of pysam battery model 
- [x] add unit test with <1 charge/discharge efficiency

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:
- Naming outputs from the storage model, and thinking about how the output naming can be aligned with the PySAM battery performance model. For example, the PySAM battery performance model outputs `battery_electricity_charge` and `battery_electricity_discharge` whereas this model outputs `storage_commodity_charge` and `storage_commodity_discharge`. To enforce a more similar naming convention, then maybe the PySAM battery should be updated to have outputs of `storage_electricity_charge` and `storage_electricity_discharge`. Would love ideas on how to name the outputs of this generic storage model that can be applied for all storage models. This is somewhat related to Issue #521 but is more related to trying to standardize the output naming convention of storage performance models.

- [ ] Documentation

## Section 3: Related Issues

- This PR would resolve _a step_ outlined in Issue #564
- This PR would fully resolve issue #441.

### Section 4.1: New Files
- `h2integrate/storage/generic_storage_pyo.py`: performance model and config for a generic (commodity-neutral) storage model that is compatible with pyomo controllers
  - `StoragePerformanceModel`: storage performance model that is commodity-neutral and can be used with the pyomo controller
- `h2integrate/storage/test/test_generic_storage_pyo.py`: tests the generic storage model without a pyomo controller
- `h2integrate/control/test/test_heuristic_with_generic_storage.py`: tests the compatibility of the generic storage model with the `HeuristicLoadFollowingController` controller
- `h2integrate/control/test/test_optimal_controller_with_generic_storage.py`:tests the compatibility of the generic storage model with the `OptimizedDispatchController` controller

- `h2integrate/core/supported_models.py`: added `StoragePerformanceModel`
- `h2integrate/storage/battery/pysam_battery.py`: updated the `simulate()` method to more closely resemble the `simulate()` method in `StoragePerformanceModel`. These developments were supported by @jaredthomas68 

- [x] **Model Structure**:
    - [-] If applicable: inherit from `CostModelBaseClass`
  - [-] If a new commodity_type is added, update `create_financial_model` in `h2integrate_model.py`
  - [-] If applicable add integration tests
- [-] **Example**: If applicable, a working example demonstrating the new model has been created
  - [-] Input file comments
  - [-] Run file comments
  - [-] Example has been tested and runs successfully in `test_all_examples.py`
- [x] **Documentation**:
    - [-] Model documentation page added to the appropriate `docs/` section
    - [-] `<model_name>.md` is added to the `_toc.yml`


h2integrate/storage/battery/pysam_battery.py
```diff
@@ -493,6 +493,12 @@ def simulate(
         storage_power_out_timesteps = np.zeros(self.config.n_control_window)
         soc_timesteps = np.zeros(self.config.n_control_window)
 
+        # discharge is positive
+        discharge_timesteps = np.zeros(self.config.n_control_window)
+
+        # charge is negative
+        charge_timesteps = np.zeros(self.config.n_control_window)
+
```
jaredthomas68: What is the need for these? I don't see them being used.

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+        Ensures that either `charge_efficiency` and `discharge_efficiency` are provided,
+        or `round_trip_efficiency` is provided. If `round_trip_efficiency` is provided,
+        it calculates `charge_efficiency` and `discharge_efficiency` as the square root
+        of `round_trip_efficiency`.
+        """
+        if self.round_trip_efficiency is not None:
+            if self.charge_efficiency is not None or self.discharge_efficiency is not None:
+                raise ValueError(
+                    "Provide either `round_trip_efficiency` or both `charge_efficiency` "
+                    "and `discharge_efficiency`, but not both."
```
jaredthomas68: The wording here tripped me up for a moment.

``` suggestion
                    "Exactly one of the following sets of parameters must be set: (a) `round_trip_efficiency`, or (b) both"  
                    "`charge_efficiency` and `discharge_efficiency`."
```

jaredthomas68: I noticed that the same error message appears later. I think you could reframe the logic so that the error is only needed in one place if you check the charge and discharge efficiency first, or use separate logic for computations and raising errors. Non-blocking comment.

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+    """
+
+    def setup(self):
+        """Set up the PySAM storage Performance model in OpenMDAO.
+
+        Initializes the configuration, defines inputs/outputs for OpenMDAO,
+        and creates a `BatteryStateful` instance with the selected chemistry.
+        If dispatch connections are specified, it also sets up a discrete
+        input for Pyomo solver integration.
+        """
```
jaredthomas68:
```suggestion
        Initializes the configuration and defines inputs/outputs for OpenMDAO.
        If dispatch connections are specified, it also sets up a discrete
        input for Pyomo solver integration.
```

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+            commodity output from the storage model in commodity_rate_units.
+
+    Methods:
+        setup():
+            Defines model inputs, outputs, configuration, and connections
+            to plant-level dispatch (if applicable).
+        compute(inputs, outputs, discrete_inputs, discrete_outputs):
+            Runs the PySAM BatteryStateful model for a simulation timestep,
+            updating outputs such as SOC, charge/discharge limits, unmet
+            demand, and unused commodities.
```
jaredthomas68: needs to be updated

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+    Notes:
+        - Default timestep is 1 hour (``dt=1.0``).
+        - State of charge (SOC) bounds are set using the configuration's
+          ``min_charge_fraction`` and ``max_charge_fraction``.
+        - If a Pyomo dispatch solver is provided, the storage will simulate
+          dispatch decisions using solver inputs.
+    """
+
+    def setup(self):
+        """Set up the PySAM storage Performance model in OpenMDAO.
```
jaredthomas68: remove "PySAM"

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+
+        self.add_output(
+            f"unused_{self.commodity}_out",
+            val=0.0,
+            shape=self.n_timesteps,
+            units=self.commodity_rate_units,
+            desc="Unused generated commodity",
+        )
+
+        # Initialize the PySAM BatteryStateful model with defaults
```
jaredthomas68: remove "PySAM"

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+            self.tech_group_name = self.pathname.split(".")
+            for _source_tech, intended_dispatch_tech in self.options["plant_config"][
+                "tech_to_dispatch_connections"
+            ]:
+                if any(intended_dispatch_tech in name for name in self.tech_group_name):
+                    self.add_discrete_input("pyomo_dispatch_solver", val=dummy_function)
+                    break
+
+    def compute(self, inputs, outputs, discrete_inputs=[], discrete_outputs=[]):
+        """Run the PySAM storage model for one simulation step.
```
jaredthomas68: This comment is incorrect in this and the pysam battery.

```suggestion
        """
```

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+
+            # determine how much of the inflow commodity was unused
+            unused_commodity = np.maximum(
+                0, combined_commodity_out - inputs[f"{self.commodity}_demand"]
+            )
+
+            # determine how much demand was not met
+            unmet_demand = np.maximum(
+                0, inputs[f"{self.commodity}_demand"] - combined_commodity_out
+            )
```
jaredthomas68, line 394: This block is nearly identical to what is in pysam battery. I would love to see this code reused instead of copied. I recognize there are some differences, but the only big thing I'm seeing that would make it hard to reuse is the inputs to simulate. Perhaps we can make this a method in a future pyomo storage base class as part #564, related to #521.

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+                # max that storage can be charged
+                max_chargeable = (
+                    np.min(
+                        [
+                            available_charge,
+                            charge_rate / self.config.charge_efficiency,
+                            -1 * dispatch_command_t,
+                        ]
+                    )
+                    * self.config.charge_efficiency
```
jaredthomas68: This is also mostly duplicate but slightly reconfigured code. I think this is a good opportunity for code reuse.

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+            self.commodity_amount_units = f"({self.commodity_rate_units})*h"
+
+
+class StoragePerformanceModel(PerformanceModelBaseClass):
+    """OpenMDAO component for a storage component.
+
+    Attributes:
+        config (StoragePerformanceModelConfig):
+            Configuration parameters for the storage performance model.
+        current_soc (float): soc at the start of each interval that the simulate()
```
jaredthomas68, line 136: `current_soc` does not make sense to me as an input. The user can specify the initial soc, but the soc at the start of each interval should be determined by the previous interval

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -0,0 +1,524 @@
...
+                    * self.config.charge_efficiency
+                )
+                #
+                if dispatch_command_t < -max_chargeable:
+                    dispatch_command_t = -max_chargeable
+
+            else:
+                # available discharge from storage
+                available_discharge = float(
+                    (soc - self.config.min_charge_fraction) * storage_capacity / self.dt_hr
```
jaredthomas68: is there a problem with using min_soc and max_soc? I think those are generic enough to work with any tech and they are much shorter.

h2integrate/storage/generic_storage_pyo.py
jaredthomas68: I made a PR into @elenya-grant fork reverting back to follow the logic in the pysam performance model to hopefully allow us to make this part of the base class and not have to maintain two different sets of logic. I think all we need to do is make some uniform methods in the class that be called instead of the pysam calls, and the pysam calls would happen inside those methods. See my PR here: https://github.com/elenya-grant/GreenHEART/pull/4

jaredthomas68 (changes requested):
This is a great push forward @elenya-grant. I would like to revert to the logic in the pysam battery performance model `simulate()` method so we can keep the same code for both. With a little work I think we can make the simulate method in the base class which I think will open up opportunities to add other storage simulation models to work with pyomo more easily and should also reduce our maintenance overhead.

h2integrate/storage/battery/pysam_battery.py
elenya-grant: yes - sorry about that - I was using that to figure out what was happening in the pysam battery and I plan on reverting these changes.

h2integrate/storage/generic_storage_pyo.py
elenya-grant: Yes, I agree,  that is the plan and is an intended step laid out in Issue #564. This is a step towards creating a storage performance baseclass that is compatible with Pyomo controllers and open-loop controllers and will include methods for repeated calculations. This PR is just intended to introduce this model (and test it), which will help inform what the storage performance model baseclass should look like. 

I reorganized Issue #564 to reflect all the steps and sub-steps in a clearer way. Now, step 7 in Issue #564 explicitly states the plan to do this (with steps 4-6 being required before step 7).

h2integrate/storage/generic_storage_pyo.py
elenya-grant: yes - that is a planned future PR but not intended to be part of this PR. We want to make the storage performance model baseclass _after_ we sync up/fix the open-loop controllers and the storage performance models that use them (storage auto-sizing and pass-through storage). This is so that we can make a storage performance baseclass that is compatible with all controllers (pyomo and open loop). So - theres a few steps laid out before we introduce the storage performance baseclass (now steps 4-6 in Issue #564). That being said, I'll check out your PR! 

I'm currently working on refactoring the open loop controllers and the other storage performance models, which I imagine will result in even more duplicative code. I think there may also be duplicative calculations in the controllers which I'd like to reduce.

h2integrate/storage/generic_storage_pyo.py
elenya-grant: `current_soc` is set to the initial soc at the start of the compute method. It **is** updated at the previous interval (its updated on the line above the `return` statement of the `simulate()` method)

h2integrate/storage/generic_storage_pyo.py
elenya-grant: I used the same naming that is used in the Pyomo controller configs and the open-loop controller configs. So - that's why they're named this way.

h2integrate/storage/generic_storage_pyo.py
elenya-grant: yeah - I'll try to find a slicker way to handle that error message. I did update the error message based on your suggestion.

h2integrate/storage/battery/pysam_battery.py
elenya-grant: changes were reverted!

h2integrate/storage/generic_storage_pyo.py
elenya-grant: done!

h2integrate/storage/generic_storage_pyo.py
elenya-grant: done!

h2integrate/storage/generic_storage_pyo.py
elenya-grant: done!

h2integrate/storage/generic_storage_pyo.py
elenya-grant: fixed!

h2integrate/storage/generic_storage_pyo.py
elenya-grant: After looking at the PR you made, I see that you're not asking for a baseclass now, but instead a way to make the PySAM battery and generic storage model have a similar simulate() method so that it could be more easily brought into a baseclass in the future (which I am down for, I just didn't want to introduce the baseclass in this PR). I will leave a review on your PR and we can discuss our own rational for how we like the `simulate()` method to look in the generic storage model. Thanks for making that PR - it helped me understand what your comment here meant.

h2integrate/storage/generic_storage_pyo.py
johnjasa, line 1: This file is both similar to and different from the `pysam_battery.py` version. I imagine you considered having the classes shared methods in some way, maybe through a base class. I want to learn from you; is there not enough similarity to benefit from that, do you expect future changes to introduce more shared code, or will the generic model stay separate from the pysam battery (and potentially other storage models) and that's okay? To be clear, I'm not advocating for one way or another, really just asking you to share your thoughts cause you've thought about this more

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
   `min_charge_fraction`, `max_charge_fraction`, and `init_charge_fraction` across all
   configuration classes, YAML configs, tests, and examples. These values are fractions
   between 0 and 1, so the previous "percent" naming was misleading. [PR 581](https://github.com/NatLabRockies/H2Integrate/pull/581)
+- Added a generic storage model that is compatible with the Pyomo controllers [PR 571](https://github.com/NatLabRockies/H2Integrate/pull/571)
```
johnjasa, line 14: Thanks for doing this, and with the PR link!

h2integrate/storage/generic_storage_pyo.py
johnjasa: To add more here, `current_soc` is an `attr` on the performance class, not the config class, so it's not used as a user-facing input but more of a persistent tracking object for the soc.

johnjasa (approved):
Thanks for this PR, Elenya! I pushed up some changes directly, mostly to the `simulate()` logic and comments, but also the doc section. Once I dug through what this PR is doing, I wanted to share some of that understanding as comments. Take a look and feel free to modify anything; I'm not saying it has to be that way, but it seems a bit clearer to me.

For your specific question in the PR body, I don't love `battery_` or `storage_` in the names there. Could we just do `commodity_charge` or is that ambiguous/already used? My other thought is to return to what we discussed around #223, but that might no longer be relevant? From there, we see these names in the storage section:
```
{commodity}_out
{commodity}_excess
{commodity}_missed_load
{technology}_SOC
{technology}_discharged
{technology}_charged
```

Despite my approval, I don't want to merge this in into we get another approval from a Jared or Gen type, both for the code and this naming discussion.

Overall, fantastic progress and a big value add! I see this as immediately useful for a few different projects.

h2integrate/storage/generic_storage_pyo.py
elenya-grant: Yes - I plan for future changes so that these two models share a baseclass that includes the shared code (see Issue #564). I first want to make these storage performance models compatible with the open-loop controllers before trying to make a storage performance baseclass. There are also two other storage performance models (the autosizing and the pass-through one) that would likely inherit the storage performance baseclass. This is why I think the smaller steps in Issue #564 should be done before introducing a baseclass. Other reasons I'm waiting to introduce the baseclass is that I imagine that folks will have opinions about 1) input/output naming and 2) what those outputs should be - which is another reason why I'm trying to split this up across multiple steps.  

Did that answer your question?

johnjasa: Yes, this is very helpful! Thank you for typing it out, I support this approach.

elenya-grant:
@johnjasa, I'm responding to your comment [here](https://github.com/NatLabRockies/H2Integrate/pull/571#pullrequestreview-3916174388)

```
{commodity}_out
{commodity}_excess
{commodity}_missed_load
{technology}_SOC
{technology}_discharged
{technology}_charged
```

I don't love the naming that has to do with the technology, but I think I understand the reasoning. Currently, I also like having the combined charge and discharge profile (which was named `storage_{commodity}_out`) because I'd like to eventually push to have this be `commodity_out` (instead of `commodity_out` being the combined input commodity, adjusted for the storage charge/discharge, and saturated at the demand). I also think that if we update the naming here, we should update the naming in the PySAM battery so that it's consistent between the two models.

Update: Thanks for the feedback on the output naming, this feedback will be taken into consideration in a future PR, such as when the storage performance model baseclass is implemented.

h2integrate/storage/generic_storage_pyo.py
jaredthomas68: @elenya-grant , as per our discussion this morning, and the changes introduced to your version by @johnjasa, my main concern now is that we work towards consistency to avoid introducing yet another set of dispatch logic to maintain. I think the plan is to update the pysam battery with yours/ @johnjasa's version here to allow for charge and discharge efficiency and the logic flow you prefer.

jaredthomas68: I would be interested to get @genevievestarke's feedback on the logic here and my PR as well.

h2integrate/storage/generic_storage_pyo.py
jaredthomas68: Thanks, I guess I was going to fast and was thinking it was in from the config class.

h2integrate/storage/generic_storage_pyo.py
jaredthomas68: And, I think I put the original names in there. Sorry. I think the reasoning was that charge fraction was less likely to be misunderstood to be a percent.

h2integrate/storage/test/test_generic_storage_pyo.py
```diff
@@ -0,0 +1,652 @@
...
+        "commodity_rate_units": "kg/h",
+        "max_capacity": 40,
+        "max_charge_rate": 10,
+        "min_charge_fraction": 0.1,
+        "max_charge_fraction": 1.0,
+        "init_charge_fraction": 0.1,
+        "n_control_window": 24,
+        "commodity_amount_units": "kg",
+        "charge_equals_discharge": True,
+        "charge_efficiency": 1.0,
```
jaredthomas68, line 23: We need a test for `round_trip_efficiency` less than one.

h2integrate/storage/generic_storage_pyo.py
jaredthomas68: Thanks for explanation and plan. Sounds good.

jaredthomas68 (changes requested):
Super close, one more test request.

h2integrate/storage/test/test_generic_storage_pyo.py
elenya-grant: just added it in!

jaredthomas68 (approved):
Looks good to me! Thanks!

Issue #497, opened 2026-02-03
Reorganize Pyomo control rule classes


elenya-grant:
# Reorganize Pyomo control strategy classes

The file `h2integrate/control/control_strategies/pyomo_controllers.py` has 3 controller strategy classes (and soon more as PR #407 is merged in) with varying level of inheritance that makes it difficult to follow the logic that distinguishes the individual control strategies. I think that the baseclasses should only include methods that are fully used across all controllers that inherit it, without any excess logic that is specific to a single controller.

I also think that the controller baseclass in `h2integrate/control/control_strategies/controller_baseclass.py` could be removed and be replaced with the `PyomoControllerBaseClass`.

I almost think it'd be nice to split out these controllers into separate files. Like `PyomoControllerBaseConfig` and `PyomoControllerBaseClass` (maybe even `SolverOptions` if that's used across the Pyomo controllers) could be moved into `h2integrate/control/control_strategies/controller_baseclass.py`. Then separate files for `SimpleBatteryControllerHeuristic`, `HeuristicLoadFollowingControllerConfig` and `HeuristicLoadFollowingController`, and `OptimizedDispatchControllerConfig` and `OptimizedDispatchController`.

genevievestarke:
I believe this is addressed now? @elenya-grant

Issue #489, opened 2026-01-29
Make sure dispatch windows are being applied appropriately


genevievestarke:
# Make sure dispatch windows are being applied appropriately

The pyomo dispatch framework has the input variables `n_control_window` and `n_horizon_window`. Currently, only `n_control_window` is being used. The correct use of `n_horizon_window` needs to be applied for the optimal control strategy.

## Proposed solution

Implement a receding horizon for the optimal control of length `n_horizon_window` and then apply the results to the outputs for length `n_control_window`

Issue #443, opened 2026-01-09
Unify the optimized and heuristic dispatch code in H2I


genevievestarke:
The heuristic and optimized dispatch code in H2I diverged during development, so it would be good to clean up both codes and unify them into a cohesive framework, focusing on reducing unused code and increasing readability of the code.

PR #587, opened 2026-03-06
Break out Pyomo controller simulation code

### LINKED ISSUE #513: Separate out the Pyomo heuristic and optimization methods during the set up
# Separate out the Pyomo heuristic and optimization methods during the setup

- We need to define dispatch rules, even though the optimization method only uses a Pyomo class structure
- The dispatch rules are needed due to the structure of `PyomoControllerBaseClass`. The code will not run if they are not defined
- We can't use the `min_operating_cost` keys because those are classes and do not work with the OpenMDAO structure in `PyomoControllerBaseClass`
- We are currently using the heuristic versions just for the `tech_config.yaml`, and using the optimization classes internally to the dispatch.

The following definition gives the heuristic version for the battery technology definition, but because the control strategy is set to `OptimizedDispatchController`, it will initialize using `PyomoRuleStorageBaseclass` in the setup, then use `PyomoRuleStorageMinOperatingCosts` in the `OptimizedDispatchController` class.

```python 
  battery:
    dispatch_rule_set:
      model: "PyomoRuleStorageBaseclass"
    control_strategy:
      model: "OptimizedDispatchController"
    performance_model:
      model: "PySAMBatteryPerformanceModel"
    cost_model:
      model: "ATBBatteryCostModel"
```

## System Information

- OS: <macOS 12.4>
  
- Python version: <3.11.4>
  
- H2Integrate version: <0.1.1>
  
  - <Installed from source using an editable installation with developer tools: `pip install -e .[develop]`>
  - Commit hash: <commit-hash>
    

### Relevant library versions

  | Package | Version |
  | ------- | ------- |
  | hopp | <3.1.1> |
  | numpy | <1.26.4> |
  | Pyomo | <6.8.0> |
  | scipy | <1.14.1> |
  | <another-relevant-package> | <version> |

genevievestarke:
# Break out code for Pyomo Controllers for readability

This PR moves code specific to individual controllers out of `pyomo_controller_baseclass.py` and into the individual controller files. This leads to some repeated code, but makes the flow of the controller code more clear. This also fixes #513, which describes how `dispatch_rule_set` needed to be set structurally to run the Optimized Dispatch Controller, even though they weren't actually used.
This PR also contributes to #443 & #497, and clears the way to address #489.

Specifically, this PR moved code in `pyomo_controller_baseclass.py` from 

- `setup()`: Moved the hard-coded dispatch block rule assignment
- `pyomo_setup()`: Moved mix of hard-coded dispatch block rule assignment and optimized dispatch definitions to each individual file, and let each controller define their own `pyomo_dispatch_solver' function to remove if statements in base class based on controller type.

- [ ] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Documentation

## Section 3: Related Issues

This PR fixes #513, which describes how `dispatch_rule_set` needed to be set structurally to run the Optimized Dispatch Controller, even though they weren't actually used.
This PR also contributes to #443 & #497, and clears the way to address #489.

### Section 4.1: New Files
- `path/to/file.extension`

  - `connect_technologies`: Added an if statement to only connect dispatch rules if they are defined in the tech_config
- `h2integrate/control/control_strategies/pyomo_controller_baseclass.py`
  - `setup`: Moved the hard-coded dispatch block rule assignment
  - `pyomo_setup`: Moved mix of hard-coded dispatch block rule assignment and optimized dispatch definitions to each individual file, and let each controller define their own `pyomo_dispatch_solver' function to remove if statements in base class based on controller type.
- `h2integrate/control/control_strategies/heuristic_pyomo_controller.py`
  - `setup`: Moved the dispatch block rule assignment here
  - `pyomo_setup`: Moved hard-coded dispatch block rule assignment here and defined the heuristic `pyomo_dispatch_solver' 
- `h2integrate/control/control_strategies/optimized_pyomo_controller.py`
  - `pyomo_setup`: Moved optimized dispatch technology definitions here and defined the heuristic `pyomo_dispatch_solver' 
- `h2integrate/control/test/test_optimal_controllers.py`
  - `tech_config`: Removed `dispatch_rule_set` from technology definitions as they are no longer necessary


h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -41,6 +43,21 @@ def setup(self):
...
+                    # When getting rules for the same tech, the tech name is not used in order to
+                    # allow for automatic connections rather than complicating the h2i model set up
+                    self.add_discrete_input("dispatch_block_rule_function", val=self.dummy_method)
+                else:
+                    self.add_discrete_input(
+                        f"{'dispatch_block_rule_function'}_{source_tech}", val=self.dummy_method
+                    )
+            else:
+                continue
+
```
jaredthomas68, line 99: Why is this code needed if the rules are no longer needed for the heuristic?

docs/control/pyomo_controllers.md
```diff
@@ -21,7 +21,7 @@ For an example of how to use the heuristic Pyomo control framework with the `Heu
 
 (optimized-load-following-controller)=
 ## Optimized Load Following Controller
-The optimized dispatch method is specified by setting the storage control to  `OptimizedDispatchController`. The same `dispatch_rule_set` for each technology connected to the storage technology is followed as in the heuristic case, where each technology must also have a `dispatch_rule_set` defined in the `tech_config`. The optimized dispatch currently uses the same dispatch rules as the heuristic dispatch. This method maximizes the load met while minimizing the cost of the system (operating cost) over each specified time window.
+The optimized dispatch method is specified by setting the storage control to  `OptimizedDispatchController`. Unlike the heuristic method, the optimized dispatch method does not use `dispatch_rule_set` as in input in the `tech_config`. This method maximizes the load met while minimizing the cost of the system (operating cost) over each specified time window.
```
jaredthomas68:
```suggestion
The optimized dispatch method is specified by setting the storage control to  `OptimizedDispatchController`. Unlike the heuristic method, the optimized dispatch method does not use `dispatch_rule_set` as an input in the `tech_config`. The `OptimizedDispatchController` method maximizes the load met while minimizing the cost of the system (operating cost) over each specified time window.
```

h2integrate/control/control_strategies/pyomo_controller_baseclass.py
```diff
@@ -109,9 +109,7 @@ def dummy_method(self, in1, in2):
     def setup(self):
         """Register per-technology dispatch rule inputs and expose the solver callable.
 
-        Adds discrete inputs named 'dispatch_block_rule_function' (and variants
-        suffixed with source tech names for cross-tech connections) plus a
-        discrete output 'pyomo_dispatch_solver' that will hold the assembled
+        Adds discrete a discrete output 'pyomo_dispatch_solver' that will hold the assembled
```
jaredthomas68:
```suggestion
 Adds discrete output 'pyomo_dispatch_solver' that will hold the assembled
```

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
jaredthomas68, line 20: doc strings

jaredthomas68 (commented):
I see this as a preliminary review because I got hung up where the optimized dispatch does not need the block rule sets. @genevievestarke, can you explain why the dispatch rules are not relevant to the `OptimizedDispatchController`? 

I am also generally concerned with the step away from code re-use here, though I understand there is benefit to having the pyomo controllers be in separate classes. I think there are some opportunities for more code-reuse in this PR, but perhaps you could explain the need to having duplicate code like this between the two pyomo control classes.

genevievestarke:
> I see this as a preliminary review because I got hung up where the optimized dispatch does not need the block rule sets. @genevievestarke, can you explain why the dispatch rules are not relevant to the `OptimizedDispatchController`?
> 
> I am also generally concerned with the step away from code re-use here, though I understand there is benefit to having the pyomo controllers be in separate classes. I think there are some opportunities for more code-reuse in this PR, but perhaps you could explain the need to having duplicate code like this between the two pyomo control classes.

Hey, @jaredthomas68! Yeah, I can see how that's confusing, which is why I didn't want to lose this point in the issues! The optimized pyomo controller doesn't use the block set rules because it instead uses a class structure. This means that you define `control_strategy` = `OptimizedDispatchController` in the storage tech_config and Pyomo does the rest using the dispatch connections defined in plant_config. I moved away from the block rule structure because the optimization pyomo models need access to more than just the block_rule_function, like the `initialize_parameters()` functions in each pyomo technology model. This worked in the heuristic case because `initialize_parameters()` was defined in the controller itself, while the optimized version collects multiple pyomo technology models into one plant model to give to the optimizer. Those, the extra functions, like `initialize_parameters()`, need to be in the pyomo files themselves, like `pyomo_storage_rule_min_operating_cost.py`. I couldn't find a way to access these functions if just the dispatch block rule was passed to the pyomo controller, so I switched it to a class structure that takes place mostly outside of the openmdao structure (one of my main concerns with this dispatch framework).

For the code re-use point, I'm happy to chat about how we could reuse more of it! My main aim with this PR was to separate out the code from the heuristic and optimization controllers so we can see what is happening, and then perhaps see more clearly what makes sense to re-use. I don't know how to use some common code in a function like `pyomo_dispatch_solver` without overwriting the whole thing. I would prefer to remove references to specific control strategies in the base class, so I'm open to ideas on how to achieve that without duplicating code. Maybe we need to break it out into multiple functions that can be overwritten in smaller blocks?

h2integrate/control/control_strategies/pyomo_controller_baseclass.py
```diff
@@ -151,43 +135,16 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
         discrete_outputs["pyomo_dispatch_solver"] = self.pyomo_setup(discrete_inputs)
 
     def pyomo_setup(self, discrete_inputs):
```
elenya-grant, line 137: should `pyomo_setup()` just raise a not implemented error in the baseclass? Then include docstrings/comments about what it needs to do (like having the `pyomo_dispatch_solver` function and what that function needs to do?).

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -57,6 +120,154 @@ def setup(self):
...
+                    self.storage_dispatch_commands,
+                    **performance_model_kwargs,
+                    sim_start_index=t,
+                )
+
+                # get a list of all time indices belonging to the current control window
+                window_indices = list(range(t, t + self.config.n_control_window))
+
+                # loop over all time steps in the current control window
+                for j in window_indices:
```
elenya-grant, line 247: I think that we could remove the part of this function where we calculate the `unmet_demand`, `unused commodity`, and `total_commodity_out` (perhaps not in this PR, since that'd require changes to the pysam battery and generic storage model) but - I do think that may be nice to remove in the future. I think that we could just have the soc and storage_commodity_out be returned in the `pyomo_dispatch_solver` method. Again, this should be a separate PR.

elenya-grant (commented):
I just left a few questions/comments but would love to discuss this more over a call (I mostly have questions). I think this is looking really great! It would be nice, but maybe really difficult (and not the point of this PR), if there was a way so that the optimized controller and heuristic controller had a similar workflow and could use the same `pyomo_setup()` method (that was in the baseclass) but without having hard-coded logic for the specific controllers that inherit it. I'd imagine that this would require a lot more changes to the heuristic controller than the optimized one. But - that is not the route we have to go, I'm just thinking that since the `pyomo_setup()` is such a complex method, it'd be nice if it could be in a baseclass and be shared. I'd love to chat and see if that's possible/what would need to be done to do that. But - I think the work you've done here will help inform whether that's possible or makes sense.

Thanks! It's looking really good and helping me better understand how the Heuristic and Optimal controllers are different in their workflow.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
elenya-grant: done

elenya-grant:
Responding to [this thread](https://github.com/NatLabRockies/H2Integrate/pull/587#issuecomment-4026939479):

@jaredthomas68 - I think the only way to make the `pyomo_setup()` method in the base class would require a larger refactor of the heuristic controller so that it's workflow is more similar to that of the optimized controller (since the optimized controller and the heuristic controller have two different workflows). This would also require refactoring `PyomoDispatchGenericConverter` and `PyomoRuleStorageBaseclass`. 

I think that we can either:
- try to align the heuristic and optimized controller to have more similar workflows, allowing shared logic to live in the baseclass (such as the `pyomo_setup()` method) OR
- keep the controllers with their own separate workflows, having somewhat duplicated logic in each controller and requiring all future controllers to overwrite the `pyomo_setup()` method in the baseclass (this is what this PR does)

h2integrate/control/control_strategies/pyomo_controller_baseclass.py
elenya-grant: update - I realize that we need it as-is so that the `setup()` method can live in the baseclass. Ignore the above comment!

jaredthomas68:
> Responding to [this thread](https://github.com/NatLabRockies/H2Integrate/pull/587#issuecomment-4026939479):
> 
> @jaredthomas68 - I think the only way to make the `pyomo_setup()` method in the base class would require a larger refactor of the heuristic controller so that it's workflow is more similar to that of the optimized controller (since the optimized controller and the heuristic controller have two different workflows). This would also require refactoring `PyomoDispatchGenericConverter` and `PyomoRuleStorageBaseclass`.
> 
> I think that we can either:
> 
>     * try to align the heuristic and optimized controller to have more similar workflows, allowing shared logic to live in the baseclass (such as the `pyomo_setup()` method) OR
> 
>     * keep the controllers with their own separate workflows, having somewhat duplicated logic in each controller and requiring all future controllers to overwrite the `pyomo_setup()` method in the baseclass (this is what this PR does)

I agree with @elenya-grant, that getting to more code reuse in the pyomo controllers would likely mean adjusting the heuristic control logic to match the optimized control logic. While the optimized version does not make the control rules passing transparent in the N2 diagram, I like that the optimized version still uses the connection inputs to build the pyomo model. I think the optimized version may actually be better because of its greater simplicity that in my mind may out weigh the transparency we tried to get with the heuristic implementation.

In my mind, the best way forward is to complete this PR as is and then adjust the heuristic to more closely match the optimized logic, and possibly even a third step for separately moving the logic to a base class.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
genevievestarke: The blocks are used for the heuristic version, but not the optimized version.

johnjasa (commented):
I like this PR, thank you for doing it! I'm generally a fan of making the codebase and setup more clear for users, even if it means having some repeated code. I think this PR does a good job of walking that line. I understand it's one of many steps in refactoring the dispatch setups.

I purposefully am just commenting here and not approving or requesting changes because I'll defer to others about the actual changed content in this PR. I like the conversations and support already happening here. Thanks!

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -16,8 +18,47 @@
...
 class HeuristicLoadFollowingControllerConfig(PyomoControllerBaseConfig):
+    """Configuration class for the HeuristicLoadFollowingController.
+
+    Attributes:
+        charge_efficiency (float | None, optional): Efficiency of charging the storage, represented
+            as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Optional if
+            `round_trip_efficiency` is provided.
+        discharge_efficiency (float | None, optional): Efficiency of discharging the storage,
+            represented as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Optional if
+            `round_trip_efficiency` is provided.
```
jaredthomas68:
```suggestion
        charge_efficiency (float | None, optional): Efficiency of charging the storage, represented
            as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Must be `None` if
            `round_trip_efficiency` is provided.
        discharge_efficiency (float | None, optional): Efficiency of discharging the storage,
            represented as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Must be `None` if
            `round_trip_efficiency` is provided.
```

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -16,8 +18,47 @@
...
+        round_trip_efficiency (float | None, optional): Combined efficiency of charging and
+            discharging the storage, represented as a decimal between 0 and 1 (e.g., 0.81 for
+            81% efficiency). Optional if `charge_efficiency` and `discharge_efficiency` are
+            provided.
```
jaredthomas68, line 33:
```suggestion
        round_trip_efficiency (float | None, optional): Combined efficiency of charging and
            discharging the storage, represented as a decimal between 0 and 1 (e.g., 0.81 for
            81% efficiency). Must be `None` if `charge_efficiency` or `discharge_efficiency` are
            provided.
```

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
```diff
@@ -16,8 +18,47 @@
...
+        Post-initialization logic to validate and calculate efficiencies.
+
+        Ensures that either `charge_efficiency` and `discharge_efficiency` are provided,
+        or `round_trip_efficiency` is provided. If `round_trip_efficiency` is provided,
+        it calculates `charge_efficiency` and `discharge_efficiency` as the square root
+        of `round_trip_efficiency`.
+        """
+
+        super().__attrs_post_init__()
+        if self.round_trip_efficiency is not None:
```
jaredthomas68, line 51: The logic here could be adjusted if we want to allow for all possible two-out-of-three combinations from `charge_efficiency`, `discharge_efficiency`, and `round_trip_efficiency`, but I think that can wait for later work.

elenya-grant (approved):
This looks good to me! Thanks for working on separating these methods out! I think this will help future development of the pyomo controllers in H2I!

h2integrate/core/h2integrate_model.py
```diff
@@ -1208,11 +1208,16 @@ def connect_technologies(self):
...
+                # Only connect dispatch rules if they are defined in the tech_config
+                tech_dispatch_rule = self.technology_config.get(tech_name, {}).get(
+                    "dispatch_rule_set", False
                 )
+                if tech_dispatch_rule:
+                    # Connect the dispatch rules output to the dispatching_tech_name input
+                    self.model.connect(
+                        f"{tech_name}.dispatch_block_rule_function",
+                        f"{dispatching_tech_name}.dispatch_block_rule_function_{tech_name}",
+                    )
```
jaredthomas68, line 1221: I'm not sure the logic here is needed if we can just leave the connections out. I think it could be confusing if the connections can be provided and not actually do anything. Can we just require the tech to dispatch connections to only be included if they are needed and can be used?

jaredthomas68 (changes requested):
Looking good. I have a few minor changes and comments.

h2integrate/control/control_strategies/heuristic_pyomo_controller.py
elenya-grant: I definitely think that this could be added in one of the planned future PRs that has some base config for controllers and/or storage performance models.

genevievestarke: Good point! I'll add an issue so we can track it! #609

h2integrate/core/h2integrate_model.py
genevievestarke: I added this check in because the optimal controller uses the plant_config dispatch connections to define what classes to call for the Pyomo dispatch: 
image
But only the heuristic controller uses this dispatch block rule function logic, so the check happens after the tech to dispatch, but before `dispatch_block_rule_function` is connected. 
I'm open to another way to do this, but I thought this was easiest to make the code functional for now!

jaredthomas68: I see, sorry I missed that. Let's leave it for now.

jaredthomas68 (approved):
Looks good to me. Thanks.

PR #591, opened 2026-03-10
reorder system when transporters are used

### LINKED ISSUE #297: Bug report: unnecessary feedback loops in models
# Unnecessary feedback loops in models

Having combiners added inside of financial models groups leads to feedback loops in the openmdao structure since the combiners take in commodity flows and output commodity flows as seen below in the N2 diagram from example 15:

Image

This behavior is not currently causing noticeable problems, but I expect it will as systems become more complex and more optimization techniques are applied.

### Solution idea
I think we need to add the combiners to the `plant` group directly. Since they need to be associated with specific finance models, we can use naming conventions to provide distinctions between which combiners are associated with what finance model subsystems while adding the combiners directly to the `plant` group so they can be executed in the appropriate order at run time and avoiding unnecessary feedback loops. 

The initial solution idea I had was to use `set_order()` to get the systems running in a more physically realistic order. However, the combiners are inside of the finance subsystem making the reordering approach impossible.

## How to reproduce
Run example 15 with the following placed just before the model.run() call:

```python
import openmdao.api as om
om.n2(model.prob)
```

## Relevant output

See N2 diagram above

## System Information

- OS: macOS 15.7.1
  
- Python version: 3.12.11
  
- H2Integrate version: 9eba75f5114cb5d78c98399a9713eaaba610deb5

jaredthomas68:
# Reorder system when transporters are used

Transport models are added at the end of system setup because they need information from technology and financial model setup. However, adding transporters last makes them appear last in the system which in turn leads to unnecessary feedback loops in the system. The unnecessary feedback loops are confusing when debugging a system and can lead to extra computations and degraded solver convergence. This PR executes a reordering command after adding the transporters so that they appear in the correct place in the system.

I also removed an unnecessary cable that was in the steel example (example 1)

Example system before this PR:
image

Example system with the corrected ordering:
image

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation

## Section 3: Related Issues

Resolves #297 

- `examples/01_onshore_steel_mn/plant_config.yaml`: removed unused combiner to steel connection
  - `connect_technologies`: added code to reorder model if transporters are added to system that were not preexisting technology models.
  - `test_system_order`:  added test to check that system order is as expected

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:

### DIFF
--- a/h2integrate/core/h2integrate_model.py
+++ b/h2integrate/core/h2integrate_model.py
@@ -995,8 +995,8 @@ class H2IntegrateModel:
                     # Add the connection component to the model
                     self.plant.add_subsystem(connection_name, connection_component)
 
-                    # reorder the subsystems so that the transporter comes after the source technology
-                    # NOTE: the private method must be used here because setup() has not yet been called
+                    # reorder the subsystems so transporters comes after their source technology
+                    # NOTE: the private method must be used because setup() has not been called
                     subsystem_names = list(self.plant._static_subsystems_allprocs)
                     subsystem_names.remove(connection_name)
                     insert_idx = subsystem_names.index(source_tech) + 1
--- a/h2integrate/core/h2integrate_model.py
+++ b/h2integrate/core/h2integrate_model.py
@@ -995,7 +995,7 @@ class H2IntegrateModel:
                     # Add the connection component to the model
                     self.plant.add_subsystem(connection_name, connection_component)
 
-                    # reorder the subsystems so transporters comes after their source technology
+                    # Reorder the subsystems so transporters comes after their source technology
                     # NOTE: the private method must be used because setup() has not been called
                     subsystem_names = list(self.plant._static_subsystems_allprocs)
                     subsystem_names.remove(connection_name)
--- a/h2integrate/core/h2integrate_model.py
+++ b/h2integrate/core/h2integrate_model.py
@@ -995,16 +995,12 @@ class H2IntegrateModel:
                     # Add the connection component to the model
                     self.plant.add_subsystem(connection_name, connection_component)
 
-                    # get order in which subsystems have been added to the plant
-                    # the private method must be used here because setup() has not yet been called
+                    # reorder the subsystems so that the transporter comes after the source technology
+                    # NOTE: the private method must be used here because setup() has not yet been called
                     subsystem_names = list(self.plant._static_subsystems_allprocs)
-                    # temporarily remove the connection component from the list of names
                     subsystem_names.remove(connection_name)
-                    # get the desired index of the connection component (directly after its source)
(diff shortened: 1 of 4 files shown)


h2integrate/core/h2integrate_model.py
```diff
@@ -995,6 +995,18 @@ def connect_technologies(self):
...
+                    # the private method must be used here because setup() has not yet been called
+                    subsystem_names = list(self.plant._static_subsystems_allprocs)
+                    # temporarily remove the connection component from the list of names
+                    subsystem_names.remove(connection_name)
+                    # get the desired index of the connection component (directly after its source)
+                    insert_idx = subsystem_names.index(source_tech) + 1
+                    # insert the connection component directly after its source
+                    subsystem_names.insert(insert_idx, connection_name)
+                    # update the subsystem order in the plant
+                    self.plant.set_order(subsystem_names)
```
RHammond2:
```suggestion
                    # reorder the subsystems so that the transporter comes after the source technology
                    # NOTE: the private method must be used here because setup() has not yet been called
                    subsystem_names = list(self.plant._static_subsystems_allprocs)
                    subsystem_names.remove(connection_name)
                    insert_idx = subsystem_names.index(source_tech) + 1
                    subsystem_names.insert(insert_idx, connection_name)
                    self.plant.set_order(subsystem_names)
```

Typically we'd use an extra line between code and comments about the next block of code, but that would excessively break up this code block. Then I realized these comments were a bit of a long way to say we're just changing the order, and it was lot easier to follow once the core information was placed in front. Feel free to update the exact wording, but I think there were too many inline comments to explain a fairly straightforward process. This is also a long winded way to say: focus more on the why than what of the code for inline comments and relegate the detailed explanation of what's happening to the docstrings where possible.

RHammond2 (approved):
This makes sense to me, @jaredthomas68. I left an optional comment (hence, the approval status) to pare back some of the inline comments. Once the tests pass, though I'm not sure on what's causing the issues, this should be good to go from my perspective.

johnjasa (approved):
Thanks for this PR! I think it's a great fix. It took me a sec to understand why this needs to occur in the `connect_technologies` method and within this logic loop, but now I understand the subsystem naming indexing better. Great stuff!

PR #605, opened 2026-03-13
Add the option for the storage charge from the grid in pyomo load following dispatch

genevievestarke:
# Add the option for the storage charge from the grid in pyomo load following dispatch

This PR enables the storage in `OptimizedDispatchController` to buy the commodity to charge. Previously, the storage could only charge if there was commodity available in the `commodity_in` variable. Now, if you use the new `allow_commodity_buying` flag, you can allow the storage to charge by drawing the commodity from somewhere else. In order to use this capability, you also need to provide the following additional inputs:

- `commodity_buy_price`: can be a float or list that holds the price of buying the commodity for charging the storage. This is used in the optimization objective function
- `max_system_capacity`: this is a single number, in commodity amount units, that describes the maximum amount of the commodity that the system can use at a given time (e.g. interconnection limit in electricity). This is to prevent the storage using the commodity to charge and the plant using the commodity to meet demand if that exceeds the maximum system capacity.

These inputs are only necessary if `allow_commodity_buying`=True, and are given default values so they are not always required inputs.

Additionally, this PR renames `commodity_met_value` to `demand_met_value` to make it more clear what it is used for, and enables `demand_met_value` to be a single value or a list. This functionality was not in the previous pyomo implementation and enables things like weighting meeting the load differently for different times of day (e.g. rating meeting the load higher during peak times).

### TODO:
- [ ] Add a grid component to the grid buying example
- [x] Add tests for different price situations (e.g. all positive, all negative, and zero)
- [x] Decide on adding commodity_purchased as an output
(We decided not the implement this third bullet point because we want the storage output to be _just_ from the storage, not the full system output)

## NEW TO DO (5/8/2026):

- [x] Remove grid charging from `optimized_pyomo_controller.py`
- [x] Make grid charging a separate controller in `optimized_load_following_with_buying.py`
- [x] Add new controller to H2I framework supported models list
- [ ] Adjust Example 30 to work with new grid charging model (`allow_commodity_buying` should no longer be an input and should be set in the config).
- [ ] Update tests to use commodity buying controller: `test_optimal_control_config_with_commodity_buying`, `test_optimal_control_with_commodity_buying_generic_storage`, `_setup_commodity_buying_problem`, `_run_standard_commodity_buying_assertions`, `test_optimal_control_commodity_buying_all_positive_prices`, `test_optimal_control_commodity_buying_all_negative_prices`, and `test_optimal_control_commodity_buying_zero_prices` tests in `test_optimal_controllers.py`
- [ ] Write commodity buying controller docs

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [ ] Documentation
- [ ] `CHANGELOG.md`

### Section 4.1: New Files
- `path/to/file.extension`

- `path/to/file.extension`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


examples/30_pyomo_optimized_dispatch/tech_config.yaml
```diff
@@ -66,7 +66,10 @@ technologies:
       control_parameters:
         cost_per_charge: 0.03  # in $/kW, cost to charge the storage (note that charging is incentivized)
         cost_per_discharge: 0.05  # in $/kW, cost to discharge the storage
-        commodity_met_value: 0.1  # in $/kW, penalty for not meeting the desired load demand
```
elenya-grant, line 66: I like this name change!

h2integrate/control/control_rules/converters/generic_converter_min_operating_cost.py
```diff
@@ -65,8 +65,12 @@ def initialize_parameters(self, inputs: dict, dispatch_inputs: dict):
         Args:
             inputs (dict):
                 Dictionary of numpy arrays (length = self.n_timesteps) containing at least:
-                    f"{commodity}_in"       : Available generated commodity profile.
-                    f"{commodity}_demand"   : Demanded commodity output profile.
+                    f"{commodity}_in"           : Available generated commodity profile.
+                    f"{commodity}_demand"       : Demanded commodity output profile.
+                    f"{commodity}_met_value_in" : Variable weight for meeting the load
+                    if allow_grid_charging:
+                        f"{commodity}_buy_price_in"   : Variable cost of energy from the grid
```
elenya-grant: could you update the doc string so that it doesn't explicitly say "grid"? Maybe like "Variable cost of commodity bought from external source"

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
```diff
@@ -250,6 +271,22 @@ def _create_parameters(self, pyomo_model: pyo.ConcreteModel, t):
...
+        if self.allow_commodity_buying:
+            pyomo_model.commodity_buy_price = pyo.Param(
+                doc=f"Commodity buy price per generation [$/{self.commodity_storage_units}]",
+                default=0.0,
+                within=pyo.Reals,
+                mutable=True,
+                units=self.cost_units_per_amount_pyo,
+            )
+            pyomo_model.load_production_limit = pyo.Param(
+                doc=f"Grid inport limit for load [$/{self.commodity_storage_units}]",
```
elenya-grant: Can you update the doc here to say "External commodity source import limit for load"?

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
```diff
@@ -383,6 +428,17 @@ def _create_constraints(self, pyomo_model: pyo.ConcreteModel, t):
             expr=pyomo_model.commodity_out
             <= pyomo_model.commodity_load_demand * pyomo_model.is_generating,
         )
+        # Add variables that allow charging from the grid
+        # NOTE: This constraint prevents buying and selling to the grid at the same time
+        if self.allow_commodity_buying:
+            pyomo_model.purchases_transmission_limit = pyo.Constraint(
```
elenya-grant: is there a way to rename the `purhcases_transmission_limit` to something more general? Currently it gives off "grid" vibes.

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
...
+                    )
+            if isinstance(self.commodity_buy_price, list):
+                if all(self.commodity_buy_price) == 0:
+                    raise ValueError(
+                        "commodity_buy_price must be defined as an input and >0 \
+                            if using grid charging"
+                    )
+
+            # Check max system capacity
+            if self.max_system_capacity == 0:
```
elenya-grant: I'm a little confused on the `max_system_capacity` input and what it entails. I think it seems very close to the variable names for the storage sizes, so I think we should brainstorm alternative names. Like `commodity_buy_max_capacity` or something?

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
             The cost per unit of charging the storage (in $/commodity_rate_units).
         cost_per_discharge (float):
             The cost per unit of discharging the storage (in $/commodity_rate_units).
-        commodity_met_value (float):
+        demand_met_value (float):
```
kbrunik: Looks like the description should be update for the additional types allowed

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -138,6 +180,23 @@ def setup(self):
...
+            "demand_met_value",
+            val=self.config.demand_met_value,
+            shape=self.n_timesteps,
+            units="USD/" + self.config.commodity_rate_units,
+            desc="Value of meeting the demand",
+        )
+
+        if self.config.allow_commodity_buying:
+            self.add_input(
+                f"{self.config.commodity}_buy_price",
```
elenya-grant, line 184: should the maximum about of commodity available to buy also be an input? Like 
```python
self.add_input(f"{self.config.commodity}_available_to_buy", val=self.config.max_system_capacity, shape=self.n_timesteps, units=self.config.commodity_rate_units)
```

Should there be an output from the **controller** that outputs how much commodity was bought from an external source?
```python
self.add_output(f"{self.config.commodity}_bought", val=0, shape=self.n_timesteps, units=self.config.commodity_rate_units)
```

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
...
     discharge_efficiency: float = field(validator=range_val(0, 1), default=None)
     cost_per_production: float = field(default=None)
     cost_per_charge: float = field(default=None)
     cost_per_discharge: float = field(default=None)
-    commodity_met_value: float = field(default=None)
+    demand_met_value: int | float | list = field(default=None)
     time_weighting_factor: float = field(validator=range_val(0, 1), default=0.995)
     time_duration: float = field(default=1.0)  # hours
+    # Can we set this to interconnection? do we want to?
+    max_system_capacity: float = field(default=0)
```
kbrunik: Could we add a validator on this? Perhaps`validator=gte_zero`?

elenya-grant: could `max_system_capacity` also be a list or array?

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
...
+    commodity_buy_price: int | float | list = field(default=0.0)
```
elenya-grant: This is perhaps a personal preference, but for attributes that are only needed if some other attribute is True, I like to default the value to None instead of zero. Like - would it be okay if someone wanted to run this with `allow_commodity_buying` as True and also have the buy price of the commodity be zero?

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
...
+        if self.allow_commodity_buying:
+            # Check grid buy price
+            if isinstance(self.commodity_buy_price, float | int):
+                if self.commodity_buy_price == 0:
+                    raise ValueError(
+                        "commodity_buy_price must be defined as an input and >0 \
+                            if using grid charging"
+                    )
+            if isinstance(self.commodity_buy_price, list):
+                if all(self.commodity_buy_price) == 0:
```
elenya-grant: could do a `if sum(self.commodity_buy_price)==0` instead. But if you go the route of defaulting these to None instead of zero, then you could just check if they `are None`.

Could commodity buy price be negative? Or should it always be positive?

elenya-grant (commented):
I think this is looking good! I left a couple nit-pick comments on variable names and doc strings, etc. I also had a few questions that I'm looking forward to your response on. This is probably obvious, but I'd like to see some tests before this is merged in :) But I think this is looking good! Thanks for adding this in!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -53,38 +53,80 @@ class OptimizedDispatchControllerConfig(PyomoControllerBaseConfig):
...
+        #         self.system_commodity_interface_limit
+        #     ] * self.n_control_window
+
+        # Check inputs for grid parameters
+        if self.allow_commodity_buying:
+            # Check grid buy price
+            if isinstance(self.commodity_buy_price, float | int):
+                if self.commodity_buy_price == 0:
+                    raise ValueError(
+                        "commodity_buy_price must be defined as an input and >0 \
```
kbrunik: Could we update these errors to say "if using commodity buying"?

h2integrate/control/control_strategies/optimized_pyomo_controller.py
```diff
@@ -138,6 +180,23 @@ def setup(self):
 
         self.dispatch_inputs = self.config.make_dispatch_inputs()
 
+        self.add_input(
```
kbrunik, line 174: Is there a particular reason to add this as an openmdao input vs just pulling it from the config when needed?

kbrunik, line 174: Do you think it would be possible to move all the `add_inputs()` together? I think it just makes it easier to read

examples/30_pyomo_optimized_dispatch/tech_config.yaml
```diff
@@ -11,7 +11,7 @@ technologies:
       wind_speed: 9.
     model_inputs:
       performance_parameters:
-        num_turbines: 25
+        num_turbines: 15
```
kbrunik: Is there a reason the number of turbines is changing?

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
```diff
@@ -520,15 +616,27 @@ def _create_hybrid_port(self, hybrid_model: pyo.ConcreteModel, tech_name: str):
...
-        tech_port = Port(
-            initialize={
-                "system_production": hybrid_model.system_production,
-                "system_load": hybrid_model.system_load,
-                "commodity_out": hybrid_model.commodity_out,
-                "charge_commodity": hybrid_model.charge_commodity,
-                "discharge_commodity": hybrid_model.discharge_commodity,
-            }
-        )
+        if self.allow_commodity_buying:
```
kbrunik, line 627: you could reduce how many lines/make it a little clearer what the if statement is doing by setting this dictionary and then initializing the port sort of like this
```
if self.allow_commodity_buy:
    port_dic["commodity_bought"] = hybrid_model.commodity_bought

tech_port = Port(initialize= port_dic)
```

examples/30_pyomo_optimized_dispatch/tech_config.yaml
kbrunik: Agreed!

kbrunik (commented):
Thanks Gen for making all of these changes! It seems like a great step forward in capabilities and exciting that you can now weight certain "high-demand" periods higher in the optimization as well as buy commodity for dispatch!

I would love to see some tests, both with electricity and another commodity. Probably would be good to test that you can't buy more than is allowed. 

One other thing I'm thinking about is that we need to think about how this gets handled on the financial model side because we'll need to account for the purchased energy. It might require a few additional outputs from the controller model so that we can connect it with maybe a feedstock cost model.

h2integrate/control/control_strategies/optimized_pyomo_controller.py
elenya-grant: The reason I think this would be needed is so that we can hook it up to other components.  We'd want to connect other components so that commodity bought from an external source could be included in the costs. I would like to expand example 30 to include a grid buying component and add another finance subgroup to test the integration of this new feature with other h2i components.

examples/30_pyomo_optimized_dispatch/tech_config.yaml
genevievestarke: This was to force the battery to charge from the grid for testing. I think I fixed the bug, so we can move this back to 25 for the final form of the PR!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: `commodity_buy_price` can be negative, which is why I took out the sum and made it all (I usually use a sinusoidal varying price to check, so it's possible that it could sum to zero). I'll change it to default to None!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: I added it as an openmdao input, as well, to enable the user to set them using get_val. I think you wouldn't be able to access them in the run script if it wasn't an openmdao input. These are variables that I think are easier to load from a csv or calculate in the run script rather than only allow them to be input as a yaml file. I'm open to suggestions/opinions on this, though!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: I think we could definitely add an output! Though I'm not sure how to integrate that with the rest of the dispatch structure right now. Ideally, the electricity bought would be folded into the unmet_demand, I think. unmet_demand = (commodity_in - demand_in + storage_commodity_out)

elenya-grant: I'd be happy to chat about integrating it and what that means! Maybe we need to work on resolving Issue #521 before getting too in the weeds. I think we gotta be careful about what's an input/output of controllers vs storage performance models (I already don't think the storage performance model should be outputting unmet demand, etc). But - I think that can be saved for a future PR but for now, I do think it'd be good to have those inputs and outputs to the controller.

h2integrate/control/control_strategies/optimized_pyomo_controller.py
elenya-grant: oooh that makes sense! Thanks!

kbrunik: On this note: I think it would be really good to separately test cases when the price is positive, negative and zero. This would help to make sure it does what we expect in each instance.

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
genevievestarke: This is pretty slick, thank you for the suggestion!!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: Done!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: Done!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: Done!

h2integrate/control/control_rules/converters/generic_converter_min_operating_cost.py
genevievestarke: Done!

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
genevievestarke: Done!

h2integrate/control/control_rules/storage/pyomo_storage_rule_min_operating_cost.py
genevievestarke: Done!

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: I renamed it to `commodity_import_limit`, is that more intuitive?

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: Yes, we could make it an array! (But that would require some code structure changes) That change might be better in a follow-on PR, though

h2integrate/control/control_strategies/optimized_pyomo_controller.py
genevievestarke: I currently don't allow that. I suppose we could, but if `commodity_buy_price` is 0, then that term is always zero in the objective function, so it's the same (or should be) as setting `allow_commodity_buying` to False

johnjasa:
@genevievestarke, @kbrunik, and I had a heart-to-heart about this PR and how it looks in the newer SLC paradigm. Based on this discussion, along with our other pressing Q3 work, we landed on closing this PR. We will take the lessons learned from this work and apply it to H2I after SLC comes in. Thank you all for your patience, review, input, and iteration here!

PR #600, opened 2026-03-12
Bugfix in Charge/Discharge efficiency handling in `StoragePerformanceModel`

elenya-grant:
# Bugfix in Charge/Discharge efficiency handling in `StoragePerformanceModel`

Fixed the way that charge and discharge efficiencies are applied in the `StoragePerformanceModel`. This came up during PR #589. 

This PR fixes how charge/discharge efficiencies are represented in the `StoragePerformanceModel` and I've added subtests to better check that charge and discharge efficiencies are being applied as expected. The way that I think about handling this is a very physical system. Pretend that our storage is a tank. There is a tube going into the tank (charge), and a tube coming out of the tank (discharge). Each tube has a small leak. 

When we are filling the tank (charging), the amount we put into the tube is greater than the amount that gets to the tank (because of the leak). The amount input into the tube > amount input to tank. Equivalently, the SOC sees the input amount with an applied efficiency loss, the amount we put into the tube does not. 

When we are releasing the tank (discharging), the amount we take out of the tank is greater than the amount output from the tube (because of the leak). The amount out from tank > amount out of tube. Equivalently, the SOC sees the output amount without an applied efficiency loss, the amount we get from the tube does. 

## Section 1: Type of Contribution

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [x] Fill out TODO list steps
- [x] Describe requested feedback from reviewers on draft PR
- [x] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [x] Update inline comments and docstring to better clarify the difference in `soc` and `storage_commodity_out_timesteps` calculations with respect to efficiencies

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:
- @kbrunik @genevievestarke any good ideas on how to use inline comments to explain the discharge/charge efficiency stuff? The tank example is the best I can do to explain it in words but I'm finding it hard to generalize the language. 
- @jaredthomas68, can you check the calculations for the `actual_charge` and `actual_discharge` variables in the `simulate()` method of `StoragePerformanceModel`?

**Other feedback**:

## Section 3: General PR Checklist

[...]

PR #589, opened 2026-03-10
Sync open-loop controllers and corresponding storage models

elenya-grant:
# Sync open-loop controllers and corresponding storage models

**This PR is slight on-hold until we:**
1. Add in tests for the Demand Open Loop Controller with <1 charge and discharge efficiencies
2. Update the generic storage Pyomo model to correctly account for charge/discharge efficiencies (PR #600) and add subtests.

**But high-level feedback (listed in Section 2 under the "Type of Reviewer Feedback Requested" section) is still appreciated at this time.**

The main changes in this PR are:
- standardized definition of `commodity_set_point` of the two open-loop storage controllers and their corresponding storage performance models (Resolves Issue #498)
- sync up the workflow of the storage performance models that are used with the open-loop storage controllers to better align with the storage performance models that are used with pyomo controllers
- begin to standardize inputs and outputs of open-loop controllers and storage models

This PR helps standardize what calculations and outputs are done by the controllers vs what calculations and outputs are done in the storage performance models. In summary, outputs are:
- Open-loop controllers compute and output the `commodity_set_point`, which is equivalent to the `storage_commodity_out` output of the `pyomo_disapatch_solver` method in the pyomo controllers (input to the `simulate` method of the storage models as `storage_dispatch_commands`). The `commodity_set_point` is the control command telling the storage to charge (negative values) or discharge (positive values)
- Storage performance models output the SOC, unmet demand, excess commodity, the output from the storage only (similar to `commodity_set_point` but may differ if the storage performance model has losses or degradation) and the combined commodity out (input - charge + discharge then saturated to the demand). 

Note: this PR changed some output names for the open-loop controllers (to align with the output names in the storage models that use pyomo controllers). User's may have to update any examples that use H2I. Examples of the name changes are:
- `battery.electricity_unused_commodity` -> `battery.unused_electricity_out`
- `battery.electricity_unmet_demand` -> `battery.unmet_electricity_demand_out`
- `battery.battery_soc` -> `battery.SOC`

The planned next-step of this work is to:
- integrate the use of open-loop controllers with the storage models that only use pyomo controllers (pysam battery and generic storage)
- integrate the use of pyomo controllers with the storage models that only use open loop controllers
- make a storage performance model baseclass and baseconfig that can be inherited by all storage performance models
- make an open loop storage controller baseclass that can be shared with the pass-through open loop controller and the demand open loop controller

## Section 1: Type of Contribution

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
[...]

Issue #566, opened 2026-03-04
Re-define the `commodity_set_point` output of the open-loop controllers and their corresponding performance models


johnjasa:
Re-define the commodity_set_point output of the open-loop controllers and their corresponding performance models, where commodity_set_point is defined as an array where negative values indicate a "charge" command and positive values indicate a "discharge" command (this definition is equivalent to the storage_dispatch_commands input to the simulate() method in the PySAM battery model)

Issue #498, opened 2026-02-03
Standardize use of controller output `commodity_set_point` across storage models


elenya-grant:
# Standardize use of controller output `commodity_set_point` across storage models

This came up as I was working on PR #493.

The relevant control strategies are:
- `DemandOpenLoopStorageController`: in `h2integrate/control/control_strategies/storage/demand_openloop_controller.py`
- `PassThroughOpenLoopController`: in `h2integrate/control/control_strategies/passthrough_openloop_controller.py`

The relevant storage performance models are:
- `StorageAutoSizingModel` in `h2integrate/storage/simple_storage_auto_sizing.py`
- `SimpleGenericStorage` in `h2integrate/storage/simple_generic_storage.py`

Examples 01, 02 and 12 use the `StorageAutoSizingModel` with the `PassThroughOpenLoopController`. Examples 03 (`co2_hydrogenation_doc`), 14, 16, 19, and 24 use the `SimpleGenericStorage` with the `DemandOpenLoopStorageController`. There are no examples/tests where `StorageAutoSizingModel` is used with `DemandOpenLoopStorageController` or where `SimpleGenericStorage` is used with `PassThroughOpenLoopController`. This is good because the storage performance models will only operate as expected if they're tied to the controller. 

 The `PassThroughOpenLoopController` outputs `commodity_set_point` (which is equal to `commodity_in`). The `commodity_set_point` input to the `StorageAutoSizingModel` is used as the commodity available to input to the storage model.  The `commodity_out` output of the of the `StorageAutoSizingModel` is calculated as `commodity_set_point - storage_charged + storage_discharged`.

The `DemandOpenLoopStorageController` outputs `commodity_set_point` as `commodity_in - storage_charged + storage_discharged`, aka the output timeseries of the available input commodity and storage used to meet the demand. The `SimpleGenericStorage` acts as a pass-through storage component, where the output `commodity_out` is equal to the input `commodity_set_point`.

If the `commodity_set_point` from the `DemandOpenLoopStorageController` was input to the `StorageAutoSizingModel`, the `StorageAutoSizingModel` would not behave appropriately because the `commodity_set_point` from the `DemandOpenLoopStorageController` is the "target" `commodity_out` rather than equivalent to `commodity_in` from the `PassThroughOpenLoopController`. 

[...]

PR #608, opened 2026-03-16
Sync Storage Autosizing Model and Pass-Through controller to Align with Pyomo Controllers

elenya-grant:
# Sync Storage Autosizing Model and Pass-Through controller to Align with Pyomo Controllers

This PR is half of the changes that were proposed in PR #589 but includes the corrected handling of charge and discharge efficiencies (in PR #600). This PR syncs the `StorageAutoSizingModel` storage performance model and the `PassThroughOpenLoopController` to reflect the proposed framework (outlined in PR #589) to distinguish storage performance models and controllers.

This PR helps standardize what calculations and outputs are done by the controllers vs what calculations and outputs are done in the storage performance models. In summary, outputs are:
   - Open-loop controllers compute and output the `commodity_set_point`, which is equivalent to the `storage_commodity_out` output of the `pyomo_disapatch_solver` method in the pyomo controllers (input to the `simulate` method of the storage models as `storage_dispatch_commands`). The `commodity_set_point` is the control command telling the storage to charge (negative values) or discharge (positive values)
   - Storage performance models output the SOC, unmet demand, excess commodity, the output from the storage only (similar to `commodity_set_point` but may differ if the storage performance model has losses or degradation) and the combined commodity out (input - charge + discharge then saturated to the demand). 

Note: this PR changed some output names for the open-loop controllers (to align with the output names in the storage models that use pyomo controllers). User's may have to update any examples that use H2I. Examples of the name changes are:
- `battery.electricity_unused_commodity` -> `battery.unused_electricity_out`
- `battery.electricity_unmet_demand` -> `battery.unmet_electricity_demand_out`
- `battery.battery_soc` -> `battery.SOC`

After this PR, similar updates will be made to the `DemandOpenLoopStorageController` and the `SimpleGenericStorage`.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Add subtests for charge/discharge profiles
- [x] Add subtests for unmet demand, excess commodity, etc
- [x] Add test for `StorageAutoSizingModel` integrated with `PassThroughOpenLoopController`

**Structural feedback**:

**Implementation feedback**:
- I think the file `h2integrate/control/control_strategies/passthrough_openloop_controller.py` should be moved to `h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py`. Thoughts?
- Should I add `max_discharge_rate` as an output to the `StorageAutoSizingModel`? Should it be the same as the charge rate or be calculated based on the discharge efficiency?
- Thoughts on usage of the new config attribute `set_demand_as_avg_commodity_in` instead of the old logic in `StorageAutoSizingModel`? I like it because it seems more explicit.
- Does the usage of `dt_hr` in the `simulate()` method of `StorageAutoSizingModel` mean that we have to be running at hourly timesteps or that the storage capacity has to be defined in `commodity_rate_units*h`? 
- @jaredthomas68: can you check that the `storage_max_fill_rate` and `storage_max_unfill_rate` variables calculated at the start of the `compute()` method in `StorageAutoSizingModel` are calculated properly to account for the charge/discharge efficiencies?

**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 3: Related Issues

- Resolves **half of** Issue #498 (which is a step in Issue #564)
- Related to (but does not resolve) Issue #566 

### Section 4.1: New Files
- `h2integrate/storage/test/test_storage_autosizing.py`: new unit and regression tests for `StorageAutoSizingModel`

- updated the tech config files for Example 1, 2, and 12.
- `h2integrate/storage/simple_storage_auto_sizing.py`
   - `StorageSizingModelConfig`
      - added inputs to align with other storage performance models, such as ....
      - added `set_demand_as_avg_commodity_in` as an attribute. This is being used to replace the previous logic that assumed the demand was the mean of the commodity in IF the sum of the `commodity_in` profile was zero.
   - `StorageAutoSizingModel`
      - added outputs to align with other storage performance models, such as `unmet_{commodity}_demand_out`, `unused_{commodity}_out`, `SOC`, `storage_{commodity}_discharge`, `storage_{commodity}_charge`, and `storage_{commodity}_out`.
      - added `storage_duration` as an output
- `h2integrate/control/control_strategies/passthrough_openloop_controller.py`
   - `PassThroughOpenLoopControllerConfig`
      - added `set_demand_as_avg_commodity_in` as an attribute. This is being used to align with the logic in `StorageAutoSizingModel` to calculate the `commodity_set_point` from the demand profile and input commodity profile.
   - `PassThroughOpenLoopController`
   
- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/control/control_strategies/passthrough_openloop_controller.py
```diff
@@ -7,18 +7,35 @@
...
+            the same units as `commodity_rate_units`. May be a scalar for constant
+            demand or a list/array for time-varying demand.
+            Only used if `set_demand_as_avg_commodity_in` is False. Defaults to 0.
+
+    """
+
     commodity: str = field()
     commodity_rate_units: str = field()
+    demand_profile: int | float | list = field(default=0.0)
+    set_demand_as_avg_commodity_in: bool = field(default=True)
```
genevievestarke: This input is a bit confusing to me. What are the situations where we would want the demand set to the average of the commodity coming in? 
It would make more sense in my head if the demand was set to the demand profile, if provided, and then defaults to the average commodity in if no demand profile is given, rather than the other way around. But I'm interested in others' thoughts on this point!

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
```diff
@@ -53,40 +80,28 @@ def setup(self):
 
     def compute(self, inputs, outputs):
         """
-        Pass through input to output flows.
+        Simple controller.
 
         Args:
             inputs (dict): Dictionary of input values.
                 - {commodity}_in: Input commodity flow.
+                - {commodity}_demand: Commodity demand profile.
```
genevievestarke, line 103: Should `set_demand_as_avg_commodity_in` be an input here, too, then?

h2integrate/control/test/test_openloop_controllers.py
```diff
@@ -73,28 +76,30 @@ def test_pass_through_controller(subtests):
...
-                show_only_incorrect=False,
-                out_stream=None,
-            )
+        expected_set_point = np.mean(np.arange(10)) - np.arange(10)
+        assert (
+            pytest.approx(prob.get_val("hydrogen_set_point", units="kg/h"), rel=1e-3)
+            == expected_set_point
         )
 
+    # # Run the test
```
genevievestarke: Remove this from final code?

h2integrate/postprocess/test/test_sql_timeseries_to_csv.py
```diff
@@ -57,7 +57,7 @@ def test_save_csv_all_results(subtests, configuration, run_example_02_sql_fpath)
     res = save_case_timeseries_as_csv(run_example_02_sql_fpath, save_to_file=True)
 
     with subtests.test("Check number of columns"):
-        assert len(res.columns.to_list()) == 34
+        assert len(res.columns.to_list()) == 40
```
genevievestarke, line 60: This is because of the extra inputs?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -1,22 +1,92 @@
...
 
 @define(kw_only=True)
 class StorageSizingModelConfig(BaseConfig):
     """Configuration class for the StorageAutoSizingModel.
 
-    Fields include `commodity`, `commodity_rate_units`, and `demand_profile`.
+    Attributes:
+        commodity (str): name of commodity
+        commodity_rate_units (str): Units of the commodity (e.g., kW or kg/h).
+        min_charge_fraction (float): Minimum allowable state of charge as a fraction (0 to 1).
```
genevievestarke, line 16: Could we change this to min and max soc (e.g. `min_soc_fraction`)? I think this is more consistent with the other methods.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -1,22 +1,92 @@
...
+        max_charge_fraction (float): Maximum allowable state of charge as a fraction (0 to 1).
+        set_demand_as_avg_commodity_in (bool, optional): If True, assume the demand is
+            equal to the mean input commodity. If False, uses the demand input.
+            Defaults to True.
+        demand_profile (int | float | list, optional): Demand values for each timestep, in
+            the same units as `commodity_rate_units`. May be a scalar for constant
+            demand or a list/array for time-varying demand.
+            Only used if `set_demand_as_avg_commodity_in` is False. Defaults to 0.
```
genevievestarke, line 23: I have the same comment as above, I think that the priority of these options should be switched!

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -1,22 +1,92 @@
...
+        charge_efficiency (float | None, optional): Efficiency of charging the storage, represented
+            as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Optional if
+            `round_trip_efficiency` is provided.
+        discharge_efficiency (float | None, optional): Efficiency of discharging the storage,
+            represented as a decimal between 0 and 1 (e.g., 0.9 for 90% efficiency). Optional if
+            `round_trip_efficiency` is provided.
+        round_trip_efficiency (float | None, optional): Combined efficiency of charging and
```
genevievestarke, line 30: Related to Jared's comments on #587, but we should prevent the efficiency from being over-constrained. I'll add to issue #609

genevievestarke: It looks like you address this below by only allowing roundtrip efficiency -or- charge and discharge efficiencies to be defined. I'll keep this hear to keep the reference to the issue on this!

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
 
         self.add_output(
             "max_charge_rate",
             val=0.0,
             shape=1,
-            units=f"{self.config.commodity_rate_units}",
+            units=self.commodity_rate_units,
+        )
+
+        self.add_output(
```
genevievestarke, line 179: I think for auto-sizing, we can keep charge and discharge rates the same. Thoughts?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
+
     def compute(self, inputs, outputs):
         # Step 1: Auto-size the storage to meet the demand
 
         # Auto-size the fill rate as the max of the input commodity
-        storage_max_fill_rate = np.max(inputs[f"{self.commodity}_in"])
+        storage_max_fill_rate = (
+            np.max(inputs[f"{self.commodity}_in"]) / self.config.charge_efficiency
+        )
+        # Auto-size the unfill rate as the max of the input commodity adjusted for
```
genevievestarke: Maybe we could go with "empty rate" to make it more clear?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
+        # discharge efficiency
+        storage_max_unfill_rate = (
+            np.max(inputs[f"{self.commodity}_in"]) / self.config.discharge_efficiency
+        )
+
+        # NOTE: maybe should replace usage of inputs[f"{self.commodity}_in"] - commodity_demand
```
genevievestarke: I agree with this! I think that would make it more consistent to bring in the commodity_set_point variable

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
+        # NOTE: commodity_storage_soc is just an absolute value and is not a percentage.
+        commodity_storage_soc = np.cumsum(inputs[f"{self.commodity}_in"] - commodity_demand)
         minimum_soc = np.min(commodity_storage_soc)
 
         # Adjust soc so it's not negative.
         if minimum_soc < 0:
             commodity_storage_soc = [x + np.abs(minimum_soc) for x in commodity_storage_soc]
 
-        # Calculate the maximum hydrogen storage capacity needed to meet the demand
+        # Calculate the maximum usable hydrogen storage capacity needed to meet the demand
```
genevievestarke: This should just be "storage" and not "hydrogen storage"?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
         commodity_storage_capacity_kg = np.max(commodity_storage_soc) - np.min(
             commodity_storage_soc
         )
 
-        # Step 2: Simulate the storage performance based on the sizes calculated
+        # Calculate the storage capacity to account for SOC limits
+        rated_storage_capacity = commodity_storage_capacity_kg / (
```
genevievestarke: I'm not sure I totally follow this logic. Maybe we can discuss?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
-                    commodity_storage_soc[j - 1]
-                    + desired_commodity_production[j]
-                    - commodity_demand[j]
-                )
 
+        # NOTE: commodity_storage_soc is just an absolute value and is not a percentage.
+        commodity_storage_soc = np.cumsum(inputs[f"{self.commodity}_in"] - commodity_demand)
         minimum_soc = np.min(commodity_storage_soc)
 
         # Adjust soc so it's not negative.
```
genevievestarke, line 291: If we adjust the SOC here so it's above the minimum SOC, could we cut out estimating the initial SOC? Or is that not possible because the min SOC is dependent on the capacity?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +141,343 @@ def setup(self):
...
+
+        # Early return when storage cannot operate: zero capacity or both
+        # charge and discharge rates are zero.
+        if storage_capacity <= 0 or (charge_rate <= 0 and discharge_rate <= 0):
+            soc_timesteps[:] = self.current_soc * 100.0
+            return storage_commodity_out_timesteps, soc_timesteps
+
+        # Pre-compute scalar constants to avoid repeated attribute lookups
+        # and redundant divisions inside the per-timestep loop.
+        charge_eff = self.config.charge_efficiency
```
genevievestarke, line 461: Are these not class variables?

genevievestarke (changes requested):
This is great work, @elenya-grant!! Thank you for pushing standardization of the storage and control models forwards!

To address the implementation feedback:

- I think we can move the controller file to the storage folder
- I vote that the charge and discharge rates are the same for the auto-sizing model. I'm not sure how you would enforce them being different, and I don't think we need to implement that unless it's asked for.
- I had lots of comments about the `set_demand_as_avg_commodity_in` flag. Let's chat about it!
- The time step question I need to think about more, but I wanted to get some initial thoughts up for the review. I'll come back to this!

h2integrate/control/control_strategies/passthrough_openloop_controller.py
elenya-grant: This is because of the "autosizing" storage performance model. Here's some context:
- The pass-through controller in the past just took `commodity_in` and output it as `commodity_set_point`. Now, `commodity_set_point` should be the charge/discharge commands to the storage.
- The autosizing storage performance model used to set the demand as the average of the input file if the sum of the demand profile was zero. The autosizing storage performance model used to calculate/estimate the charge/discharge commands as `demand - commodity_in`. Now, this calculation has to be done in the pass-through controller. 
- Because of the two above points, we want to make sure the demand profile is the same between the autosizing performance model and the pass through controller. I thought specifying a boolean variable was more explicit than replicating the "if sum(inputs[f"{commodity}_demand"])<0" logic that used to exist in the autosizing model to the pass through controller.

We can use the same logic as before and remove this variable, but that logic would have to be in both the pass-through and autosizing performance models. Whichever one is okay with me. We could also remove the defaults for these variables and require them as inputs.

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: its not an input (in the openmdao sense), its more of a configuration option.

h2integrate/postprocess/test/test_sql_timeseries_to_csv.py
elenya-grant: its because of the extra outputs in the storage performance model, such as storage_commodity_out, SOC, charge profile, discharge profile, unmet demand, excess commodity, etc.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: yeah - I think that this would be nice to rename when doing the final step of this sync-up work, which would be to introduce some base config and base class to the storage performance models. I'd like to rename this in a future PR. I could change it in this PR if this is a blocking change, but I'd also want to change it in all the other storage performance models and configuration classes (which may make this PR look larger than it is).

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: Yeah ... I think so, but I was thinking about any case where the discharge efficiency would not be the same as the charge efficiency so maybe we'd want the discharge rate to be larger than the charge rate? I think having the discharge rate as its own output would be good.

I also think that the auto-sized charge and discharge rate are very conservative and could be updated where the charge rate would be sized based the maximum increase in SOC and the discharge rate would be sized as the maximum decrease in SOC, rather than having it be the max input commodity (so I think they could be different). This would be future work (and a separate PR because it'd change test values for examples 1, 2, and 12).

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: smart!

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: We need to estimate the initial SOC because it's needed in the `simulate()` method. Maybe I don't fully understand what you're thinking though?

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: what do you mean by class variables? I also just copied the `simulate()` function from the generic storage pyomo model (using the correct logic in PR 600).

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: renamed to use "empty" instead of "unfill"

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: done!

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: thanks! Updated!

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: I tried to update the doc string in the compute() method to be more clear, happy to discuss though!

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: I made `set_demand_as_avg_commodity_in` a required input and added a warning in the initialization of `PassThroughOpenLoopControllerConfig` if a user provided the demand profile and set `set_demand_as_avg_commodity_in` to True. What are your thoughts on that?

h2integrate/control/test/test_openloop_controllers.py
elenya-grant: removed

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: updated to avoid redundant SOC calculations!

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
```diff
@@ -53,40 +97,41 @@ def setup(self):
...
-
-        # Declare partials sparsely for all elements as an identity matrix
-        # (diagonal elements are 1.0, others are 0.0)
-        self.declare_partials(
-            of=f"{self.config.commodity}_set_point",
-            wrt=f"{self.config.commodity}_in",
-            rows=np.arange(size),
-            cols=np.arange(size),
-            val=np.ones(size),  # Diagonal elements are 1.0
+        if (
```
kbrunik, line 111: Isn't this warning duplicative with the ` __attrs_post_init__(self)` warning?

Also I think we should raise an error rather than a warning.

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
kbrunik, line 1: I think this should be moved to the `storage` folder since it cannot be used by a converter.

h2integrate/storage/hydrogen/h2_storage_cost.py
```diff
@@ -96,7 +96,7 @@ def setup(self):
         )
 
         self.add_input(
-            "max_capacity",
+            "storage_capacity",
```
kbrunik, line 99: Nice!

h2integrate/storage/simple_storage_auto_sizing.py
kbrunik: Is this captured in an issue? Don't want to forget, but I think it could make sense to wait if other models also have to change.

h2integrate/storage/simple_storage_auto_sizing.py
kbrunik: I think it should be calculated to include the charge/discharge efficiency. It seems like a good harmonization of the code for that to be the same across models.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,379 @@ def setup(self):
...
+            the demand is the average of the `commodity_in` profile.
+        6) Calculate the unmet demand, unused commodity, SOC, combined commodity output, etc.
+
+        """
+        # Part 1: Auto-size the storage to meet the demand
+
+        # 1. Auto-size the fill rate as the max of the input commodity adjusted for
+        # charge efficiency
+        storage_max_fill_rate = (
+            np.max(inputs[f"{self.commodity}_in"]) / self.config.charge_efficiency
```
kbrunik: Doesn't this make the max fill rate greater than it actually is? shouldn't this be multipled?

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,379 @@ def setup(self):
...
+            np.max(inputs[f"{self.commodity}_in"]) / self.config.discharge_efficiency
+        )
 
+        # Auto-size the storage capacity to meet the demand as much as possible
+        # 2. Estimate the storage SOC in `commodity_amount_units`
+        # NOTE: commodity_storage_soc is just an absolute value and is not a percentage.
+        # `{self.commodity}_set_point` is negative when charging and positive when discharging,
+        # the negative of `{self.commodity}_set_point` can be used to estimate the SOC
+        # (which increases when charging and decreases when discharging)
+        commodity_storage_soc = np.cumsum(-1 * inputs[f"{self.commodity}_set_point"])
```
kbrunik, line 286: This is an incredible use of NumPy functionality to reduce lines of code!!

kbrunik (changes requested):
Looking good! I can see the incremental progress towards a more harmonious framework. I had a few comments, nothing major. One blocking comment is changing the warning to an error.

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: yes - this is because with Pyomo Controllers (which are not yet compatible with the storage auto-sizing model), the demand profile is set using `set_val()`, so its duplicated in case someone tries to set the demand using set_val instead of initializing it from the config.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: Added to Issue #564

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: I made it so it raises a `ValueError` in the `__attrs_post_init__` of the config, and a warning in the `compute()`

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: I made it so that the charge rate isn't over-sized

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: moved it to the storage folder!

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
```diff
@@ -0,0 +1,137 @@
...
+            user_input_dmd = True if self.demand_profile > 0 else False
+
+        if self.set_demand_as_avg_commodity_in and user_input_dmd:
+            # If using the average commodity in as the demand,
+            # warn users if they input the demand profile
+            msg = (
+                "A non-zero demand profile was provided but set_demand_as_avg_commodity_in is True."
+                " The provided demand profile will not be used, the demand profile will be "
+                f"calculated as the mean of ``{self.commodity}_in``. "
+            )
```
jaredthomas68, line 42: I'm inclined to throw an error here since this could be a significant departure from expected behavior.

jaredthomas68: It's not that hard to not provide a demand profile.

kbrunik (approved):
Thanks for making those changes! It looks good to me :)

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            commodity_demand = inputs[f"{self.config.commodity}_demand"]
+
+        # Assign the set point as the difference between the demand and the input commodity
+        # when demand > input, the set point is positive to command discharging
+        # when demand < input, the set point is negative to command charging
+        outputs[f"{self.config.commodity}_set_point"] = (
+            commodity_demand - inputs[f"{self.config.commodity}_in"]
+        )
```
jaredthomas68, line 135: This is not really the intent of the passthrough controller, the idea was to literally pass out what came in. I think both the actual passthrough control and what you have here are useful, but they are not the same thing.

jaredthomas68: I think we could add this as a new controller, or provide options to allow both the old and the new behavior.

h2integrate/storage/test/test_generic_storage_pyo.py
```diff
@@ -636,7 +636,10 @@ def test_generic_storage_without_controller_with_losses(plant_config, subtests):
         )
 
     with subtests.test("Cumulative charge/discharge does not exceed storage capacity"):
-        assert np.cumsum(prob.get_val("storage_hydrogen_out", units="kg/h")).max() <= capacity
+        assert (
+            np.cumsum(prob.get_val("storage_hydrogen_out", units="kg/h")).max()
+            <= capacity * discharge_eff
```
jaredthomas68, line 641: Thanks for getting these changes in here

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -525,7 +525,7 @@ def simulate(
                 headroom = (soc - soc_min) * storage_capacity / self.dt_hr
 
                 # Clip to the most restrictive limit without applied efficiency.
-                # Efficiency losses occur after energy has left storage.
+                # Efficiency losses occur as energy leaves storage.
```
jaredthomas68:
```suggestion
                # Discharge efficiency losses occur as energy leaves storage.
```

h2integrate/storage/simple_storage_auto_sizing.py
genevievestarke: Sounds good! We can revisit this in the base class, too, as we standardize!

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,385 @@ def setup(self):
...
+            val=0.0,
+            shape=self.n_timesteps,
+            units=self.commodity_rate_units,
+            desc="Unused generated commodity",
+        )
+
+        self.dt_hr = self.dt / 3600  # convert from seconds to hours
+
+    def compute(self, inputs, outputs):
+        """Part 1: calculate the storage sizes (charge rate, discharge rate, and capacity)
```
genevievestarke, line 245: Thank you for this detailed outline of the steps in the code!

h2integrate/storage/simple_storage_auto_sizing.py
genevievestarke: Perfect, looks good!

genevievestarke (approved):
It looks good to me, thank you for addressing the comments :)

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
jaredthomas68: Let's raise an error always if both are provided, and we should only need it in compute. It can be moved later, but I think that is better than duplicating.

h2integrate/storage/simple_storage_auto_sizing.py
jaredthomas68: I think the `demand_profile` and the `set_demand_as_avg_commodity_in` should be mutually exclusive and throw an error if both are provided (True).

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -1,23 +1,91 @@
...
+        of `round_trip_efficiency`.
+        """
+        if self.round_trip_efficiency is not None:
+            if self.charge_efficiency is not None or self.discharge_efficiency is not None:
+                raise ValueError(
+                    "Exactly one of the following sets of parameters must be set: (a) "
+                    "`round_trip_efficiency`, or (b) both `charge_efficiency` "
+                    "and `discharge_efficiency`."
+                )
+
```
jaredthomas68:
```suggestion
        if (self.round_trip_efficiency is not None) and (
            self.charge_efficiency is None and self.discharge_efficiency is None
            ):
```

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
jaredthomas68, line 135: This significantly changes the behavior of the passthrough control. However, I think this is close to the same purpose. I think with the demand component we have now, a strict pass-through controller may not be necessary.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,385 @@ def setup(self):
...
-                    + desired_commodity_production[j]
-                    - commodity_demand[j]
-                )
+        # Auto-size the storage capacity to meet the demand as much as possible
+        # 2. Estimate the storage SOC in `commodity_amount_units`
+        # NOTE: commodity_storage_soc is just an absolute value and is not a percentage.
+        # `{self.commodity}_set_point` is negative when charging and positive when discharging,
+        # the negative of `{self.commodity}_set_point` can be used to estimate the SOC
+        # (which increases when charging and decreases when discharging)
+        commodity_storage_soc = np.cumsum(-1 * inputs[f"{self.commodity}_set_point"])
```
jaredthomas68, line 286: Name `commodity_storage_soc` is confusing since soc is usually a ratio or percentage, not absolute value. maybe just `commodity_stored` would work

jaredthomas68: On further note, it is treated as a ratio later, but commodity_set_point should be in commodity rate units, not a ratio.

jaredthomas68: There's probably something I'm missing here.

jaredthomas68: I think the real issue is unit clarity. I think it would be valuable to append units to each of the variables you are using in the sizing steps, even if just `_ratio` or `_crunits`

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,385 @@ def setup(self):
...
+
+        # 1. Calculate the minimum SOC as a fraction
+        min_soc_fraction = np.min(commodity_storage_soc) / rated_storage_capacity
+        if min_soc_fraction < self.config.min_charge_fraction:
+            # 2. adjust the storage SOC profile so that the minimum SOC equal
+            # to config.min_charge_fraction
+            soc_adjustment = (
+                self.config.min_charge_fraction - min_soc_fraction
+            ) * rated_storage_capacity
+            commodity_storage_soc = commodity_storage_soc + soc_adjustment
```
jaredthomas68: I feel like we should not need this, but I'm not positive. It seems to indicate possible errors elsewhere.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,385 @@ def setup(self):
...
+        """
+        # Part 1: Auto-size the storage to meet the demand
+
+        # 1. Auto-size the fill rate as the max of the input commodity
         storage_max_fill_rate = np.max(inputs[f"{self.commodity}_in"])
+        # Auto-size the empty rate as the max of the input commodity adjusted for
+        # discharge efficiency
+        storage_max_empty_rate = (
+            np.max(inputs[f"{self.commodity}_in"]) / self.config.discharge_efficiency
+        )
```
jaredthomas68, line 279: Because you are using these as inputs to simulate, I think they should not account for efficiencies. The simulate method should be handling that.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -71,147 +139,385 @@ def setup(self):
...
+            rated_storage_capacity,
+        )
+        storage_commodity_out = np.array(storage_commodity_out)
 
-        # Initialize output arrays of charge and discharge
-        discharge_storage = np.zeros(self.n_timesteps)
-        charge_storage = np.zeros(self.n_timesteps)
-        output_array = np.zeros(self.n_timesteps)
+        # 5. Calculate the demand profile
+        if self.config.set_demand_as_avg_commodity_in:
```
jaredthomas68, line 322: should only do this is no demand is provided AND the flag is true

jaredthomas68 (changes requested):
Thanks for doing this @elenya-grant. Looks pretty good. I do have a few changes and comments I'd like addressed before it gets merged.

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: I think if the old functionality of the pass-through controller becomes needed, then we can make a new one that is a proper in=out type of pass-through controller. I could rename "PassThroughOpenLoopController" to "SimpleStorageOpenLoopController" if you want?

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: The current code is exactly copied from `StoragePerformanceModelConfig` in `h2integrate/storage/generic_storage_pyo.py`. Do you want that change in both places?

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: agreed!

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: commodity_set_point is in commodity_rate_units. The only time any of the variables are used as a fraction is when we calculate the `min_soc_fraction` or use the `min_charge_fraction` or `max_charge_fraction` from the inputs. So basically every time there's a ratio, it has `fraction` in the name in the `compute()` method (with the exception of the soc output from the `simulate()` method)

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: What do you mean? This is just to calculate the SOC at the first timestep and ensure that it matches the SOC bounds. I suppose I could do `self.current_soc = np.max([self.config.min_charge_fraction, commodity_storage_soc[0]/rated_storage_capacity])` instead

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: I do think this whole confusion about the demand profile stuff may also be partially addressed in the future PR as we try to figure out how to sync-up all the storage and performance models together. That being said, I will update the `compute()` method to throw an error rather than raise a warning

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: actually maybe I'm confused on the proposed logic. Lemme think about this one.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: Do you want an error thrown if the flag is True and the demand is provided?

h2integrate/storage/generic_storage_pyo.py
elenya-grant: done

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: updated to throw an error in compute()

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: I replaced that all soc adjustment stuff with the line I suggested above.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: @kbrunik - what are your thoughts on applying the discharge efficiency for the discharge rate? I think it should be included so that the storage is conservatively autosized to meet all the demand. I still think that the charge and discharge rate are calculated more conservatively than needed so I don't have strong feelings since I think this calc should be refactored in the future.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: Made it so that an error is thrown if demand is provided and the flag is true

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
kbrunik: I wonder if we could add the `PassTrhoughOpenLoopController` to a dropped feature issue? It seems to me that if there aren't any test values that change in this PR that there was something funky going on the with pass through controller. I agree this seems like a slightly different compute but seems to work with the existing infrastrucuture.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: There are 3 places where an error will be thrown if the user tries to input a non-zero demand profile and the flag is True. (1 in the pass-through controller compute method, 1 in the pass-through controller config attrs_post_init, and 1 in the storage autosizing compute method)

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: removed the discharge efficiency.

kbrunik: I also don't have strong feelings, I feel like this is what was done when it came over from Green Steel but don't think it needs to stay that way. It does seem like a conservative estimate

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
```diff
@@ -119,7 +117,7 @@ def compute(self, inputs, outputs):
                 " The input demand profile will not be used, the demand profile will be "
                 f"calculated as the mean of ``{self.config.commodity}_in``. "
             )
-            warnings.warn(msg, UserWarning)
+            ValueError(msg)
```
jaredthomas68:
```suggestion
            raise ValueError(msg)
```

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
jaredthomas68: I think I agree with the above

h2integrate/storage/simple_storage_auto_sizing.py
jaredthomas68: I get that this is copied over. I didn't like it before either. The proposed behavior is exactly the same as before, but makes it so that the error only needs to be written in the code once.

h2integrate/storage/simple_storage_auto_sizing.py
jaredthomas68: The simulate method is doing the adjustment already, so doing here is applying it twice.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -338,13 +325,23 @@ def compute(self, inputs, outputs):
...
+        # 3. Calculate the demand profile
         if self.config.set_demand_as_avg_commodity_in:
-            commodity_demand = np.mean(inputs[f"{self.commodity}_in"]) * np.ones(self.n_timesteps)
+            if inputs[f"{self.commodity}_demand"].sum() > 0:
+                msg = (
+                    "A non-zero demand profile was input but set_demand_as_avg_commodity_in is "
+                    "True. The input demand profile will not be used, the demand profile will be "
+                    f"calculated as the mean of ``{self.config.commodity}_in``. "
+                )
+                ValueError(msg)
```
jaredthomas68:
```suggestion
                raise ValueError(msg)
```

h2integrate/storage/simple_storage_auto_sizing.py
jaredthomas68: I would like it changed in both places if we can.

jaredthomas68 (changes requested):
Just a couple more things. Thanks for addressing my previous comments.

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: the code here is for _sizing_ - not simulating. Regardless, its removed now and can part of a future discussion.

h2integrate/control/control_strategies/storage/passthrough_openloop_controller.py
elenya-grant: done

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: done

h2integrate/storage/generic_storage_pyo.py
```diff
@@ -81,21 +81,14 @@ def __attrs_post_init__(self):
...
+        ):
             # Calculate charge and discharge efficiencies from round-trip efficiency
             self.charge_efficiency = np.sqrt(self.round_trip_efficiency)
             self.discharge_efficiency = np.sqrt(self.round_trip_efficiency)
-        elif self.charge_efficiency is not None and self.discharge_efficiency is not None:
-            # Ensure both charge and discharge efficiencies are provided
-            pass
-        else:
+            self.round_trip_efficiency = None
+        if self.charge_efficiency is None and self.discharge_efficiency is None:
```
jaredthomas68: Maybe I made a mistake in my previous suggestion, but the current logic does not quite work. I think the error should be raised if either charge or discharge efficiency is None.
```suggestion
            self.round_trip_efficiency = None
        if self.charge_efficiency is None or self.discharge_efficiency is None:
```

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -61,21 +61,14 @@ def __attrs_post_init__(self):
...
+        ):
             # Calculate charge and discharge efficiencies from round-trip efficiency
             self.charge_efficiency = np.sqrt(self.round_trip_efficiency)
             self.discharge_efficiency = np.sqrt(self.round_trip_efficiency)
-        elif self.charge_efficiency is not None and self.discharge_efficiency is not None:
-            # Ensure both charge and discharge efficiencies are provided
-            pass
-        else:
+            self.round_trip_efficiency = None
+        if self.charge_efficiency is None and self.discharge_efficiency is None:
```
jaredthomas68:
```suggestion
        if self.charge_efficiency is None or self.discharge_efficiency is None:
```

jaredthomas68 (changes requested):
So close

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: thanks for catching those - I've fixed them

jaredthomas68 (approved):
Looks good, thanks!

PR #615, opened 2026-03-20
Bugfix: Update storage models so can use different control types per storage type

elenya-grant:
# Bugfix: Update storage models so can use different control types per storage type

This is a bugfix that only applies to `h2integrate/storage/storage_performance_model.py` and `h2integrate/storage/battery/pysam_battery.py`.

The use-case where I encountered this example is not tested in H2I. But - basically I'm using a pyomo controller with the pysam battery and using the `StoragePerformanceModel` with the demand open-loop controller for hydrogen storage. This means that for the `StoragePerformanceModel`, I need the storage model to have the input `hydrogen_set_point`. I need the pysam battery to be input `pyomo_dispatch_solver`. The existing logic in these two controllers lead to a case where the `hydrogen_set_point` was not being added as an input to the hydrogen storage.

The code prior to this PR and how that logic was not working for the hydrogen storage is below:

```python
        if "tech_to_dispatch_connections" in self.options["plant_config"]: # this is True, because of the battery
            self.tech_group_name = self.pathname.split(".")
            for _source_tech, intended_dispatch_tech in self.options["plant_config"][
                "tech_to_dispatch_connections"
            ]:
                if any(intended_dispatch_tech in name for name in self.tech_group_name):
                    self.add_discrete_input("pyomo_dispatch_solver", val=dummy_function) # this is not added as an input because the tech_to_dispatch_connections is for the battery
                    break

       else: # this loop isn't gone into because the if statement evaluated to true
            # using an open-loop storage controller
            self.add_input(
                "electricity_set_point",
                val=0.0,
                shape=self.n_timesteps,
                units="kW",
            )
```

New logic that works:
```python
        using_feedback_control = False
        # create inputs for pyomo control model
        if "tech_to_dispatch_connections" in self.options["plant_config"]:
            # get technology group name
            # TODO: The split below seems brittle
            self.tech_group_name = self.pathname.split(".")
            for _source_tech, intended_dispatch_tech in self.options["plant_config"][
                "tech_to_dispatch_connections"
            ]:
                if any(intended_dispatch_tech in name for name in self.tech_group_name):
                    self.add_discrete_input("pyomo_dispatch_solver", val=dummy_function)
                    # set the using feedback control variable to True
                    using_feedback_control = True
                    break

        if not using_feedback_control:
            # using an open-loop storage controller
            self.add_input(
                f"{self.commodity}_set_point",
                val=0.0,
                shape=self.n_timesteps,
                units="kW",
            )
```

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 3: Related Issues

- Made Issue #620 as part of this PR

### Section 4.1: New Files
- `h2integrate/control/test/test_multistorage_pyomo_openloop_control.py`: new tests to check that the introduced logic works as expected.

- `h2integrate/storage/storage_performance_model.py`: updated logic for setting control input in `setup()`
- `h2integrate/storage/battery/pysam_battery.py`: updated logic for setting control input in `setup()`
- `h2integrate/control/control_strategies/heuristic_pyomo_controller.py`: (bugfix) moved setting the attributes of charge and discharge efficiency from `setup()` to `initialize_parameters()`
- `h2integrate/control/test/test_heuristic_with_generic_storage.py`: updated tech config to test that the bugfix in `heuristic_pyomo_controller.py` worked.
- `h2integrate/control/control_strategies/storage/simple_openloop_controller.py`: had to change `strict=False` when creating config to prevent errors when using the simple openloop controller in cases where the cost model and the performance model have shared parameters that are not shared with the openloop controller.

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


genevievestarke (commented):
Ok, so you're breaking the previous if/else into two if statements, based on whether feedback control is used or not, correct? 
Could you just take away the "break" part of the loop? I'm assuming not, because the if loop will still be triggered by the pyomo tech to dispatch connection?

jaredthomas68 (changes requested):
The logic here seems a little brittle. I'm ok with it for now, but if we keep the logic as is, can you make a test or two (one for each option in the logic maybe)?

One possible fix for the logic is to have a list of controllers that are or are not feedback controllers, or that do or do not need the electricity_set_point input. I don't really like that idea though because it is hard to maintain (like electricity_producing_techs).

jaredthomas68:
> The logic here seems a little brittle. I'm ok with it for now, but if we keep the logic as is, can you make a test or two (one for each option in the logic maybe)?
> 
> One possible fix for the logic is to have a list of controllers that are or are not feedback controllers, or that do or do not need the electricity_set_point input. I don't really like that idea though because it is hard to maintain (like electricity_producing_techs).

Just to clarify, I think tests are always important, but particularly in bug fixes.

h2integrate/storage/storage_performance_model.py
```diff
@@ -306,8 +309,10 @@ def setup(self):
             ]:
                 if any(intended_dispatch_tech in name for name in self.tech_group_name):
```
kbrunik, line 310: I understand that this isn't the point of this PR but it illuminated to me that we have some weird logic in the `plant_config`. I vote that we potentially make an issue about it.

I think it's ambiguous that it says "tech_to_dispatch_connections" but excludes the open-loop controllers. I vote either we update this name to be more clear that this isn't used for open loop dispatch or we add all dispatch to the "tech_to_dispatch_connections". We also need to spell this out in documentation. It's currently not included.

h2integrate/control/test/test_multistorage_pyomo_openloop_control.py
kbrunik, line 1: Thanks for adding all these tests! Great to see all the different combinations :)

kbrunik (approved):

johnjasa (approved):
Thanks for these changes! I think it's a step in a great direction. I appreciate you already adding tests based on Jared's feedback. I do think Kaitlin's suggestion for creating an issue for the `tech_to_dispatch_connections` is useful and reveals a gap in our documentation there already. Could you please create that issue and share your thoughts? Otherwise this is good to come in imo!

h2integrate/storage/storage_performance_model.py
elenya-grant: Made Issue #620

jaredthomas68 (approved):
Looks good to me. Thanks for adding the tests and making the issue.

Issue #616, opened 2026-03-20
`generic_splitter` not functioning properly in `prescribed_commodity` mode


jmartin4u:
# `generic_splitter` not functioning properly in `prescribed_commodity` mode

When the `generic_splitter` is used in `prescribed_commodity` mode with a connected `<commodity_in>` input, the value of the input `"prescribed_commodity_to_priority_tech"` is not being preserved from the `setup()` to the `compute()` step. This input appears to be set properly in `setup()` to be an array with all values equal to the value of the config parameter `prescribed_commodity_to_priority_tech`, but then in `compute()`, it is mysteriously always changing back to an array of ones.

## How to reproduce

I set of a branch to reproduce this issue, since it occurred when running the ammonia synloop in a different repo. In the branch ["splitter-bad2"](https://github.com/jmartin4u/H2Integrate/tree/splitter-bad2), if you run the example `examples/12_ammonia_synloop/run_ammonia_synloop.py`, I have set up print statements to show the mean values of the `prescribed_commodity_to_priority_tech` during both the `setup()` and `compute()`. Running this shows how the array changes back to ones at some point, which causes ammonia production to be severely limited (only 1 MW of electricity is passed to ammonia production, and it needs more).

<img width="349" height="113" alt="Image" src="https://github.com/user-attachments/assets/e1e81833-ecf0-4479-918e-6733a1bce60d" />

I have ALSO created a branch which fixes the problem, called ["splitter-fix2"](https://github.com/jmartin4u/H2Integrate/tree/splitter-bad2). Running the same example, the `prescribed_commodity_to_priority_tech` value is now being preserved in `compute()`. I'm not sure why this fixes it, but the input has to be shaped using `shape=n_timestamps` rather than `shape_by_conn` in the setup():

<img width="889" height="217" alt="Image" src="https://github.com/user-attachments/assets/13f016aa-3033-46b7-9856-5a552ea0dfdc" />
 
With this change, the splitter is working properly and much more sensible LCOAs are being calculated:

<img width="394" height="113" alt="Image" src="https://github.com/user-attachments/assets/9ad72752-d85e-45a5-b200-6fd04ba1253d" />

## System Information

- OS: <Win 11 24H2>
  
- Python version: <3.11.15>
  
- H2Integrate version: <0.7.1>
  
  - Installed from source using an editable installation with developer tools: `pip install -e .[develop]`

### Relevant library versions

absl-py==2.4.0
accessible-pygments==0.0.5
[...]

PR #632, opened 2026-03-26
Removed some shape_by_conn calls due to OM issues

johnjasa:
# Remove shape by conn calls

As noted by @jmartin4u in #616 the splitter is failing to correctly split in certain edge cases; he had a fix and shared it.
Based on some quick checking, there were many behind-the-scenes changes in the connection setup between OM v3.42 and v3.43.
Without digging too much, there's enough changed in OM that I don't think it's worth figuring out _why_ this isn't working; just to use this fix.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes


jmartin4u (approved):
Verified that this fixes splitter operation with openmdao==3.43.0 on my Win11 machine. Added a fix for the `test_pipe_with_hydrogen` that was failing, should be passing now

johnjasa:
Awesome, thanks for verifying and for the test fix, @jmartin4u! I appreciate it.

PR #653, opened 2026-04-03
Framework to allow for variable dt

### LINKED ISSUE #638: Enable grid model to work with non-hourly timesteps
# Enable grid model to work with non-hourly timesteps

Unclear how much effort this would take, if any.
Resolving this issue would involve assessing that effort, addressing it, adding grid to the supported non-hourly timesteps list, and adding any relevant tests for different timesteps.

### LINKED ISSUE #648: Make non-hourly dt error model specific
# Make non-hourly dt model specific
H2I will throw an error regardless of the model if a non-hourly time step is used. 

## Proposed solution
Make the error model-specific using a dictionary with allowed time-step duration range to check against and base the error on.

## Alternatives considered
- individual error statements for each model (difficult to maintain)
- require all models to work on the same time-step ranges (too hard to make manageable chunks of work)

jaredthomas68:
# Framework to allow for variable dt

This PR contains changes to the framework to handle variable dt between different models so we can start integrating model changes one at a time to allow variable time steps.

I also adjusted the grid components to work with variable dt to demonstrate how the framework changes will work with actual models and test the integration.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

Ready for full review, I'm looking for any and all feedback.

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 3: Related Issues

This PR will resolve #648 and makes steps towards resolving #204.

The work on the grid components will also resolve #638. 

- `docs/developer_guide/adding_a_new_technology.md`: variable time step info added
- `docs/intro.md`: variable time step info added
- `docs/technology_models/grid.md`: variable time step info added
- `h2integrate/converters/grid/grid.py`: add 5 m to 1 h time step bounds
- `h2integrate/converters/grid/test/test_grid.py`: unit and integration test for non-hour time step
- `h2integrate/core/h2integrate_model.py`: check time step on a per-model basis
- `h2integrate/core/inputs/validation.py`:  make simulation length check account for variable dt and remove dt check
- `h2integrate/core/model_baseclasses.py`:  add self.dt to cost model base class
- `h2integrate/core/test/test_framework.py`: add unit tests for time step check

This PR is just the first step in a series of expected PRs to allow more flexibility in time step selection.

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/converters/grid/grid.py
```diff
@@ -45,6 +45,9 @@ class GridPerformanceModel(PerformanceModelBaseClass):
         electricity_out (array): Power flowing out of the grid (buying) (kW).
     """
 
+    _time_step_min = 300
+    _time_step_max = 3600
+
```
elenya-grant, line 51: I like how you've added these attributes. I could see a similar thing perhaps being implemented in the future for simulation length. I'd be curious what your (and other people's thoughts) would be on whether these should be two separate variables (like you have) vs one tuple like this instead:
```python
_time_step_bounds = (300, 3600)
```

h2integrate/core/h2integrate_model.py
```diff
@@ -591,8 +596,29 @@ def _process_model(self, model_type, individual_tech_config, tech_group):
...
+        if hasattr(model_object, "_time_step_min") or hasattr(model_object, "_time_step_max"):
+            if dt < model_object._time_step_min or dt > model_object._time_step_max:
+                msg = (
+                    f"Performance model {model_name} is compatible with time steps "
+                    f"between {model_object._time_step_min} and {model_object._time_step_max} "
+                    f"but a time step of {dt} (s) was specified"
+                )
+                raise ValueError(msg)
+
+        elif dt != 3600:
```
elenya-grant: I think an alternative to this `elif` catch would be to add the _time_step_min and time_step_max to the CostModelBaseClass and PerformanceModelBaseClass and have them both default to 3600. I'm not saying this is better - but its an alternative approach.

h2integrate/core/h2integrate_model.py
```diff
@@ -591,8 +596,29 @@ def _process_model(self, model_type, individual_tech_config, tech_group):
             ),
             promotes=["*"],
         )
+
         return om_model_object
 
+    def _check_time_step(self, model_name, model_object):
+        dt = int(self.plant_config["plant"]["simulation"]["dt"])
+
+        if hasattr(model_object, "_time_step_min") or hasattr(model_object, "_time_step_max"):
```
elenya-grant: for some reason I'm wanting to suggest `getattr` instead (since you can provide defaults if it doesnt have the attribute) but I think that'd make the logic more sloppy. So - I like how you have it.

elenya-grant (commented):
Thanks for introducing this PR! I had a couple of high-level suggestions/questions for other folks. I think this is the cleanest and least complicated way of introducing this flexibility - nice job!

Kinda big-picture consideration - should there be a valid time-step interval defined too? Like - for the grid, what if I tried to run a timestep of `563` seconds? Maybe thats fine but it seems weird. Should there be another parameter defined as like `_time_step_interval`? So that models can be run for timesteps:
```python
valid_dts = np.arange(_time_step_min, _time_step_max, _time_step_interval)
```
(not asking for this to be implemented, but a question I think it worth considering)

I know this is still a draft, but I would like to see a test for the grid component using a smaller timestep.

h2integrate/converters/grid/grid.py
elenya-grant: Another question - why can't this have a min timestep of 1 sec? Why 300?

johnjasa (approved):
I like this PR! It is simple and helpful. I'd love to see your answers to @elenya-grant's questions as I had some similar thoughts.

For future work, as we build out more of the timestep handling and methods related to that, I could imagine a separate utilities file may be useful to reduce the size of `h2integrate_model.py`. I think as you work through other changes needed for this timestep capability, keep that in mind!

h2integrate/converters/grid/grid.py
genevievestarke: I think making the min 300 seconds for now makes a lot of sense, since that's the smallest resource step that is available and also the smallest grid price data resolution. I think that letting the time step be 1 second is not realistic for the steady state models we have in H2I. This is something I feel pretty strongly about, but what are others' opinions about this?

h2integrate/core/h2integrate_model.py
```diff
@@ -595,8 +600,29 @@ def _process_model(self, model_type, individual_tech_config, tech_group):
...
```
genevievestarke: I'm a little bit confused by this part. The if/else logic seems to use both values, but you're only checking for one or the other? Can a model just have one defined (i.e. only _time_step_max = 3600, and then any large step is fine)?

genevievestarke (commented):
Thank you for kicking this off, @jaredthomas68!! Just a few thoughts!

h2integrate/converters/grid/grid.py
jaredthomas68: Thank you @elenya-grant. Also, I like the idea of using a tuple.

I agree with @genevievestarke on the minimum time step, at least for now. The 5-minute resource data floor, and the fact we are not modeling any transient behavior, is why I choose to limit the grid to 5-minute minimum time steps.

jaredthomas68: I think the only reason to not use timestep bounds would be so one or the other could be left out. Realistically, I think there will always be upper and lower limits, so I think it will be fine to use a tuple.

jaredthomas68: I switched to using a tuple

h2integrate/core/h2integrate_model.py
jaredthomas68: I started doing this, but it seems like a lot of changes to accomplish a small thing. I think I'll stick with the elif for now.

h2integrate/core/h2integrate_model.py
jaredthomas68: I see the confusion. I switched to using a tuple as suggested by @elenya-grant, which solves the odd logic. I also think that having an upper and lower bound is important, even if they are wildly different (e.g. 300 s, 1 week), just to make sure the behavior is reasonable.

jaredthomas68:
> Thanks for introducing this PR! I had a couple of high-level suggestions/questions for other folks. I think this is the cleanest and least complicated way of introducing this flexibility - nice job!
> 
> Kinda big-picture consideration - should there be a valid time-step interval defined too? Like - for the grid, what if I tried to run a timestep of `563` seconds? Maybe thats fine but it seems weird. Should there be another parameter defined as like `_time_step_interval`? So that models can be run for timesteps:
> 
> ```python
> valid_dts = np.arange(_time_step_min, _time_step_max, _time_step_interval)
> ```
> 
> (not asking for this to be implemented, but a question I think it worth considering)
> 
> I know this is still a draft, but I would like to see a test for the grid component using a smaller timestep.

Thanks for the review and questions. I think I would like to not include valid time step interval yet. We have essentially baked in a 1s increment. I think 563 s would be an odd time step, but there is nothing inherently wrong with it. If we do run into models that need to have an interval defined, it would not be difficult to include that capability at that point.

I have also added a unit test and an integration test for the variable dt grid component

h2integrate/converters/grid/grid.py
kbrunik: I wonder if in the more general implementation for the framework--like in `h2integrate_model.py` and the utilities files that house dt methods (not the already include converter/storage models)--we can allow dt to be less than 300 because I don't see a reason we couldn't theoretically allow users to put a signal of their choice in at a 1 second time resolution and use the other h2i capabilities such as the financial modeling with finer resolution performance data.

h2integrate/core/h2integrate_model.py
kbrunik: I think my prefer approach would be to add `_time_step_min` and `_time_step_max` to every model and it's never inherited from the Performance and Cost base classes. Even for models that can only accept dt=3600.

Then we can remove the `elif dt !=3600` because this will be caught in the above if statement.

kbrunik (approved):
I like where this is headed! Thanks for a functional draft and the quick testing of the added capabilities. I think we will need a lot of rigorous testing around the cost modeling and financial modeling to ensure we are handling everything correctly (especially with feedstocks).

jaredthomas68:
> I like where this is headed! Thanks for a functional draft and the quick testing of the added capabilities. I think we will need a lot of rigorous testing around the cost modeling and financial modeling to ensure we are handling everything correctly (especially with feedstocks).

Thank you @kbrunik, I don't think the cost and financial modeling will need much work because they use annual commodity amounts and capacity factor rather than the actual time-series. I looked into them but I don't think they need any changes.

h2integrate/core/h2integrate_model.py
johnjasa: I think we have a few different ways to essentially achieve the same checking capability here, thanks for the discussion. @kbrunik, I like your idea because of how explicit it is and how it will make it very clear what time steps each model accepts. I think this will make future updates to models as we expand out flexible dt values more trackable. I also think it will encourage (force) folks making new models to include `_time_step_bounds`, which will be useful to define at model creation time.

johnjasa: I'm done with those changes and welcome input from @jaredthomas68 and @elenya-grant! Are you okay with this? What would you change?

h2integrate/core/inputs/validation.py
```diff
@@ -140,20 +140,18 @@ def load_tech_yaml(finput):
...
-        )
-        raise ValueError(msg)
-    if int(plant_config["plant"]["simulation"]["dt"]) != 3600:
-        msg = (
-            "H2Integrate does not currently support simulations with a time step that is "
-            "less than or greater than 1-hour. Please ensure that "
-            "plant_config['plant']['simulation']['dt'] is set to 3600."
+            "plant_config['plant']['simulation']['n_timesteps'] times "
+            "plant_config['plant']['simulation']['dt'] equals 31536000 (s)."
         )
```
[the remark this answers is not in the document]
jaredthomas68: I took the int*int approach.

h2integrate/core/model_baseclasses.py
```diff
@@ -151,6 +151,9 @@ def setup(self):
             "cost_year", val=self.config.cost_year, desc="Dollar year for costs"
         )
 
+        # dt is seconds per timestep
+        self.dt = self.options["plant_config"]["plant"]["simulation"]["dt"]
```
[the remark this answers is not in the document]
jaredthomas68: I specified int.

PR #654, opened 2026-04-06
Removed remaining naming dependencies

### LINKED ISSUE #374: Remove any dependence on the name of the technologies in the `tech_config` files
# Remove any dependence on the name of the technologies in the `tech_config` files

There are some sections of `h2integrate_model.py` that have logic that depends on the name of the technology in the `tech_config`.
This is a holdover from the earlier days of H2I when we needed to conditionally create, connect, or manipulate technologies based on their type.
To provide maximum user flexibility and increase clarity of behavior in the models, we should remove any dependence on these names.

## Proposed solution

The smoothest solution is probably to take any existing logic that depends on names and translate that logic to instead depend on the `performance_model` or `cost_model` type.
For example, `electricity_producing_techs` currently depends on the name. It shouldn't! Instead, we should define the technology models themselves there instead of the name needed in the `tech_config`.
This logic also exists elsewhere, such as for the electrolyzer producing hydrogen and passing the total hydrogen produced or the time between replacement.

## Alternatives considered

We could force users to use specific names for different techs, but that is artificially limiting and becomes more challenging if users add multiple different instances of the same technology.

## Additional context

Related to #223 
Related to #127; might be worthwhile to do both at the same time

johnjasa:
# Use model class instead of tech name to skip no-cost technologies in finance connections

Previously, the finance connection logic in `connect_technologies` used string-based checks on the technology name (e.g., `"splitter" in tech_name`, `tech_name == "cable"`) to decide which technologies to skip when connecting CapEx/OpEx/VarOpEx to finance subgroups. This was fragile because renaming a technology could silently break the skip logic.

Now the code looks up each technology's `performance_model.model` from the tech config and checks it against a set of known no-cost model classes. This decouples the behavior from naming conventions.

```python
no_cost_models = {
    "GenericSplitterPerformanceModel",
    "GenericCombinerPerformanceModel",
    "GasStreamCombinerPerformanceModel",
    "CablePerformanceModel",
    "PipePerformanceModel",
}

for tech_name in tech_configs.keys():
    perf_model = tech_configs[tech_name].get("performance_model").get("model")
    if perf_model in no_cost_models:
        continue
```

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

## Section 2: General PR Checklist

- [ ] Documentation
- [x] `CHANGELOG.md`

## Section 3: Related Issues

Fully resolved #374 

### Section 4.1: New Files
N/A

  - `connect_technologies`: Replaced name-based `"splitter"/"combiner"/"cable"/"pipe"` string checks with a lookup against `no_cost_models` using the technology's configured model class.

N/A

N/A

N/A


h2integrate/core/h2integrate_model.py
```diff
@@ -1232,12 +1232,24 @@ def connect_technologies(self):
...
+                }
+
                 # Only connect technologies that are included in the finance stackup
                 for tech_name in tech_configs.keys():
-                    # For now, assume splitters and combiners do not add any costs
-                    if "splitter" in tech_name or "combiner" in tech_name:
-                        continue
-                    if tech_name == "cable" or tech_name == "pipe":
+                    # Skip technologies whose models doesn't add costs
+                    perf_model = tech_configs[tech_name].get("performance_model").get("model")
```
elenya-grant, line 1242: would using dictionary defaults prevent issues in case a technology (maybe not one that exists now) has a cost model but not a performance model?
```python
perf_model = tech_configs[tech_name].get("performance_model", {}).get("model", None)
```

h2integrate/core/h2integrate_model.py
```diff
@@ -1232,12 +1232,24 @@ def connect_technologies(self):
                         f"finance_subgroup_{group_id}.capacity_factor",
                     )
 
+                # Model classes that do not contribute costs to the finance stackup
+                no_cost_models = {
```
elenya-grant: could you extend this comment to mention that these models are like "internal models" so that's why they don't contribute to the costs for the finance modeling?

h2integrate/core/h2integrate_model.py
```diff
@@ -1232,12 +1232,24 @@ def connect_technologies(self):
...
+                    "GenericSplitterPerformanceModel",
+                    "GenericCombinerPerformanceModel",
+                    "GasStreamCombinerPerformanceModel",
```
elenya-grant: why did you put these in h2integrate_model.py instead of supported_models.py? I almost think that I'd be nice to have these in supported models so its a bit more "upfront" than buried within a class method but I'd love to hear your reasoning

docs/user_guide/expected_naming_convention.md
elenya-grant, line 1: Yay! Love that this file can be removed!

elenya-grant (commented):
Thanks for this PR John! I had a few small questions but thanks for making this PR to finally remove all naming-specific logic!

h2integrate/core/h2integrate_model.py
johnjasa: I love your suggestion! That makes total sense. I didn't have a great reason for putting them here; I think with your suggestion, we'll also have a good setup for other similar groupings. Will change it now, thank you!

h2integrate/core/h2integrate_model.py
johnjasa: Good consideration. This would avoid issues in the future for models that don't necessarily exist now, but it may also obfuscate the issue. Or at least I'd want to catch it sooner than this (connection time). So based on this comment, I've added `required: [performance_model]` to the `tech_schema.yaml` to require models there and do the checking.

elenya-grant: Okay - I think requiring the performance model makes sense for now (I was spooked when you said performance and cost model since splitters and combiners can be defined in the tech config but don't have an associated cost model). Thanks for explaining!

johnjasa: Yeah, I underthought it at first and thought we could require both but you're exactly right!

elenya-grant (approved):
Thanks for making these changes!

PR #655, opened 2026-04-07
Bugfix: Connecting resource to multiple techs

elenya-grant:
# Bugfix: Connecting resource to multiple techs

Added in a bugfix so a resource model can be connected to multiple technologies without throwing an error.

For example, if you have two different solar technologies (`solar_farm_A`, `solar_farm_B`) that are co-located, the below would've thrown an error saying theres missing resource models, even though there wasn't a missing resource model:
```yaml
resource_to_tech_connections: 
- [site.solar_resource, solar_farm_A, solar_resource_data]
- [site.solar_resource, solar_farm_B, solar_resource_data]
```

This was because the `resource_models` dictionary built in `connect_technologies()` had keys of the format `site_name-resource_model` (ex: `site-solar_resource`) and this was being checked against the formatting in `resource_to_tech_connections` which uses a `.`. This check was only done if there were more resource to tech connections than resources defined for the site, which was not tested to handle this case.

The bugfix was just changing the notation for the `resource_models` dictionary to use a period instead of a dash. 

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Documentation
- [x] `CHANGELOG.md`

### Section 4.1: New Files
- `path/to/file.extension`

  - `connect_technologies()`: changed the notation for the `resource_models` dictionary keys to use a period instead of a dash. 
  - `h2integrate/core/test/test_framework.py`: added test to make sure that no error is thrown if connecting one resource to multiple technologies. New test is named `test_no_resource_connection_error_resource_to_multiple_techs`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/core/test/test_framework.py
```diff
@@ -324,6 +324,30 @@ def test_resource_connection_error_missing_resource(temp_dir):
...
+        ["site.wind_resource", "wind_plant2", "wind_resource_data"],
+    ]
+    plant_config["resource_to_tech_connections"] = resource_to_tech_connections
+    input_config = {
+        "plant_config": plant_config,
+        "technology_config": tech_config,
+        "driver_config": driver_config,
+    }
+    H2IntegrateModel(input_config)
+    assert True
```
johnjasa: An important point here that I'm modifying but I want to draw attention to it -- to actually check this connection, we need to call `final_setup()` (usually done as part of `run()`) to actually test these connections.

h2integrate/core/h2integrate_model.py
```diff
@@ -1167,7 +1167,7 @@ def connect_technologies(self):
             resource_models = {}
             for site_grp, site_grp_inputs in self.plant_config["sites"].items():
                 for resource_key, resource_params in site_grp_inputs.get("resources", {}).items():
-                    resource_models[f"{site_grp}-{resource_key}"] = resource_params
+                    resource_models[f"{site_grp}.{resource_key}"] = resource_params
```
johnjasa, line 1170: This seems like a small but impactful change but I can't quite place it since there aren't other changes related to this. Is this the simple bugfix (naming from `-` to `.`) and that's all that's needed to enable this? If so, love it! Just making sure I understand.

johnjasa (approved):
Fun stuff! I like the enabling multiple techs _and_ your nice test. I modified the test so the `setup()` methods are called (this would trigger the actual connection testing). One small non-blocking question for my own understanding!

h2integrate/core/h2integrate_model.py
elenya-grant: Yes. `resource_source_connections` are the first element in the `resource_to_tech_connections` which are formatted as `site.wind_resource`. The checks are checking the keys in `resource_models` against the `resource_source_connections` - so if they have a different format (like `resource_models` using dashes) then that makes the comparison fail. Aka - the `missing_resource` variable would be populated with all the resources because of the difference in formatting.

h2integrate/core/test/test_framework.py
elenya-grant: thanks!

PR #664, opened 2026-04-09
Bugfix on storage so storage can't charge with unavailable input commodity

elenya-grant:
# Bugfix on storage so storage can't charge with unavailable input commodity

Storage performance models were not constrained to only charge with the available input commodity. If a controller were to tell the storage performance model to charge more than the available input commodity, then the storage would charge more than the available input commodity. This shouldn't be allowed to happen, storage performance models should only be able limited to only be able to charge with the commodity_in. This PR fixes that.

This PR also fixes some of the logic in the storage model baseclass to decide whether it's using feedback control. The previous logic was:
```python
if any(intended_dispatch_tech in name for name in self.tech_group_name):
```

and was updated to:
```python
if any(intended_dispatch_tech == name for name in self.tech_group_name):
```

Previously, if you had a technology named "battery" that uses a pyomo controller and another technology named "battery_1" that uses an open-loop controller. Then the logic in "battery_1" would have said it's using feedback control (when it isn't). Using `==` instead of `in` makes it so that the logic in `battery_1` will run as expected. 

Both of these changes have new tests!

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [ ] `CHANGELOG.md`

- `examples/34_plm_optimized_dispatch/`: tech config and plant config were updated to use an electricity feedstock instead of the grid. This change was made based on input from @vijay092 and the planned future work thats captured in Issue #749.
- `h2integrate/storage/battery/pysam_battery.py`: modified `PySAMBatteryPerformanceModel`
   - `simulate()`: updated to take commodity_available as an input and constrain charge based on the available commodity.
- `h2integrate/storage/storage_baseclass.py`: modified `StoragePerformanceBase`
   - `run_storage()`: updated calls to `simulate()` and added `commodity_available` as a key in the kwargs dictionary to pass to feedback controllers.
   - `simulate()`: updated to take commodity_available as an input and constrain charge based on the available commodity.
   - `setup()`: minor change in logic for deciding whether using feedback control or not
#### New tests
- `h2integrate/control/test/test_multistorage_pyomo_openloop_control.py::test_battery_pyomo_battery_openloop`: test that would've failed before this PR that tests the naming logic for deciding whether a technology has a feedback control strategy or not.
- `h2integrate/storage/test/test_storage_performance_model.py::test_generic_storage_charge_more_than_available`: test that would've failed before this PR. Tests when the commodity_set_point tells the storage to charge more than what's available.

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


jaredthomas68 (changes requested):
The storage base class changes to make all controllers take in the available commodity makes sense on the surface, but could ultimately limit the ability to charge storage from the grid because storage would never charge. I went and checked on the PLM optimized control and found it has not been correctly connected to the grid yet and so I would like to get that figured out first so we can appropriately address it in this PR. I made a draft PR with my ideas in #745 that I think should help clarify my concerns. 

That said, I really the ideas in this PR and expect them to come in once we have a clear path forward (probably this week I think). I just want to make sure we are all on the same page and don't negatively impact existing capabilities.

jaredthomas68 (approved):
I think I'm ok with where this is at now. Thank you for iterating. The run time is increased by about 7x to get convergence instead of doing a direct calculation, but it is still fast enough. I hope we can improve the run time in the future.

vijay092 (approved):
I looked at Example 34 and it looks good to me! Thanks, Elenya!

PR #631, opened 2026-03-26
Update Naming Convention in Open-Loop Converter Control Strategies

elenya-grant:
# Update Naming Convention in Open-Loop Converter Control Strategies

NOTE: The intention of this PR is a step in resolving Issue #521. Eventually, the unmet demand and unused commodity calculations done in storage models would instead be done with the use of these demand components (see the follow-on PR, #666). Please keep this in mind when reviewing the new outputs of these models (please see [this comment](https://github.com/NatLabRockies/H2Integrate/issues/521#issuecomment-4172661140) on how that is expected to change)

Update output naming convention in "open-loop converter control strategies" to align with the naming convention of the storage performance models.
- `f"{commodity}_unmet_demand"` -> `f"unmet_{commodity}_demand_out"`
- `f"{commodity}_unused_commodity"` -> `f"unused_{commodity}_out"`
- Removed output `f"total_{commodity}_unmet_demand"`
- Updated so that these inherit from the `PerformanceModelBaseClass`

**Update as of 04/06/2026:**

The above changes motivated a larger discussion around these "open-loop converter control strategies" and how they're actually used/what they were intended for. Some additional context on why these were originally put as "control strategies" is listed in Section 5. **But - my motivating philosophy to renaming these classes and moving the files is that these components are not used as controllers and are instead ways of representing some generic demand signal**. I think a common push-back on this idea is that demand should come from a converter component, but in some cases, we care more about the demand being met than what the demand is "turned into". For example (a common use-case), defining the energy demand of a data center is important so we can design a system to meet that demand and calculate the cost of energy to meet that demand. We are not trying to calculate the cost of the output from the data center, we only care about the inputs. These demand components are used to define a demand, subtract some initial amount from that demand, and output the remaining and excess demand. The unmet demand from the demand component can then be passed to some converter technology as the demand (example 23 is a good reflection of this). These models are primarily used to calculate demand metrics, not define a control strategy.

I propose the following renaming (already done): 
- `ConverterOpenLoopControlBase` -> `DemandComponentBase`
- `DemandOpenLoopConverterController` -> `GenericDemandComponent`
- `FlexibleDemandOpenLoopConverterController` -> `FlexibleDemandComponent`
- `openloop_controller_base.py` -> `demand_base.py`
- `demand_openloop_converter_controller.py` -> `generic_demand.py`
- `flexible_demand_openloop_controller.py` -> `flexible_demand.py`

Other changes:
[...]

PR #666, opened 2026-04-10
Remove system-level outputs from storage and replace with demand component (follow-on to 631)

elenya-grant:
# Remove system-level outputs from storage and replace with demand component (follow-on to 631)

There has been a blurry line distinguishing system-level calculations and calculations done in storage performance models. **This PR builds on PR #631** to resolve Issue #521. **The storage performance outputs that depend on demand have been removed and the usage of demand components in the examples is more common.** The outputs removed from storage performance models are:
- `unmet_{commodity}_demand_out`: demand that has not been met
- `unused_{commodity}_out`: production that exceeds the demand
- `storage_{commodity}_out`: charge/discharge profile

The outputs with different calculations in the storage performance models are:
- `{commodity}_out`: this used to be the _**combined**_ commodity out from storage and whatever was input to storage, minus any production that exceeds the demand. This is now what `storage_{commodity}_out` used to be
- `capacity_factor`: used to be the based on the combined commodity out, now it based on the sum of the charge/discharge profile

demand_calcs
demand_commodity_out

This PR helps to better distinguish the differences between individual storage performance models and "system-level" performance calculations. This PR also enables the ability for the storage demand to be distinct from the overall demand. For example, if we wanted to use a battery to try to just keep an electrolyzer system _on_ (at 10% power), then the demand profile to the battery would be 10% of the electrolyzer capacity, whereas the demand for the demand component would be equal to the electrolyzer capacity. 

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Get PR #631 merged in
- [ ] Update `storage_baseclass.py` so that demand is only an input if using feedback control (not doing just yet)
- [x] make figures to show what's happening in demand components for different use-cases 
   - these figures have been added to `docs/demand/demand_components.md`
- [x] add test to test_all_examples for example 13
- [x] update or add doc page for example 13 in controller docs

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [ ] `CHANGELOG.md`

## Section 3: Related Issues

This would resolve Issue #521

### Section 4.1: New Files
- `examples/13_dispatch_for_electrolyzer/`: new example that highlights setting the storage demand as a different value than the demand component.
- `resource_files/wind/30.6617_-101.7096_2013_wtk_v2_60min_utc_tz.csv`: wind resource file for example 23
- `docs/demand/demand_demo.md`: new doc page for Example 13 that highlights use of demand components.
- `docs/technology_models/storage_models_index.md`: new doc page describing the inputs and output calculations of the storage performance models.

  - `create_finance_model()`: Updated logic so that any model without a cost model does not have to be included in the technologies list for a finance group. Was previously hard-coded so that only combiners didn't have to be included in the technologies list.
- `h2integrate/storage/storage_baseclass.py`: modified `StoragePerformanceBase`
   - `setup()`: removed outputs related to demand
   - `run_storage()`: removed calculations of outputs related to demand. Updated calculation of capacity factor.
- `examples/test/test_all_examples.py::test_ng_demand_example`: new test for Example 23
- `examples/test/test_all_examples.py::::test_electrolyzer_demand`: new test for Example 13

#### Examples that were updated

- Examples that had added combiners to combine storage outputs with the input to storage: 01, 02, 03/co2_hydrogenation_doc, 09/doc, 09/oae, 30
- Examples that had added combiners and load demand components: 12, 16, 24, 30
- Examples that had storage models as commodity streams (required additional combiner(s) and a new load demand component): 14, 18, 19, 29

- `h2integrate/core/test/test_framework.py::test_system_order`
- `h2integrate/core/test/test_framework.py::test_technology_connections`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


bayc:
@elenya-grant Trying to look through this, and wondering if you can put together a diagram or figure showing the inputs/outputs of the current system and proposed solution? Would help a lot to see what the differences are.

docs/control/controller_demonstrations.md
kbrunik, line 1: I was looking at the docs for this and I think it could make sense to update example 14 and the outputs in this demonstration to show how the demand to the `h2_storage` and the `h2_load_demand` can be different and how that impacts the analysis. I would also recommend updating the `tech_config` executable block because it doesn't include the `h2_load_demand` and I think that would be helpful to showcase.

kbrunik (commented):
I think this all makes sense for separating out the demand feature into the demand component. I do think a figure highlighting how the demand component works with the storage in this new configuration would be helpful.

docs/control/controller_demonstrations.md
elenya-grant: That's a good idea - I will work on that!

elenya-grant: I think that example 1 or 2 may be a better candidate. Those ones have a battery and we could set the battery demand profile to 10% of the electrolyzer rated capacity and then add/set an electricity component demand profile equal to the electrolyzer rated capacity. There is nothing that hydrogen is going to in example 14 which makes it hard to come up with a storage why the storage demand profile would be different than the demand component demand profile.

bayc:
> @elenya-grant Trying to look through this, and wondering if you can put together a diagram or figure showing the inputs/outputs of the current system and proposed solution? Would help a lot to see what the differences are.

Thanks for adding the plots! That definitely helps me understand what the code is doing, which is moving the difference calculations out of the storage model. I think these outputs were originally developed with converters in mind as well and trying to apply the same set of inputs/outputs across technologies to be able to easily know how much of a commodity was consumed and how much, if any, demand is left unmet after sequentially working through each component.

My understanding now is that the difference calculations are being moved to demand components, and agree with @kbrunik that it would be helpful to see this built out in an example, so supportive of that! This would help me understand any other impacts to the dispatch/control flow and demand passing through converters as well.

docs/control/controller_demonstrations.md
elenya-grant: I made a new example (example 13) and used that to make a demand_demo.md page in the `demand` docs folder.

examples/13_dispatch_for_electrolyzer/plant_config.yaml
```diff
@@ -0,0 +1,76 @@
...
+  # combine the generation and input it to the battery
+  - [distributed_wind_plant, gen_combiner, electricity, cable]
+  - [distributed_pv, gen_combiner, electricity, cable]
+  - [gen_combiner, battery, electricity, cable]
+  # combine the battery output with the wind and solar generation
+  - [battery, elec_combiner, electricity, cable]
+  - [gen_combiner, elec_combiner, electricity, cable]
+  # subtract the electricity generated from the electricity demand
+  - [elec_combiner, elec_load_demand, electricity, cable]
+  # connect the electricied supplied to the demand to the electrolyzer
```
genevievestarke:
```suggestion
  # connect the electricity supplied to the demand to the electrolyzer
```

examples/13_dispatch_for_electrolyzer/run_dispatch_for_electrolyzer.py
```diff
@@ -0,0 +1,40 @@
...
+    "driver_config": driver_config,
+}
+
+h2i = H2IntegrateModel(input_config)
+
+h2i.setup()
+
+electrolyzer_capacity_MW = 60
+h2i.prob.set_val("battery.electricity_demand", 0.1 * electrolyzer_capacity_MW, units="MW")
+h2i.prob.set_val("elec_load_demand.electricity_demand", electrolyzer_capacity_MW, units="MW")
```
genevievestarke, line 31: Ooooh, nice, I really like this example!

h2integrate/demand/flexible_demand.py
```diff
@@ -49,6 +49,8 @@ class FlexibleDemandComponent(DemandComponentBase):
     and used to compute unmet demand, unused commodity, and delivered output.
     """
 
+    _time_step_bounds = (3600, 3600)  # (min, max) time step lengths compatible with this model
```
genevievestarke, line 52: Is this something that can be inherited from demand_base.py?

h2integrate/storage/storage_baseclass.py
```diff
@@ -316,16 +279,11 @@ def run_storage(
...
-        # System-level outputs calculated in storage
-        outputs[f"unmet_{self.commodity}_demand_out"] = unmet_demand
-        outputs[f"unused_{self.commodity}_out"] = unused_commodity
-        outputs[f"{self.commodity}_out"] = total_commodity_out
+        outputs[f"{self.commodity}_out"] = storage_commodity_out
 
         # Performance model outputs
         outputs[f"rated_{self.commodity}_production"] = discharge_rate
-        outputs[f"total_{self.commodity}_produced"] = np.sum(total_commodity_out)
+        outputs[f"total_{self.commodity}_produced"] = np.sum(storage_commodity_out)
```
genevievestarke, line 297: I think this should be based just on the sum of storage discharge (otherwise you'll get a very small number).

h2integrate/control/test/test_openloop_controllers.py
```diff
@@ -16,6 +16,16 @@
 )
 
 
+def calculate_combined_outputs(storage_charge_discharge, commodity_in, commodity_demand):
```
genevievestarke, line 19: I think this makes sense to keep the tests the same, but maybe as a future PR we can think about whether we need to calculate all of these for the controller tests. (Or just test these for the controllers by themselves)

genevievestarke (commented):
Overall, I think this looks really good! Great progress on this, @elenya-grant ! I have a few comments, and I'd like to look at the final docs pages when they're ready, as well :)

h2integrate/demand/flexible_demand.py
elenya-grant: yeah - but I just put in each one separately since one could be updated and not the other. I think it'll be a lot easier to make the generic demand component flexible to different time-steps, but it may be harder for the flexible demand one.

h2integrate/storage/storage_baseclass.py
elenya-grant: for sure! I hear you - the benefit of this method is that, suppose we have something a system that has technology connections that look like this:
```yaml
technology_interconnections:
  # combine the generation and input it to the battery
  - [wind, gen_combiner, electricity, cable]
  - [pv, gen_combiner, electricity, cable]
  - [gen_combiner, battery, electricity, cable]
  # combine the battery output with the wind and solar generation
  - [battery, elec_combiner, electricity, cable]
  - [gen_combiner, elec_combiner, electricity, cable]
```

Then, it's nice because it follows the same logic as the converters and the following two things are equal:

```python
tot_electricity_out = model.prob.get_val("battery.total_electricity_produced") + model.prob.get_val("wind.total_electricity_produced") + model.prob.get_val("wind.total_electricity_produced")

assert tot_electricity_out == model.prob.get_val("elec_combiner.electricity_out").sum()
```

I'd be interested to hear what other folks think.

h2integrate/control/test/test_openloop_controllers.py
elenya-grant: I agree - it was only necessary because most of the subtests were on these variables that are now calculated in demand components - but I agree.

h2integrate/storage/storage_baseclass.py
kbrunik: I think this gets to Jared's comment in #521 to have "outputs of commodity flows mutually exclusive and collectively exhaustive" and it's nice to have the storage not output values from other components.

docs/demand/demand_demo.md
jaredthomas68, line 1: This is a very nice demo, well done putting this together!

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,233 @@
...
+- `23_solar_wind_ng_demand`: compares usage of the `GenericDemandComponent` and `FlexibleDemandComponent`
+- `24_solar_battery_grid`: sells excess electricity to grid
+
+This demonstration will focus on the `13_dispatch_for_electrolyzer` example
+
+## Electrolyzer load demand
+
+The following example is an expanded form of `examples/13_dispatch_for_electrolyzer`.
+
+The technology interconnections:
```
jaredthomas68, line 26: I think the user would benefit from an xdsm diagram here.

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,233 @@
...
+ax.fill_between(x, generation[start_hour:end_hour], generation_with_battery[start_hour:end_hour], where=where_charge, color="tab:cyan", alpha=0.5, zorder=0, label="battery charging")
+ax.fill_between(x, generation[start_hour:end_hour], generation_with_battery[start_hour:end_hour], where=where_discharge, color="tab:orange", alpha=0.5, zorder=0, label="battery discharging")
+
+
+ax.spines[['right', 'top']].set_visible(False)
+ax.set_xlim([start_hour, end_hour])
+ax.set_ylim([0, 70])
+ax.legend(bbox_to_anchor=(1.0, 0.5), loc="center left", borderaxespad=0, framealpha=0.0)
+ax.set_ylabel("Electricity (MW)")
+ax.set_xlabel("Time (hours)")
```
jaredthomas68, line 210: You did a very nice job with the figures in this page. Clean and clear. Non-blocking suggestion here, but I would like to see the storage SOC and charge and discharge plots directly. I think it would make the storage operation a little more clear.

examples/03_methanol/co2_hydrogenation_doc/plant_config_co2h.yaml
```diff
@@ -15,9 +15,18 @@ technology_interconnections:
...
+  # connect the h2 supply to the methanol system
+  - [h2_combiner, methanol, hydrogen, pipe]
+  # connect the doc to co2 storage
   - [doc, co2_storage, co2, pipe]
-  - [co2_storage, methanol, co2, pipe]
+  # combine the co2 from the doc and the co2_storage
+  - [doc, co2_combiner, co2, pipe]
+  - [co2_storage, co2_combiner, co2, pipe]
+  # connect the co2 supply to the methanol system
+  - [co2_combiner, methanol, co2, pipe]
```
jaredthomas68, line 29: I am a little concerned with the amount of connections needed here, but I don't think it can be helped if we are going to move demand outside of storage, which I do see value in.

h2integrate/control/test/test_openloop_controllers.py
jaredthomas68: `calculate_combined_outputs` may be better placed in `tools.py` or similar instead of copying it around to various scripts. I also hesitate to base so many tests on a function, but it would be rather clunky to store values for all of these.

jaredthomas68: I agree with @genevievestarke that we should remove tests on values not calculated by these components any more, but in a future PR.

docs/control/controller_demonstrations.md
```diff
@@ -17,28 +17,39 @@ kernelspec:
...
+
 Using the primary configuration, we can create, run, and postprocess an H2Integrate model.
 
 ```{code-cell} ipython3
 # Create an H2Integrate model
-model = H2IntegrateModel(Path("../../examples/14_wind_hydrogen_dispatch/inputs/h2i_wind_to_h2_storage.yaml"))
+model = H2IntegrateModel(EXAMPLE_DIR/"14_wind_hydrogen_dispatch"/"inputs"/"h2i_wind_to_h2_storage.yaml")
 
 # Run the model
 model.run()
```
kbrunik, line 55: Can we remove the `model.post_process()` call?

docs/demand/figures/flex_demand_fig.png
kbrunik, line 1: Looking at this figure, the red is really hard to see even for a color seeing person. I realize this isn't a new addition in this PR but do we have to code to maybe change to color and that wouldn't be too hard? If the answer is no then don't do it.

docs/demand/demand_components.md
```diff
@@ -88,3 +88,23 @@ model_inputs:
     ramp_up_rate_fraction: 0.5
     min_utilization: 0
 ```
+
+(demand-component-inputs-and-outputs)=
```
kbrunik: I would move this above the section for the generic demand component

docs/demand/figures/demand_calcs.png
kbrunik, line 1: are the added figures color-blind friendly?

docs/demand/demand_components.md
kbrunik, line 1: I think these figures really help illustrate what the calculations are doing :) thank you!

examples/13_dispatch_for_electrolyzer/tech_config.yaml
```diff
@@ -0,0 +1,144 @@
...
+        max_charge_rate: 12000  # equal to 2x electrolyzer turndown ratio
+        max_capacity: 48000  # 4-hour battery
+      # performance_parameters:
+        init_soc_fraction: 0.0
+        charge_equals_discharge: true
+        charge_efficiency: 1.0
+        discharge_efficiency: 1.0
+        min_soc_fraction: 0.0
+        max_soc_fraction: 1.0
+        demand_profile: 6000  # equal to electrolyzer turndown ratio
```
kbrunik: Can we have the battery include in MW ratings? I think it would be clearer for the text included in the demand demonstration that says 6MW is the turndown rating

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,213 @@
...
+```{literalinclude} ../../examples/13_dispatch_for_electrolyzer/tech_config.yaml
+:language: yaml
+:lineno-start: 121
+:linenos: true
+:lines: 121-123,126-127,130,131,135
+```
+
+We want to dispatch the battery to *keep the electrolyzer on*. We set the demand profile for the battery as the minimum power required to keep the electrolyzer on, 6 MW.
+
+```{note}
```
kbrunik, line 50: Helpful note!

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,213 @@
...
+# Setup the model
+h2i.setup()
+```
+
+If we wanted to change the demand profiles for the battery (`battery`) or the demand component (`elec_load_demand`) to be different than the demand profiles specified in the technology config, we can do that using `set_val`:
+
+```{code-cell} ipython3
+electrolyzer_capacity_MW = 60
+
+# Set the battery demand equal to the minimum electricity needed to keep the electrolyzer on
```
kbrunik: Wait, didn't you already set this in the `tech_config`?

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,213 @@
...
+ax.plot(x, h2i.prob.get_val("elec_load_demand.electricity_out", units="MW")[start_hour:end_hour], color="tab:purple", lw=2.0, ls='solid', zorder=4, label="commodity_out")
+ax.spines[['right', 'top']].set_visible(False)
+ax.set_xlim([start_hour, end_hour])
+ax.set_ylabel("GenericDemandComponent \nElectricity (MW)")
+ax.set_xlabel("Time (hours)")
+ax.legend(bbox_to_anchor=(0.0, 0.95), loc="lower left", ncols=3, borderaxespad=0.0, framealpha=0.0)
+```
+
+
+Plot the battery performance:
```
kbrunik, line 182: Sick plot!!!!

docs/demand/demand_demo.md
```diff
@@ -0,0 +1,213 @@
...
+  display_name: native-ard-h2i
+  language: python
+  name: python3
+---
+
+# Demand Demonstration
+Different usage of demand components is shown in the following examples:
+- `13_dispatch_for_electrolyzer`: sets the battery demand differently than the demand profile of the demand component
+- `23_solar_wind_ng_demand`: compares usage of the `GenericDemandComponent` and `FlexibleDemandComponent`
+- `24_solar_battery_grid`: sells excess electricity to grid
```
kbrunik: Couple of small edits

```
Different usage of demand components are shown in the following examples:
- `13_dispatch_for_electrolyzer`: showcases how to set an overall demand for the system and a separate demand for the battery system.
- `23_solar_wind_ng_demand`: compares usage of the `GenericDemandComponent` and `FlexibleDemandComponent`.
- `24_solar_battery_grid`: highlights how to sell the electricity in excess of the demand to the grid.
```

resource_files/wind/30.6617_-101.7096_2013_wtk_v2_60min_utc_tz.csv
kbrunik, line 1: Was there not an existing wind resource file that worked for this example?

h2integrate/control/test/test_openloop_controllers.py
kbrunik: It could have been my fault in the merge too, sorry if this isn't used any more

kbrunik (approved):
Thanks @elenya-grant! I think this PR helps continue to clarify/separate "storage" models from non-storage/more generic things like the controls logic and the demand. I like that none of the actual calculations/test values changed in this PR just mostly reorganizing and adding some sweeeeet docs. I have a few non-blocking comments but I'd love if you'd take a look and respond to them before merging.

h2integrate/storage/simple_storage_auto_sizing.py
```diff
@@ -79,6 +79,22 @@ def __attrs_post_init__(self):
         if self.commodity_amount_units is None:
             self.commodity_amount_units = f"({self.commodity_rate_units})*h"
 
+        # Check that the demand profile is zero if set_demand_as_avg_commodity_in is True
+        if isinstance(self.demand_profile, list | np.ndarray):
+            user_input_dmd = True if sum(self.demand_profile) > 0 else False
+        else:
+            user_input_dmd = True if self.demand_profile > 0 else False
+
```
jaredthomas68: np.sum works on scalars and array-like. Also, the comment was confusing to me. Also, what if 
user defines a negative profile?

```suggestion
        # Check for user-define demand profile
        user_input_dmd = True if np.sum(self.demand_profile) > 0 else False
```

h2integrate/storage/storage_baseclass.py
jaredthomas68: I see your reasoning @elenya-grant and I think it works for your use case. My concern is that the capacity factor is based on the total commodity production, but a storage systems capacity factor is calculated based only on discharge and divided by possible discharged continuously. If we make sure the capacity factor calculation is being handled correctly, then I think I'm ok with the positive and negative included in the sum. We will need to be very clear about its meaning in docs and doc strings. I am still concerned about the misleading name "produced" when the sum includes commodity used.

jaredthomas68 (changes requested):
This is looking pretty good. I think it makes operation more transparent to the users and will allow more flexibility ultimately. I do have a few comments/changes. My biggest concern is that the storage capacity factor be calculated correctly.

docs/demand/demand_demo.md
elenya-grant: yes - I may have pushed up changes since your comment (the line numbers are confusing me). Before the part where I call set_val() now - I have this line:


> If we wanted to change the demand profiles for the battery (`battery`) or the demand component (`elec_load_demand`) to be different than the demand profiles specified in the technology config, we can do that using `set_val`:

resource_files/wind/30.6617_-101.7096_2013_wtk_v2_60min_utc_tz.csv
elenya-grant: unfortunately no.. I don't like that I had to add it either (it was for example 23, not the new example I added). I can change the location to a location that we already have weather data files for and remove this - I was just trying to not change example 23 too much.

h2integrate/demand/flexible_demand.py
genevievestarke: That's a good point, thank you!

docs/demand/demand_demo.md
elenya-grant: heard - how do I make that?

h2integrate/storage/simple_storage_auto_sizing.py
elenya-grant: Thanks on the suggestion - I will push up that change shortly (also with an updated comment). 

The question of a negative demand profile is interesting - that is something that we don't have a way of handling for any converter model that can take demand as an input or in the demand components. I think that we should discuss how we'd want to handle that and make a separate PR with whatever route seems best.

examples/13_dispatch_for_electrolyzer/tech_config.yaml
elenya-grant: I gave it a try and for some reason I'm getting a different LCOH. I think this could be due to some small floating point error difference that propagates. I will work on debugging it!

docs/demand/demand_demo.md
elenya-grant: done!

docs/demand/demand_components.md
elenya-grant: done!

docs/demand/figures/flex_demand_fig.png
elenya-grant: I don't remember where the code is that I had to make this figure initially - so - I can try to re-make it but it'd take 30min+

docs/control/controller_demonstrations.md
elenya-grant: done!

docs/demand/figures/demand_calcs.png
elenya-grant: I'm not sure...could you suggest some colors that you'd like instead?

docs/demand/demand_demo.md
elenya-grant: done!

h2integrate/control/test/test_openloop_controllers.py
elenya-grant: I think that maybe we revisit this in a future PR. the options are:
- update the tests to test different outputs (don't test the unmet demand, excess commodity, etc)
- keep the tests as-is and move the `calculate_combined_outputs` to `conftest.py`?
- update the tests to use a demand component for the calculations instead of `calculate_combined_outptus`

h2integrate/storage/storage_baseclass.py
elenya-grant: Okay - so I think the main question is whether aggregated outputs from the storage models (`annual_commodity_produced`, `total_commodity_produced`, and `capacity_factor`) should be based on either 1) charge and discharge profile or 2) just the discharge profile.

My main concern is with the capacity factor, since that's passed to combiners and is a metric used in finance calculations. I don't have strong feelings about `annual_commodity_produced` and `total_commodity_produced`, since those are more useful as "summary outputs" and not used in downstream models.

If storage model outputs are based on the charge and discharge profile - then this is similar to how converter outputs are calculated wrt the `commodity_out` profile (nice). But - if the battery charges more than discharges, then we would have negative values (could cause problems). If the outputs are negative though - then I think `produced` makes sense in the name.

If storage model outputs are based only on the discharge profile - then we'd always have positive outputs, but we couldn't sum these together with the converter technologies like I mentioned [above](https://github.com/NatLabRockies/H2Integrate/pull/666#discussion_r3076270529). Aka - it'd require special treatment of converter technologies.

Regardless of what we decide, I will update the doc page for the storage performance models to detail how these calculations are done.

examples/13_dispatch_for_electrolyzer/tech_config.yaml
elenya-grant: Using MW: 196866.67725101 MWh/year to electrolyzer (elec cf of 38.6**4**543851 (year0), load cf of 37.45560830, lcoh of 10.20311181)

Using kW: 196866.67725101 MWh/year to electrolyzer (elec cf of 38.6**8**285970 (year0), load cf of 37.45560830, lcoh of 10.22009).

docs/demand/demand_demo.md
johnjasa: image

We could use the built-in XDSM-maker in H2I, callable via `create_xdsm` on the `H2IntegrateModel` object.
Or, we could use [pyXDSM](https://github.com/mdolab/pyXDSM) manually like I did for this above.
@jaredthomas68, is this in line with what you're thinking? What would you change?

h2integrate/storage/storage_baseclass.py
elenya-grant: Larger question that I think would help us figure out what direction to go in. How would you calculate the capacity factor of a wind+battery system?

```python
wind_aep = sum(wind_generation)
wind_cf = wind_aep/(wind_capacity*8760)

# Like below?
hybrid_cf = sum(wind_generation)/(8760*(wind_capacity + battery_discharge_rate))
```

docs/demand/demand_demo.md
jaredthomas68: Yes, that is exactly what I was thinking.  It would be nice to use the built-in method as part of the example, but all I was asking for is including the figure to visualize the connections.

elenya-grant: thanks! I added it in

h2integrate/storage/storage_baseclass.py
jaredthomas68: I think it is important to keep standard procedures. I also think we can just take the positive part of the commodity out for all techs and divide by their nameplate capacity and then the capacity factor calculation will be the same for all technologies.

```
     cf = sum(positive(commodity out)/(rate unit rating * hours in year)
```

From the [NLR ATB](https://atb.nrel.gov/electricity/2024b/utility-scale_battery_storage):

"Capacity Factor

The cost and performance of the battery systems are based on an assumption of approximately one cycle per day. Therefore, a 4-hour device has an expected capacity factor of 16.7% (4/24 = 0.167), and a 2-hour device has an expected capacity factor of 8.3% (2/24 = 0.083). Degradation is a function of the usage rate of the model, and systems might need to be replaced at some point during the analysis period. We use the capacity factor for a 4-hour device as the default value for ATB because 4-hour durations are anticipated to be more typical in the utility-scale market.

The hybrid capacity factor is an important question. I do want to avoid having our own internal definition of capacity factor, but I also see the utility of @elenya-grant's proposal for using positive and negative in the CF calc

elenya-grant: I added `standard_capacity_factor` as an output that uses the equation @jaredthomas68 proposed. This is documented in the new storage doc page `docs/technology_models/storage_models_index.md`. Please let me know your thoughts.

h2integrate/control/test/test_openloop_controllers.py
elenya-grant: Decided we're going to do the first option in a future PR. made Issue #676

docs/demand/demand_demo.md
elenya-grant: commented out the python code in that section but left the note so folks understand.

docs/demand/figures/demand_calcs.png
elenya-grant: Noted in Issue #667

docs/demand/demand_demo.md
```diff
@@ -101,16 +104,16 @@ h2i = H2IntegrateModel(input_config)
...
 electrolyzer_capacity_MW = 60
 
 # Set the battery demand equal to the minimum electricity needed to keep the electrolyzer on
-h2i.prob.set_val("battery.electricity_demand", 0.1 * electrolyzer_capacity_MW, units="MW")
+# h2i.prob.set_val("battery.electricity_demand", 0.1 * electrolyzer_capacity_MW, units="MW")
 
 # Set the demand of the demand component equal to the rated electrical capacity of the electrolyzer
-h2i.prob.set_val("elec_load_demand.electricity_demand", electrolyzer_capacity_MW, units="MW")
+# h2i.prob.set_val("elec_load_demand.electricity_demand", electrolyzer_capacity_MW, units="MW")
 ```
```
jaredthomas68, line 117: Since you want to show the code but not run the code, just render the code in the standard non-running code block.

```python
electrolyzer_capacity_MW = 60

# Set the battery demand equal to the minimum electricity needed to keep the electrolyzer on
h2i.prob.set_val("battery.electricity_demand", 0.1 * electrolyzer_capacity_MW, units="MW")

# Set the demand of the demand component equal to the rated electrical capacity of the electrolyzer
h2i.prob.set_val("elec_load_demand.electricity_demand", electrolyzer_capacity_MW, units="MW")
```
Which, non-rendered, looks like this:

Image

docs/demand/demand_demo.md
```diff
@@ -207,6 +210,30 @@ ax.set_ylabel("Electricity (MW)")
 ax.set_xlabel("Time (hours)")
 ```
 
+Plot the battery SOC and charge/discharge profile
```
jaredthomas68, line 213: Thanks for adding this. I would suggest either removing the demand line or adding a legend.

docs/technology_models/storage_models_index.md
```diff
@@ -0,0 +1,60 @@
...
+The following sections detail the inputs and outputs of the storage performance models.
+
+```{note}
+The inputs and outputs of storage performance models are generalized here for any commodity. If input and output names include the word `commodity`, the actual variable name would be the commodity defined for that storage model. For example, the `PySAMBatteryPerformanceModel` can only be used for the commodity  `electricity`. Therefore, the `commodity_in` input to the `PySAMBatteryPerformanceModel` is actually named `electricity_in`.
+```
+
+### Inputs
+- `commodity_in`: commodity available to use for charging storage
+
+If using a **feedback control strategy** (this means that `tech_to_dispatch_connections` are defined in the plant configuration file for the storage technology), the control-related inputs for control to the storage performance:
```
jaredthomas68:
```suggestion
If using a **feedback control strategy** (this means that the controller received the actual storage state periodically), the control-related inputs for control to the storage performance include:
```

docs/technology_models/storage_models_index.md
```diff
@@ -0,0 +1,60 @@
...
+
+The aggregated or summarized performance outputs are (single values):
+- `storage_duration`: the storage capacity divided by the storage discharge rate. Units are time-based, such as `h` (hours).
+- `rated_commodity_production`: the storage discharge rate in units of `commodity_rate_units`.
+- `total_commodity_produced`: the summation of `commodity_out` over the simulation in units of `commodity_amount_units`. This value may be negative if the storage charges more than discharges.
+
+The results that are output per-year of the `plant_life` are:
+- `annual_commodity_produced`: each value is the `total_commodity_produced` scaled to a 1-year (8760 hours) simulation. This value may be negative if the storage charges more than discharges.
+- `capacity_factor`: the storage capacity factor calculated as capacity factors are calculated in converter technologies, a ratio of the sum of *`total_commodity_produced`* to the discharge rate (or `rated_commodity_production`). This value may be negative if the storage charges more than discharges.
+    $$
```
jaredthomas68: just single "$". The extras are rendering around your equations.

docs/technology_models/storage_models_index.md
```diff
@@ -0,0 +1,60 @@
...
+    CF = \frac{\sum_{t=0}^{n_{timesteps}}(\text{commodity_out})}{\text{discharge_rate}*n_{\text{timesteps}}}
```
jaredthomas68: Rather than time steps, it should be time.

docs/technology_models/storage_models_index.md
```diff
@@ -0,0 +1,60 @@
...
+    $$
+- `standard_capacity_factor`: the storage capacity factor as defined by the [NLR ATB](https://atb.nrel.gov/electricity/2024b/utility-scale_battery_storage). The ratio of the total *commodity discharged* to the discharge rate (or `rated_commodity_production`). This value will always be greater than or equal to zero.
+    $$
+    CF_{\text{standard}} = \frac{\sum_{t=0}^{n_{timesteps}}(\text{storage_commodity_discharge})}{\text{discharge_rate}*n_{\text{timesteps}}}
```
jaredthomas68: over time*rating, not timesteps*rating

h2integrate/storage/test/test_storage_performance_model.py
```diff
@@ -376,6 +388,26 @@ def test_generic_storage_with_simple_control_charge_rate_lessthan_demand(plant_c
...
+
+    with subtests.test("Expected capacity factor"):
+        assert (
+            pytest.approx(0.0, rel=1e-6)
+            == prob.get_val("storage.capacity_factor", units="percent")[0]
+        )
+
+    with subtests.test("Expected standard capacity factor"):
+        assert (
+            pytest.approx(15.416666, rel=1e-6)
```
jaredthomas68, line 407: I like these tests, thanks for adding them!

h2integrate/storage/storage_baseclass.py
```diff
@@ -293,10 +301,16 @@ def run_storage(
...
             outputs["capacity_factor"] = 0.0
+            outputs["standard_capacity_factor"] = 0.0
         else:
             outputs["capacity_factor"] = outputs[f"total_{self.commodity}_produced"] / (
                 outputs[f"rated_{self.commodity}_production"] * self.n_timesteps
             )
+            # standard_capacity_factor is the ratio of commodity discharged to the discharge rate
+            total_commodity_discharged = outputs[f"storage_{self.commodity}_discharge"].sum()
+            outputs["standard_capacity_factor"] = total_commodity_discharged / (
+                outputs[f"rated_{self.commodity}_production"] * self.n_timesteps
```
jaredthomas68, line 312: This will need to be made time step agnostic, but we can leave as is for now.

jaredthomas68 (changes requested):
This is looking really good. I only have a few simple changes requested, mostly typos in docs.

docs/demand/demand_demo.md
kbrunik: Because of the executables within the doc page it doesn't work to use just a python block or a bash block. It causes the doc build to fail. I tried to update it but ended up leaving it as is.

jaredthomas68: Ok, thanks for trying.

docs/technology_models/storage_models_index.md
```diff
@@ -39,13 +39,13 @@ The aggregated or summarized performance outputs are (single values):
 The results that are output per-year of the `plant_life` are:
 - `annual_commodity_produced`: each value is the `total_commodity_produced` scaled to a 1-year (8760 hours) simulation. This value may be negative if the storage charges more than discharges.
 - `capacity_factor`: the storage capacity factor calculated as capacity factors are calculated in converter technologies, a ratio of the sum of *`total_commodity_produced`* to the discharge rate (or `rated_commodity_production`). This value may be negative if the storage charges more than discharges.
-    $$
-    CF = \frac{\sum_{t=0}^{n_{timesteps}}(\text{commodity_out})}{\text{discharge_rate}*n_{\text{timesteps}}}
-    $$
+    $
+    CF = \frac{\sum_{t=0}^{n_{time}}(\text{commodity_out})}{\text{discharge_rate}*n_{\text{time}}}
+    $
```
jaredthomas68, line 44: I don't think `n_time` is correct. I'm sorry I was not more specific.

```suggestion
    $
    CF=\frac{\sum_{t=0}^{n_{\text{time\_steps}}}(\text{commodity\_out}*dt)}{\text{discharge\_rating}*\text{total\_time}}
    $
```

docs/technology_models/storage_models_index.md
```diff
@@ -39,13 +39,13 @@ The aggregated or summarized performance outputs are (single values):
...
 - `standard_capacity_factor`: the storage capacity factor as defined by the [NLR ATB](https://atb.nrel.gov/electricity/2024b/utility-scale_battery_storage). The ratio of the total *commodity discharged* to the discharge rate (or `rated_commodity_production`). This value will always be greater than or equal to zero.
-    $$
-    CF_{\text{standard}} = \frac{\sum_{t=0}^{n_{timesteps}}(\text{storage_commodity_discharge})}{\text{discharge_rate}*n_{\text{timesteps}}}
-    $$
+    $
+    CF_{\text{standard}} = \frac{\sum_{t=0}^{n_{time}}(\text{storage_commodity_discharge})}{\text{discharge_rate}*n_{\text{time}}}
+    $
```
jaredthomas68, line 48:
```suggestion
$
CF=\frac{\sum_{t=0}^{n_{\text{time\_steps}}}(\text{storage\_commodity\_discharge}*dt)}{\text{discharge\_rating}*\text{total\_time}}
$
```

examples/test/test_all_examples.py
```diff
@@ -790,28 +790,28 @@ def test_electrolyzer_demand(subtests, temp_copy_of_example):
     lcoh = h2i.prob.get_val("finance_subgroup_hydrogen.LCOH", units="USD/kg")[0]
 
     with subtests.test("LCOE of electricity generated"):
-        assert pytest.approx(138.905228, rel=1e-6) == lcoe_gen
+        assert pytest.approx(217.53810477, rel=1e-6) == lcoe_gen
```
jaredthomas68, line 793: Can you explain why these test values changed so much?

jaredthomas68 (changes requested):
Pretty close. I still have a few blocking comments.

docs/technology_models/storage_models_index.md
elenya-grant: I think n_timesteps is correct. That's the variable we use in H2I so I think it makes sense as it was before.

examples/test/test_all_examples.py
elenya-grant: resource data changed, was using data with UTC timezone and now using data with local timezone. Tests pass if timezone is 0, but those resource files (for UTC timezone) aren't available in the resource_files directory

docs/technology_models/storage_models_index.md
elenya-grant: I don't think the dt stuff in the eqn necessarily important at this time. The storage performance models are not yet flexible to non-hourly dt - so I think we can keep it as-is. These are intended to just show the differences in the CF calculations.

examples/test/test_all_examples.py
jaredthomas68: Does that mean it is different data, or just different starting point in the data?

docs/technology_models/storage_models_index.md
jaredthomas68: I think we should write our math correctly, even if we don't have our code written properly. The dt in the code is implicit in code because the time steps are 1 hour, but dt is technically there.

docs/technology_models/storage_models_index.md
jaredthomas68: I still disagree and would like to have the equations as I suggested. Please see my comment on the next equation.

elenya-grant: I added the dt in

docs/technology_models/storage_models_index.md
elenya-grant: okay. I added in dt

examples/test/test_all_examples.py
johnjasa: @elenya-grant, tell me if I'm thinking about this wrong, but this test was introduced as part of this PR, right? This is a fully new line as part of this PR. I don't think we need to resolve the differences in test values done during the development of the PR. This just shows that the PR introduces a new test and establishes the new test value. I don't think more work is needed here in this PR.

elenya-grant: definitely the second one but maybe the first one too. Using local timezone resource data, the wind CF is 18.119 and PV CF is 6.5585. Using the UTC timezone resource data, the wind CF is 18.119 and the solar resource data is 26.9244. So - big change in solar resource.

elenya-grant: @johnjasa  - you're totally correct.

docs/technology_models/storage_models_index.md
elenya-grant: I think for consistency with the rest of the doc page, that using `n_timesteps` and `discharge_rate` are more appropriate variables in this equation. I added in `*dt` and the `\` to improve formatting for the underscores in text variables. I also don't know how the `total time` variable you're proposing is different than `n_timesteps`. Is it equal to `n_timesteps*dt`? (in which case, wouldn't the `dt` cancel out from the numerator and denominator ?)

jaredthomas68: Ultimately, we are going for total_energy_produced/total_nameplate_energy.  I think we should shoot for having energy in the numerator and denominator even if terms would technically cancel.

examples/test/test_all_examples.py
jaredthomas68: Sounds good, thanks for the explanation.

docs/technology_models/storage_models_index.md
jaredthomas68: I'm fine with n_timesteps*dt

elenya-grant: heard - done!

jaredthomas68: I think this version would achieve what we need
```markdown
$
CF = \frac{\sum_{i=0}^{n_{\text{time\_steps}}}(\text{commodity\_out$_i$}*dt)}{\text{discharge\_rate}*n_{\text{time\_steps}}*dt}
$
```

jaredthomas68: I suggested switching _t to _i, but I can roll with where it is at now

jaredthomas68 (approved):
Thanks for iterating on this. I think the equations are technically correct. I made one suggestion, but non-blocking.

docs/technology_models/storage_models_index.md
elenya-grant: am I right that the only difference is the $\text{commodity}_i$ (I've been using $t$ though, so I'd make it $\text{commodity}_t$) instead of $\text{commodity}$ (it'd be commodity_out instead of commodity but GitHub doesn't like the underscores)?

If so - I just pushed up that change.

Issue #575, opened 2026-03-05
Roll resource data to a consistent timezone


johnjasa:
# Roll resource data to a consistent timezone

This is somewhat handled by shared default behavior where most resource data is downloaded in UTC.

jaredthomas68:
Close enough I decided not to make a separate issue. Some time zones (like MT) have multiple UTC offsets depending on the time of year due to daylight savings time. We need to make sure how we are handling time works for these cases.

PR #679, opened 2026-04-15
Demand Response Optimization for Battery Energy Storage

vijay092:
# Demand Response Optimization for Battery Energy Storage (Stage 1) 

This PR introduces a Pyomo-based formulation for demand response, which will be implemented in two stages.

As the first stage, this work implements a rolling horizon optimization for battery operations. The battery dispatch logic is based on a pre-defined signal, such as LMP, load, or a combination of both. This is the G&T level dispatch signal for demand response. The next stage will implement the peak load management logic. 

- [X] Feature Enhancement
  - [ ] Framework
  - [X] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [X] Open draft PR
- [X] Describe the feature that will be added

### TODO:
- [ ] Add unit tests
- [ ] Add documentation

**Structural feedback**: Is this in the right place? 

**Implementation feedback**: I used the same style as Gen's pyomo implementation. Appreciate feedback here. 

**Other feedback**:

- [X] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Documentation
- [ ] `CHANGELOG.md`

### Section 4.1: New Files
- Main Implementation: `h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py`

- Usage Example: `examples/34_plm_optimized_dispatch`

- `h2integrate/core/supported_models.py`
  - Added the new model

- [ ] **Model Structure**:
  - [x] If applicable add integration tests
- [x] **Example**: If applicable, a working example demonstrating the new model has been created
  - [x] Input file comments
  - [x] Run file comments
  - [x] Example has been tested and runs successfully in `test_all_examples.py`
- [ ] **Documentation**:
    - [x] Model documentation page added to the appropriate `docs/` section


h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,473 @@
...
+
+    max_charge_rate: float = field()
+    supervisory_signal: list = field()
+    peak_window: dict = field()
+    performance_incentive: float = field()
+    charge_efficiency: float = field(validator=range_val(0, 1), default=1.0)
+    discharge_efficiency: float = field(validator=range_val(0, 1), default=1.0)
+    n_max_events: int = field(default=10)
+    n_control_window: int = field(default=24 * 30)  # one month of hourly data
+    signal_threshold_percentile: float = field(default=0.0)
```
elenya-grant: Small comment - may be nice to add a `range_val` validator for the signal_threshold_percent (unless it can be over 100)?
```suggestion
    signal_threshold_percentile: float = field(default=0.0, validator=range_val(0,100))
```

This is a suggestion/thought/nitpick (but not necessary) - sometimes it's nice to use fraction instead of percentages (specifically with `range_val` validators) because it is easier to catch with validators.

Aka - if a parameter is defined as a percent (between 0 and 100), but a user thinks its defined as a fraction (suppose they want it to be 40%, but set the value to 0.4 because they think its a fraction), then this would not throw an error with a validator. But - if a parameter is defined as a fraction (between 0 and 1) and a user thinks its defined as a percentage (they want it to be 40% so they set the value to 40), then the validator _does_ thrown an error.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,473 @@
...
+    signal_threshold_percentile (float)
+        Percentile (0-100) used to compute the signal threshold for each
+        rolling window. Only hours at or above this percentile of the
+        available window signal are eligible for dispatch.
+        Default 0.0 (all hours eligible).
+    """
+
+    max_charge_rate: float = field()
+    supervisory_signal: list = field()
+    peak_window: dict = field()
```
elenya-grant, line 79: Is there a limit/constraint on `peak_window` specifically with respect to `n_control_window`?  Like what if someone defined `n_control_window = 6` and `peak_window["start"] = 12:00:00` and `peak_window["end"] = 21:00:00` (peak window is 9 hours, meaning that peak_window>n_control_window). 

Should there be a minimum value on `n_control_window`? (not necessary - just asking!)

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,473 @@
...
+        Should be in [0, 1].
+    supervisory_signal (list[float])
+        Price, demand, or price*demand forecast time series.
+        The rolling solver uses one window of length n_control_window per solve.
+    peak_window (dict)
+        Hours eligible for dispatch. 'start' and 'end' as HH:MM:SS strings.
+    performance_incentive (float)
+        Incentive revenue ($/kW per dispatch hour).
+    n_max_events (int)
+        Maximum discharge events per control window. Default 10.
```
elenya-grant: great doc-strings! These are so helpful!

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,473 @@
...
+        Should be in [0, 1].
+    discharge_efficiency (float)
+        Should be in [0, 1].
+    supervisory_signal (list[float])
+        Price, demand, or price*demand forecast time series.
+        The rolling solver uses one window of length n_control_window per solve.
+    peak_window (dict)
+        Hours eligible for dispatch. 'start' and 'end' as HH:MM:SS strings.
+    performance_incentive (float)
+        Incentive revenue ($/kW per dispatch hour).
```
elenya-grant:
```suggestion
        Incentive revenue ($/commodity_rate_units per dispatch hour).
```

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
elenya-grant: Just checking - can peak_window['end'] be before peak_window['start']? Like if I wanted to define my peak window from 11pm (23:00:00) to 3am (03:00:00) - would that work?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,473 @@
...
+
+        Returns
+        -------
+        pyomo.ConcreteModel
+            Fully formed model ready for solve.
+        """
+        m = pyomo.ConcreteModel(name="plm_dr")
+
+        # Parameters
+        P_max = self.config.max_charge_rate
```
elenya-grant: In order for the controller to be compatible with a case where we want to optimize the storage charge rate or capacity, then any method called in `compute()` should use `inputs["max_charge_rate"]` instead of `self.config.max_charge_rate` and `inputs["storage_capacity"]` instead of `self.config.max_capacity`. 

An example of how to test this is somewhat shown in `h2integrate/storage/test/test_storage_performance_model.py::test_generic_storage_with_simple_control_charge_rate_lessthan_demand`, using the `prob.set_val` function (happy to chat about it if you want)

elenya-grant (commented):
Hi Sanjana! Just putting a few high-level comments now - but I will do a deeper dive shortly! I love the thorough testing of `PLMOptimizedControllerConfig` and `PLMOptimizedStorageController`! The docstrings also look awesome! Besides the comments and questions that I left, other high-level things I'd like to see included before this is merged in are:
1. integration tests of the controller with the `StoragePerformanceModel` - similar to the tests in `h2integrate/control/control_strategies/storage/test/test_optimal_controllers.py`
2. a test for the example you added in `examples/test/test_all_examples.py` (this can be a pretty basic test, but it'll be good to ensure that the example you added is updated and runs without error as H2I is further developed)

Truly awesome work! Please message or call me if you want to discuss any of the comments I left! I will do another review that focuses on the logic within the controller - but wanted to get these initial questions and comments to you early on!

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Typically, percentiles are not written in terms of fractions (unlike percentages). Have you seen this done before?

vijay092: Also, that sounds good about the `range_val`. I hadn't added it because `np.percentile` will catch it anyway.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: No, there is no constraint on `peak_window` wrt `n_control_window`. `n_control_window` is just iterating over the 8760 is small chunks.

vijay092: `_compute_peak_window_mask` raises a `ValueError` if end < start.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I've only thought about this PR as a battery use case. We should chat about what this would mean.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
jaredthomas68: I think it would be worth including the validator even if np.percentile will catch it later because the validator will provide a very clear error message and appear readable for users. non-blocking comment.

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+# Optimized Demand Response Controller
+
+This controller optimizes the dispatch of a Battery Energy Storage System (BESS) based on a pre-defined supervisory signal. This pre-defined signal could be the Locational Marginal Price (LMP) or a demand profile. It could also be a $LMP\times demand$ depending on the application. 
+
+## Definitions
+
+**Given:**
+- $\lambda_t$ := `supervisory_signal`: price, demand, or price $\times$ demand time series at time $t$
+- $\mathcal{W}$ := `peak_window`: set of hours eligible for dispatch (e.g., 12:00--19:00)
+- $\gamma$ := performance incentive (\$/kW per dispatch hour)
```
jaredthomas68: is this truly required to be per hour, or is it per time-step? In other words, could it be per-minute if the user requested a smaller time step?

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+- $\bar{P}$ := `max_charge_rate` (kW): maximum charge and discharge rate, used as deemed capacity since the battery is assumed to always dispatch at full rated power
+- $E_{\max} :=$ `max_capacity` $\times$ (`max_soc_fraction` $-$ `min_soc_fraction`): usable energy capacity (kWh)
+- $\eta_c$ := `charge_efficiency`, $\quad \eta_d$ := `discharge_efficiency`
+- $\overline{\text{SoC}}$ := `max_soc_fraction`, $\quad \underline{\text{SoC}}$ := `min_soc_fraction`
+- `n_control_window` := Horizon length for optimization
+- $\mathcal{T} := \{0, 1, \ldots, T\}$: hourly time steps over `n_control_window`
+- $\mathcal{M}_m$ := set of hours in month $m$, for $m = 1, \ldots, 12$
```
jaredthomas68: Same as above, does this have to be hours?

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+**Given:**
+- $\lambda_t$ := `supervisory_signal`: price, demand, or price $\times$ demand time series at time $t$
+- $\mathcal{W}$ := `peak_window`: set of hours eligible for dispatch (e.g., 12:00--19:00)
+- $\gamma$ := performance incentive (\$/kW per dispatch hour)
+- $\bar{P}$ := `max_charge_rate` (kW): maximum charge and discharge rate, used as deemed capacity since the battery is assumed to always dispatch at full rated power
+- $E_{\max} :=$ `max_capacity` $\times$ (`max_soc_fraction` $-$ `min_soc_fraction`): usable energy capacity (kWh)
+- $\eta_c$ := `charge_efficiency`, $\quad \eta_d$ := `discharge_efficiency`
+- $\overline{\text{SoC}}$ := `max_soc_fraction`, $\quad \underline{\text{SoC}}$ := `min_soc_fraction`
+- `n_control_window` := Horizon length for optimization
+- $\mathcal{T} := \{0, 1, \ldots, T\}$: hourly time steps over `n_control_window`
```
jaredthomas68: Hours or time steps?

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+- $\mathcal{M}_m$ := set of hours in month $m$, for $m = 1, \ldots, 12$
+
+## Decision Variables
+
+- $u_t \in \{0, 1\}$ := discharge binary: 1 if battery dispatches at hour $t$, 0 otherwise
+- $v_t \in \{0, 1\}$ := charge binary: 1 if battery charges at hour $t$, 0 otherwise
+
```
jaredthomas68, line 159: hours or time steps?

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+## Optimization Problem
+
+This optimization is executed for each window, during which the performance model is invoked and the initial conditions are set.
+
+### Objective
+
+Maximize total annual incentive revenue:
+
+$$
+\max_{u_t,\, v_t} \quad \gamma \cdot \bar{P} \sum_{t \in \mathcal{T}} u_t
```
jaredthomas68: should the objective be based on max charge/discharge rate, or on actual discharge rate at each time step?

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+u_t = 0 \qquad \forall\, t \notin \mathcal{W}
+$$
+
+- Dispatch only on high supervisory signal:
+
+$$
+u_t = 1 \implies \lambda_t \geq \lambda^*_m \qquad \forall\, t \in \mathcal{M}_m
+$$
+
+where $\lambda^*_m$ is the threshold selecting the high LMP/peak load hours within month $m$.
```
jaredthomas68: I think it makes sense to use a threshold as you have it here. We may want to consider ramp rate or some additional criteria as well in the future.

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+
+- Maximum 10 events per month:
+
+$$
+\sum_{t \in \mathcal{M}_m} u_t \leq N_{\max} \qquad \forall\, m, \quad N_{\max} = 10
```
jaredthomas68: I think this constraint is limiting the number of hours (or time steps?) rather than the events, where each event could be multiple hours. I

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+$$
+
+- SoC evolution with charge and discharge:
+
+$$
+\text{SoC}_{t+1} = \text{SoC}_t + \frac{\eta_c \cdot v_t \cdot \bar{P}}{E_{\max}} - \frac{u_t \cdot \bar{P}}{\eta_d \cdot E_{\max}} \qquad \forall\, t \in \mathcal{T}
```
jaredthomas68: I think we need to use actual charge/discharge rate here instead of rated

jaredthomas68: We need to discuss whether we want to assume SOC is based on nameplate capacity or actual capacity. In other components we have based it on nameplate capacity, but here you are using actual capacity. I think we should pick a standard for H2I and stick with it. I'm up for conversation on this for sure and would be very interested to get feedback from others, such as @genevievestarke  and @aditiegarg9.

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,90 @@ tech_to_dispatch_connections: [
...
+$$
+
+- SoC bounds:
+
+$$
+\underline{\text{SoC}} \leq \text{SoC}_t \leq \overline{\text{SoC}} \qquad \forall\, t \in \mathcal{T}
```
jaredthomas68: Did you mean to include bars and underscores here to indicate max and min SOC?

docs/control/pyomo_controllers.md
jaredthomas68, line 1: Super helpful to have the math laid out, thanks!

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,123 @@
...
+    ]["supervisory_signal"]
+)[:N]
+
+sim = model.plant_config["plant"]["simulation"]
+n_timesteps = int(sim["n_timesteps"])
+dt_seconds = int(sim["dt"])
+tz = int(sim["timezone"])
+start = pd.Timestamp(sim["start_time"], tz=tz)
+freq = pd.to_timedelta(dt_seconds, unit="s")
+time_index = pd.date_range(start=start, periods=n_timesteps, freq=freq)
```
jaredthomas68: Can you switch to using the H2I utility for this?

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,123 @@
...
+
+battery_power = model.prob.get_val("battery.storage_electricity_discharge", units="kW")
+soc_pct = model.prob.get_val("battery.SOC", units="percent")
+
+peak_mask = (time_index.hour >= 14) & (time_index.hour <= 18)
+threshold_pct = np.percentile(lmp, percentile)
+discharge_mask = battery_power > 1.0
```
jaredthomas68: I think it would helpful to include methods or functions for these instead instead of asking the user to provide a peak mask, allow them to just provide the hours

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,123 @@
...
+dt_seconds = int(sim["dt"])
+tz = int(sim["timezone"])
+start = pd.Timestamp(sim["start_time"], tz=tz)
+freq = pd.to_timedelta(dt_seconds, unit="s")
+time_index = pd.date_range(start=start, periods=n_timesteps, freq=freq)
+
+battery_power = model.prob.get_val("battery.storage_electricity_discharge", units="kW")
+soc_pct = model.prob.get_val("battery.SOC", units="percent")
+
+peak_mask = (time_index.hour >= 14) & (time_index.hour <= 18)
```
jaredthomas68: I think the upper bound should be exclusive

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,123 @@
...
+            day + pd.Timedelta(hours=14),
+            day + pd.Timedelta(hours=18),
+            color="orange",
+            alpha=0.10,
+            linewidth=0,
+            zorder=0,
+        )
+
+
+w_discharge = discharge_mask[:time_window]
```
jaredthomas68: I think we will end up doing this a lot. Can we pull this information from the model rather than recalculating it? For some we may want to add convenience methods.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,123 @@
...
+)
+ax.set_ylabel("Power (MW)", fontsize=8)
+ax.tick_params(axis="x", labelrotation=30, labelsize=7)
+
+for ax in axes:
+    ax.tick_params(labelsize=7)
+    ax.grid(True, which="major", alpha=0.3, linewidth=0.5)
+
+plt.tight_layout()
+plt.savefig("plm_optimized_dispatch.png", dpi=150, bbox_inches="tight")
```
jaredthomas68: I think this is similar to what I did. Non-blocking comment here, but perhaps we can create some easy standard plot capabilities for the dispatch plots that pull the info directly from the model and make good plots without so much duplicate code.

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,54 @@
+name: technology_config
+description: 300 kW / 1.2 MWh battery with PLM MILP-optimized dispatch controller
```
jaredthomas68: Can you be more specific that this is for peak load management?

examples/34_plm_optimized_dispatch/plant_config.yaml
```diff
@@ -0,0 +1,12 @@
+name: plant_config
+description: Simulation for PLM MILP-optimized battery dispatch
+plant:
+  plant_life: 1
+  simulation:
+    n_timesteps: 8760     # full year (hourly)
+    dt: 3600              # 1-hour timesteps
+    timezone: -6          # MDT (UTC-6)
```
jaredthomas68: non-blocking, and definitely not for this PR, but some time zones, like MT have multiple UTC offsets depending on daylight savings time. I think we may need to address that explicitly somehow in the code. I don't know how yet.

jaredthomas68: I added a comment on this to #575

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,54 @@
...
+        
+      performance_parameters:
+        demand_profile: 0.0             
+      control_parameters:
+        system_commodity_interface_limit: 1.0e9
+        supervisory_signal: !include demand_profiles/supervisory_signal.yaml
+        peak_window:
+          start: "14:00:00"             # 2 pm MDT
+          end: "18:00:00"               # 6 pm MDT
+        performance_incentive: 14.0     # $/kW per dispatch hour
```
jaredthomas68: I think this should be per event, not per hour

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,54 @@
...
+        n_max_events: 10                # maximum discharge events per month
```
jaredthomas68: I think it may be helpful to allow for different event count periods, like per week or per year.

jaredthomas68: non-blocking

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,54 @@
...
+        signal_threshold_percentile: 95.0  # only dispatch when LMP >= this percentile
+        n_control_window: 10      # number of time steps in the control window (e.g., 10 hours)
```
jaredthomas68: I think it would be more general and useful to give a time code for the control wind, and rename to control_horizon. The backend can then determine the number of time steps in the horizon

docs/control/pyomo_controllers.md
jaredthomas68: I don't see how this objective will lead to reducing peaks. It seems like it could just reduce the total demand during the peak window without actually reducing the peak. I understand the threshold on percentile of peak in the horizon constraint, but I think we may want to add something to the objective related to reducing the peaks rather than just maximizing revenue. I think this is probably what you are holding for the next PR though.

jaredthomas68: I realized the objective here is what is supposed to manage for the supervisor, but revenue is for the local entity. The supervisor wants to reduce their peaks during the most critical times of the month, but I'm sure maximizing the local revenue will achieve that. We can assume that maximizing revenue locally will maximize revenue upstream as well, which assumes a perfect incentive structure. I think I've convinced myself this is ok for now, but let's make sure we continue to keep this in mind moving forward.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+        if len(self.time_index) != self.n_timesteps:
+            raise ValueError(
+                f"Time series length {len(self.time_index)} != n_timesteps {self.n_timesteps}"
+            )
+
+        self.in_peak_window = self._compute_peak_window_mask()  # bool array, shape (T,)
+        self.month_ids = self._compute_month_ids()              # int array,  shape (T,)
+
+    @staticmethod
+    def _build_time_index(plant_config: dict) -> pd.DatetimeIndex:
```
jaredthomas68: can you use the utility that came in with 641?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+            # respected across window boundaries.
+            events_used_per_month = {}
+
+            n_w = self.config.n_control_window
+            window_start_indices = list(range(0, self.n_timesteps, n_w))
+
+            for window_start in window_start_indices:
+                window_len = min(n_w, self.n_timesteps - window_start)
+
+                n_windows = len(window_start_indices)
```
jaredthomas68: n_w and n_windows are easy to conflate. see my naming suggestion on the doc page for n_control_window

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                report_every = max(1, n_windows // 4)
```
jaredthomas68: what is the reasoning on the  `//4`?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
jaredthomas68, line 1: This file would benefit from many more comments, specifically why something is being done

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                window_idx = window_start // n_w
+                if window_idx % report_every == 0:
+                    pct = round(window_start / self.n_timesteps * 100)
```
jaredthomas68: I think it is helpful to avoid relying on python for order of operations, particularly because it makes it hard to read
```suggestion
                    pct = round((window_start / self.n_timesteps) * 100)
```

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                    print(f"{pct}% done with PLM rolling dispatch")
```
jaredthomas68: I would like to protect any print statements, especially repetitive ones, with a verbosity input so the user can adjust the verbosity easily

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                    window_start=window_start,
+                    window_len=window_len,
+                    init_soc=self.updated_initial_soc,
+                    remaining_budget=remaining_budget,
+                )
+                self.problem_state = DispatchProblemState()
+
+                self.solve_dispatch_model(
+                    start_time=window_start,
+                    n_days=self.n_timesteps // 24,
```
jaredthomas68: Can you make this time-step length agnostic so the code can handle non-hourly time steps?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                    **performance_model_kwargs,
+                    sim_start_index=window_start,
+                )
+
+                # Performance model returns SOC in percent.
+                self.updated_initial_soc = soc_window[-1] / 100.0
+
+                for j in range(window_len):
+                    storage_out[window_start + j] = storage_out_window[j]
+                    soc_out[window_start + j] = soc_window[j]
```
jaredthomas68: could be list comprehension I think, but non-blocking

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+
+            print(
```
jaredthomas68: please protect this as I mentioned earlier, or remove. I don't think we need this print statement generally after development

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: That sounds good, I added it.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+        m.charge = pyomo.Var(m.T, domain=pyomo.Binary, doc="Charge binary: 1 = charging at timestep t")
+        m.soc = pyomo.Var(
+            m.T,
+            domain=pyomo.NonNegativeReals,
+            bounds=(soc_min, soc_max),
+            doc="State of charge SoC_t",
+        )
+
+        m.objective = pyomo.Objective(
+            expr=-incentive * P_max * sum(m.discharge[t] for t in m.T),
```
jaredthomas68: is there a good way to use possible discharge instead of rating (see my similar comment in docs)

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+
+        m.soc_init = pyomo.Constraint(expr=m.soc[0] == init_soc)
+
+        def soc_evolution_rule(mdl, t):
+            if t == 0:
+                return pyomo.Constraint.Skip
+            return mdl.soc[t] == (
+                mdl.soc[t - 1]
+                + eta_c * mdl.charge[t] * P_max / E_max
+                - mdl.discharge[t] * P_max / (eta_d * E_max)
```
jaredthomas68: see my comments in docs

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+                Used only for error messages and metrics. Defaults to 0.
+            n_days (int): Total simulation days. Passed to
+                ``DispatchProblemState.store_problem_metrics``.
+                Defaults to 0.
+
+        Raises:
+            RuntimeError: If GLPK returns a non-OK status or an
+                unacceptable termination condition.
+        """
+        from pyomo.opt import SolverStatus, TerminationCondition
```
jaredthomas68: why is the import statement inside the method? Can you move it to the header?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+            )
+
+        self.problem_state.store_problem_metrics(
+            solver_results,
+            start_time,
+            n_days,
+            pyomo.value(self.dr_model.objective),
+        )
+
+    def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
```
jaredthomas68: I think it generally makes sense to keep the standard methods (init, setup, and compute) at the top of the class definition and have all other methods come after.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,536 @@
...
+
+    @property
+    def storage_dispatch_commands(self) -> list:
+        """Net dispatch commands for the solved window.
+
+        Returns:
+            list[float]: ``(u_t - v_t) * P_max`` for each timestep in
+            the solved window. Positive = discharge, negative = charge.
+        """
+        P_max = self.config.max_charge_rate
```
jaredthomas68: I think P_max should be adjusted to not overcharge the battery, but still allow it to go to max charge. Maybe you have the handled somehow and I missed it.

jaredthomas68 (changes requested):
This is great Sanjana! I'm excited about where this is going. I gave a lot of comments, but some of them may be irrelevant due to my experience with pyomo. The doc page details were very helpful, thank you.

docs/control/pyomo_controllers.md
vijay092: I was assuming that the battery always charges and discharges at max discharge rate. If we want to solve for the actual discharge rate as well as when (the binary var) we want to discharge, the math becomes a little harder. I can work on that tomorrow.

docs/control/pyomo_controllers.md
vijay092: Yes.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
vijay092: Yep!

docs/control/pyomo_controllers.md
vijay092: I was under the impression that H2I only supports hourly timesteps. I can make this more general.

genevievestarke: We are currently in the process of enabling smaller time steps, so we are building the capability into new additions to the code!

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Totally fair point, I am working on it.

docs/control/pyomo_controllers.md
vijay092: Oh, that's awesome!

brookeslawski:
I've reviewed this PR from a high level (without running the code), mostly focusing on the documentation and overall functionality. The documentation explains the added features compared to existing models clearly, which is very helpful. Nice job!

docs/control/pyomo_controllers.md
vijay092: I have made changes to handle any `dt`.

docs/control/pyomo_controllers.md
vijay092: It can handle any timestep now.

docs/control/pyomo_controllers.md
vijay092: It can handle any timestep now.

docs/control/pyomo_controllers.md
vijay092: It can handle any timestep now.

docs/control/pyomo_controllers.md
vijay092: For now, I've assumed the battery charges and discharges at its maximum rate. I'll submit another PR to implement your version. When the state of charge approaches its limits, the optimizer handles edge cases as follows:

-If the SOC is too close to the minimum, discharging is disabled.
-If the SOC is too close to the maximum, charging is disabled.

docs/control/pyomo_controllers.md
vijay092: Great point! The way I have handled this is:
If we know the limit on the number of events, we can also get a limit on the number of timesteps because we know the event duration. See more in the docs.

docs/control/pyomo_controllers.md
vijay092: For now, I've assumed the battery charges and discharges at its maximum rate. I'll submit another PR to implement your version. When the state of charge approaches its limits, the optimizer handles edge cases as follows:

-If the SOC is too close to the minimum, discharging is disabled.
-If the SOC is too close to the maximum, charging is disabled.

vijay092: This formulation uses `max_charge_rate` throughout. I can jump on a call and explain this further.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
vijay092: Done

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
vijay092: I agree, and happy to do that. Do you want it in this PR?

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: Done

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: This becomes a bit complicated. What if the battery shows up for 3 out of 4 hours in the event duration?

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: Done

docs/control/pyomo_controllers.md
vijay092: That's right. Abhineet will implement that piece.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Done

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: If we had to rename it, I would change it to `control_horizon_timesteps` to align with standard MPC terminology.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I've added more comments.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I deleted print statements, so this is deleted.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Done

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Done

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: See my previous comments.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Sorry, moved it to the header.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I deleted print statements, so this went away.

examples/34_plm_optimized_dispatch/demand_profiles/demand_profile.yaml
abhineet-gupta, line 1: Is this file redundant or are you planning to use it later? I don't see it being referenced in the example.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I will submit another PR to solve for the optimal charge rate, and this comment will be addressed then. 
SOC will never go out of bounds because we have constraints on it.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Done

examples/34_plm_optimized_dispatch/demand_profiles/demand_profile.yaml
vijay092: I wanted to use it show how to use demand*LMP in the example but I think it is too much. Deleted it.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I changed the code to use `inputs`. Could you check if that's what you had in mind?

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
vijay092: The only way I can think of to get access to the methods from model is:
`controller = model.control_strategies[0]`
`pw_start, pw_end = controller._parse_peak_window()`. I implemented this but let me know if you have other suggestions.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
jaredthomas68: I agree that it makes sense to make this general for storage since there is not anything that is really battery specific.

docs/control/pyomo_controllers.md
jaredthomas68: Thank you!

docs/control/pyomo_controllers.md
jaredthomas68: Does this mean that a one hour battery with a 0.1 min soc and 0.9 max soc will never charge or discharge because it will be too close to the limits? If so, then I think we should adjust this PR. If the battery can still reach the max and min allowable SOC then I am happy with where this is at.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
jaredthomas68: Thanks!

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
jaredthomas68: A latter PR is fine, and it does not have to be you. Any of us working on dispatch and control could tackle it.

examples/34_plm_optimized_dispatch/tech_config.yaml
jaredthomas68: It depends on the contract. Some contracts language can include separate language for responding to an event and responding with a given amount of energy. We would ideally track both events and hours in the code.

docs/control/pyomo_controllers.md
```diff
@@ -61,65 +61,78 @@ tech_to_dispatch_connections: [
 
 # Optimized Demand Response Controller
 
-This controller optimizes the dispatch of a Battery Energy Storage System (BESS) based on a pre-defined supervisory signal. This pre-defined signal could be the Locational Marginal Price (LMP) or a demand profile. It could also be a $LMP\times demand$ depending on the application. 
+This controller optimizes the dispatch of a Battery Energy Storage System (BESS) based on a pre-defined supervisory signal. This signal could be the Locational Marginal Price (LMP), a demand profile, or a $LMP\times demand$ product depending on the application. The objective is to maximize incentive payments to the battery, subject to constraints on the maximum number of dispatch events per month and on the battery state of charge.
```
jaredthomas68:
```suggestion
This controller optimizes the dispatch of a Battery Energy Storage System (BESS) based on a pre-defined supervisory signal. This signal could be the Locational Marginal Price (LMP), a demand profile, or an $LMP\times demand$ product depending on the application. The objective is to maximize incentive payments to the battery, subject to constraints on the maximum number of dispatch events per month and on the battery state of charge.
```

docs/control/pyomo_controllers.md
jaredthomas68: I'm not sure why, but the under and over bars are not rendering in the docs, at least in the tests.

docs/control/pyomo_controllers.md
```diff
@@ -134,14 +147,19 @@ $$
...
 $$
-v_t = 0 \qquad \forall\, t \in \mathcal{W}
+v_t = 0 \qquad \forall\, t \in \mathcal{D}
 $$
 
 - Binary variables:
 
 $$
-u_t \in \{0, 1\}, \quad v_t \in \{0, 1\}, \quad \text{SoC}_t \in [0, 1] \qquad \forall\, t, m
+u_t \in \{0, 1\}, \quad v_t \in \{0, 1\}, \quad \text{SoC}_t \in [\underline{\text{SoC}},\, \overline{\text{SoC}}] \qquad \forall\, t
```
jaredthomas68: the underline and overline are not rendering in the test docs build

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -27,51 +30,77 @@
...
+pw_start, pw_end = controller._parse_peak_window()
+pw_start_h = pw_start.hour
+pw_end_h = pw_end.hour
+
+control_params = model.technology_config["technologies"]["battery"]["model_inputs"][
+    "control_parameters"
+]
+event_dur_cfg = control_params.get("event_duration")
+
+half_td = None
```
jaredthomas68, line 55: did you mean to name this "half_dt"?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -109,119 +158,74 @@ def setup(self):
...
-        self.month_ids = self._compute_month_ids()              # int array,  shape (T,)
+        self.month_ids = self._compute_month_ids()  # int array, shape (T,)
+
+        # Computes the number of timesteps in an event based on event_duration and dt_seconds,
+        # rounded to nearest int and at least 1.
+        if self.config.event_duration is not None:
+            self.steps_per_event: int = max(
+                1,
+                int(
+                    round(
```
jaredthomas68: would it make sense to enforce rounding up here to make sure the full event duration is included?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
jaredthomas68: The floor division here still assumes 24 time steps per day.

docs/control/pyomo_controllers.md
vijay092: Hi @jaredthomas68, I'm now solving for the optimal power (not assuming it charges/discharges at rated) . Check out the new formulation.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Made changes to solve for the optimal charge rate.

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: So I added an option for the user to provide incentive/event. This internally gets converted to a $/kwh when running the optimization. Happy to chat more about improving this.

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,122 @@ tech_to_dispatch_connections: [
   ["battery", "battery"],
 ]
 ```
+
+# Optimized Demand Response Controller
+
+This controller optimizes the dispatch of a Battery Energy Storage System (BESS) based on a pre-defined supervisory signal. This signal could be the Locational Marginal Price (LMP), a demand profile, or a $LMP\times demand$ product depending on the application. The objective is to maximize incentive payments to the battery, subject to constraints on the maximum number of dispatch events per month and on the battery state of charge.
+
+The controller works at any simulation timestep resolution (`dt`). All time-based parameters ( `event_duration`, `min_peak_separation`) are specified in physical time units (hours, minutes, etc.) and are internally converted to timesteps using `dt`.
```
abhineet-gupta, line 118: Perhaps we should include a short description of event_duration in the docs.

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,122 @@ tech_to_dispatch_connections: [
...
+- $\mathcal{W}$ := `peak_window`: set of timesteps eligible for dispatch (e.g., 12:00--20:00 each day)
+- $\lambda_*$ := signal threshold = `signal_threshold_percentile`-th percentile of $\lambda_t$ over $\mathcal{W}$
+- $\mathcal{E}$ := eligible peak timesteps: $\{t \in \mathcal{W} : \lambda_t \geq \lambda_*\}$, respecting `min_peak_separation`
+- $\mathcal{D}$ := dispatch window:  $\pm\,\tfrac{\text{event\_duration}}{2}$  around each peak in $\mathcal{E}$
+- $\gamma$ := `performance_incentive` (\$/kWh): incentive revenue per kWh discharged
+- $\bar{P}$ := `max_charge_rate` (kW): maximum charge and discharge rate
+- $E_{\max} :=$ `max_capacity` $\times$ (`max_soc_fraction` $-$ `min_soc_fraction`): usable energy capacity (kWh)
+- $\eta_c$ := `charge_efficiency`, $\quad \eta_d$ := `discharge_efficiency`
+- $\overline{\text{SoC}}$ := `max_soc_fraction`, $\quad \underline{\text{SoC}}$ := `min_soc_fraction`
+- `n_control_window_hours` := rolling horizon length in hours; converted to $T = \lceil \text{n\_control\_window\_hours} / \Delta t \rceil$ timesteps
```
abhineet-gupta: Might need to remove `\` from inside math mode in "n_control_window_hours"

docs/control/pyomo_controllers.md
```diff
@@ -58,3 +58,122 @@ tech_to_dispatch_connections: [
...
+
+## Definitions
+
+**Given:**
+- $\lambda_t$ := `supervisory_signal`: price, demand, or price $\times$ demand time series at timestep $t$
+- $\Delta t$ := simulation timestep duration (hours), derived from `dt` in the plant config
+- $\mathcal{W}$ := `peak_window`: set of timesteps eligible for dispatch (e.g., 12:00--20:00 each day)
+- $\lambda_*$ := signal threshold = `signal_threshold_percentile`-th percentile of $\lambda_t$ over $\mathcal{W}$
+- $\mathcal{E}$ := eligible peak timesteps: $\{t \in \mathcal{W} : \lambda_t \geq \lambda_*\}$, respecting `min_peak_separation`
+- $\mathcal{D}$ := dispatch window:  $\pm\,\tfrac{\text{event\_duration}}{2}$  around each peak in $\mathcal{E}$
```
abhineet-gupta: Might need to remove `\` from inside math mode in "event_duration".

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,59 @@
...
+      model: StoragePerformanceModel
+    cost_model:
+      model: ATBBatteryCostModel
+    control_strategy:
+      model: PeakLoadManagementOptimizedStorageController
+    model_inputs:
+      shared_parameters:
+        commodity: electricity
+        commodity_rate_units: kW
+        max_charge_rate: 300.0  # kW rated power
```
abhineet-gupta, line 15: Do we need separate charge and discharge rates for the battery?

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,59 @@
...
+        init_soc_fraction: 0.90
+        charge_efficiency: 0.95
+        discharge_efficiency: 0.95
+      performance_parameters:
+        demand_profile: 0.0
+      control_parameters:
+        system_commodity_interface_limit: 1.0e9
+        supervisory_signal: !include demand_profiles/supervisory_signal.yaml
+        peak_window:
+          start: 12:00:00  # 2 pm MDT
```
abhineet-gupta: Comments on these lines do not match the times.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
elenya-grant: You're right - I haven't seen a percentile written in terms of fractions. I think that with the addition of the range_val - that this is good for me! Thanks!

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
elenya-grant: thanks for clarifying!

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
elenya-grant: yes! It looks great to me now!

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -0,0 +1,59 @@
...
+        system_commodity_interface_limit: 1.0e9
+        supervisory_signal: !include demand_profiles/supervisory_signal.yaml
+        peak_window:
+          start: 12:00:00
+          end: 20:00:00
+        performance_incentive_per_event: 16800.0  # $/event
+        n_max_events: 10  # maximum discharge events per month
+        signal_threshold_percentile: 95.0  # only dispatch when LMP >= this percentile
+        n_control_window_hours: 24  # hours; converted to timesteps using simulation dt
+        event_duration:  # omit or set to null to use static peak_window
```
abhineet-gupta, line 34: Do we need to add another constraint `(event_duration==null) or (min_peak_window == null) or (event_duration <= min_peak_separation)`?

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: Fixed

examples/34_plm_optimized_dispatch/tech_config.yaml
vijay092: Spoke to Elenya. Some performance models don't have separate charge and discharge rates, so I'll leave it as is for now and come back later.

docs/control/pyomo_controllers.md
vijay092: Done

docs/control/pyomo_controllers.md
vijay092: Done

docs/control/pyomo_controllers.md
vijay092: Done

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,748 @@
...
+                changes to those values are reflected in each solve.
+
+        Returns:
+            callable: ``pyomo_dispatch_solver(performance_model,
+            performance_model_kwargs, inputs)`` that iterates over the
+            simulation in windows of ``n_control_window_hours`` timesteps.
+            For each window it:
+
+            1. Builds a fresh MILP from the window's signal slice.
+            2. Solves the MILP with GLPK.
```
genevievestarke:
```suggestion
            2. Solves the MILP with a MILP solver
```
This is a bit of a silly line that I'm suggesting, but I think we should take GLPK out of the comment and keep it more general

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,748 @@
...
+
+        Args:
+            start_time (int): Timestep index of the window start.
+                Used only for error messages and metrics. Defaults to 0.
+            n_days (int): Total simulation days. Passed to
+                ``DispatchProblemState.store_problem_metrics``.
+                Defaults to 0.
+
+        Raises:
+            RuntimeError: If GLPK returns a not OK status or an
```
genevievestarke:
```suggestion
            RuntimeError: If solver returns a not OK status or an
```

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,748 @@
...
+        # Expand eligible timesteps into event windows based on event_duration
+        dispatch_window_w = self._compute_event_window_mask(eligible_t_w)
+        eligible_t_w = dispatch_window_w
+
+        months_in_window = np.unique(month_ids_w).tolist()
+
+        # Sets we need to iterate on in the constraints and objective
+        m.T = pyomo.Set(initialize=range(window_len), doc="Timesteps in window")
+        m.M = pyomo.Set(initialize=months_in_window, doc="Months in window")
+
```
genevievestarke, line 575:
```suggestion

## Variables
```

genevievestarke (approved):
Looks good to me! Thank you @vijay092 for adding in this capability! 
I had a few suggestions for making the solver references more general.

examples/test/test_all_examples.py
```diff
@@ -2782,3 +2784,150 @@ def test_multivariable_streams_example(subtests, temp_copy_of_example):
...
+        )
+        grid_purchase = model.prob.get_val("grid_buy.electricity_out", units="kW")
+        assert battery_unmet_demand.sum() == pytest.approx(grid_purchase.sum(), rel=1e-3)
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "example_folder,resource_example_folder", [("34_plm_optimized_dispatch", None)]
+)
+def test_plm_optimized_dispatch_example(subtests, temp_copy_of_example):
```
jaredthomas68, line 2993: Thanks for the added tests

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -232,17 +237,69 @@ def test_optimizer_dispatch_respects_soc_constraints(base_config):
...
-        charge = pyomo.value(model.charge[t])
-        discharge = pyomo.value(model.discharge[t])
-        soc += charge - discharge
-        assert soc >= base_config.min_soc_fraction * base_config.max_capacity
-        assert soc <= base_config.max_soc_fraction * base_config.max_capacity
+        p_charge = pyomo.value(model.p_charge[t])  # type: ignore[index]
+        p_discharge = pyomo.value(model.p_discharge[t])  # type: ignore[index]
+        if t > 0:
+            soc += eta_c * p_charge * dt_hours / E_max - p_discharge * dt_hours / (eta_d * E_max)
+        assert soc >= base_config.min_soc_fraction - 1e-6
```
jaredthomas68: It is best to use subtests if you have multiple assert statements in a single test. There are many examples of subtests in our test suite. Can you make sure that all your tests with multiple asserts are protected with subtests?

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -232,17 +237,69 @@ def test_optimizer_dispatch_respects_soc_constraints(base_config):
...
+    """A later peak within min_peak_separation of an earlier peak is dropped."""
+    controller = _make_controller()
+    controller.config = SimpleNamespace(
+        signal_threshold_percentile=50.0,
+        min_peak_separation={"units": "h", "val": 3},
+    )
+    controller.dt_seconds = 3600
+    signal = np.array([1.0, 1.0, 1.0, 1.0, 1.0, 10.0, 1.0, 8.0, 1.0, 1.0])
+    mask = controller._compute_eligible_mask(signal)
+    print(mask)
```
jaredthomas68: Please remove print statements

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -254,11 +311,219 @@ def test_optimizer_dispatch_respects_charge_discharge_exclusivity(base_config):
...
+        "P_max": 1.0,
+        "storage_capacity": 10.0,
+    }
+    model_kwh = _make_controller_with_config(config_kwh)._build_dr_model(**build_kwargs)
+    model_event = _make_controller_with_config(config_event)._build_dr_model(**build_kwargs)
+
+    PeakLoadManagementOptimizedStorageController.glpk_solve_call(model_kwh)
+    PeakLoadManagementOptimizedStorageController.glpk_solve_call(model_event)
+
+    for t in range(24):
```
jaredthomas68, line 416: If testing multiple cases in a loop, it is best to use subtests. Subtests allow all tests to complete even if one in the loop fails.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -254,11 +311,219 @@ def test_optimizer_dispatch_respects_charge_discharge_exclusivity(base_config):
...
+    prob.setup()
+    prob.run_model()
+
+    soc = prob.get_val("SOC", units="unitless")
+    # print
+    print("SOC:", soc)
+    discharge = prob.get_val("storage_electricity_discharge", units="kW")
+    print("Discharge:", discharge)
+
+    with subtests.test("SOC never below min"):
```
jaredthomas68, line 512: Thank you for using subtests here!

jaredthomas68 (changes requested):
This is looking really good! Thank you for addressing the feedback. I think there are a few more changes that need to be made before merging, mostly some minor adjustments to the tests.

docs/control/pyomo_controllers.md
vijay092: How are you seeing the preview? It looks fine to me. Is the overbar problem only on SOC, or on P too? 
I went ahead and changed the notation to SOC_min and SOC_max.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
vijay092: td is short for `timedelta`. Half the event duration as `pd.Timedelta`.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: Currently, a 3.5-timestep event rounds to 4 steps and a 3.4-timestep event rounds to 3 steps. For our use case, ceil makes more sense- I'd rather dispatch one extra step than miss revenue. Good catch, changed it to ceil.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -0,0 +1,126 @@
...
+ax.plot(time_index[:time_window], soc_pct[:time_window], color="g", linewidth=1.0)
+ax.axhline(90, color="gray", linestyle=":", linewidth=0.7)
+ax.axhline(10, color="gray", linestyle=":", linewidth=0.7)
+ax.set_ylabel("SOC (%)", fontsize=8)
+ax.set_ylim([0, 105])
+
+
+plt.tight_layout()
+plt.savefig(
+    "examples/34_plm_optimized_dispatch/plm_optimized_dispatch.png", dpi=150, bbox_inches="tight"
```
abhineet-gupta: This line prevents the running the example from inside the example 34 directory. Using the `EXAMPLE_DIR` variable here should solve this.

abhineet-gupta (approved):
The rest of the PR looks good to me.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -304,7 +303,7 @@ def pyomo_dispatch_solver(
                 # Solve the optimzation problem
                 self.solve_dispatch_model(
                     start_time=window_start,
-                    n_days=self.n_timesteps // 24,
+                    n_days=int(round(self.n_timesteps * self.dt_seconds / 86400)),
```
jaredthomas68: I think ceiling would be more appropriate here so we don't have 0 days. I'm not positive though, so this is a non-blocking comment.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -67,16 +67,18 @@ def base_config():
...
-    assert end.minute == 40
-    assert end.second == 20
+    with subtests.test("start time"):
+        assert start.hour == 8
+        assert start.minute == 0
+        assert start.second == 0
+    with subtests.test("end time"):
+        assert end.hour == 18
+        assert end.minute == 40
+        assert end.second == 20
```
jaredthomas68: Technically I think these asserts should separated, but they are close enough I could let it go. Non-blocking.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -253,12 +261,13 @@ def test_optimizer_dispatch_respects_soc_constraints(base_config):
         p_discharge = pyomo.value(model.p_discharge[t])  # type: ignore[index]
         if t > 0:
             soc += eta_c * p_charge * dt_hours / E_max - p_discharge * dt_hours / (eta_d * E_max)
-        assert soc >= base_config.min_soc_fraction - 1e-6
-        assert soc <= base_config.max_soc_fraction + 1e-6
+        with subtests.test(f"SOC in bounds at t={t}"):
+            assert soc >= base_config.min_soc_fraction - 1e-6
+            assert soc <= base_config.max_soc_fraction + 1e-6
```
jaredthomas68: Two asserts in one subtest should be separated. Non-blocking.

jaredthomas68 (approved):
I think is looking good. I left a few suggestions, but no blocking comments.

docs/control/pyomo_controllers.md
```diff
@@ -110,3 +110,124 @@ tech_to_dispatch_connections: [
...
+
+- $u_t \in \{0, 1\}$ := discharge binary: 1 if a discharge event is active at timestep $t$; used for event counting and window feasibility constraints only
+- $v_t \in \{0, 1\}$ := charge binary: 1 if a charge event is active at timestep $t$
+- $p_{d,t} \in [0,\, P_{\max}]$ := discharge power (kW) actually dispatched at timestep $t$
+- $p_{c,t} \in [0,\, P_{\max}]$ := charge power (kW) actually consumed at timestep $t$
+- $\text{SoC}_t \in [\text{SoC}_{\min},\, \text{SoC}_{\max}]$ := state of charge (fraction) at timestep $t$
+
+## Optimization Problem
+
+This optimization is executed for each rolling window. At each window boundary the terminal SoC is carried forward as the initial condition for the next window.
```
aditiegarg9, line 162: It might be good to highlight the variable used for rolling window. I believe you are incorporating MPC here?

aditiegarg9 (approved):

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -0,0 +1,748 @@
...
+    Each call to the dispatch solver iterates over the full simulation in
+    windows of length ``n_control_window_hours``. For each window it builds
+    and solves a MILP that maximizes incentive revenue, then passes the
+    resulting dispatch commands to the performance model. The terminal SOC
+    of each window is carried forward as the initial SOC of the next window.
+    """
+
+    dr_model: Any
+    problem_state: DispatchProblemState
+
```
jaredthomas68: This is where you add the line for time step bounds. Right now the hour limits are coming from the PyomoStorageControllerBaseClass inheritance.

You can adjust as shown below, which indicates compatibility with time steps between 5 minutes and 1 hour.
```suggestion
    _time_step_bound = (300, 3600)
```

PR #698, opened 2026-04-23
Ability to have a user-defined resource model

elenya-grant:
# Ability to have a user-defined resource model

Currently, H2I allows for users to provide custom finance models and custom technology models. With this PR, users can also specify a custom resource model!

```yaml
sites:
  site:
    latitude: 32.34
    longitude: -98.27
    resources:
      river_resource:
        resource_model: CustomRiverResource
        resource_model_location: river_resource/river_resource_model.py
        resource_parameters:
          filename: river_data.csv
```

Another update was made to the custom model definition. The custom model definition was still using the logic of "nicknaming" classes in `supported_models` (we have since updated the model naming convention to just use the model class name). Now - the custom models are simply defined by the class name and the model location (rather than requiring a nickname). For example, before we would've done this:

```yaml
resource_model: custom_river_resource
resource_model_class_name: CustomRiverResource
resource_model_location: river_resource/river_resource_model.py
```

but now we do this:

```yaml
resource_model: CustomRiverResource
resource_model_location: river_resource/river_resource_model.py
```

- [ ] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [ ] `CHANGELOG.md`

### Section 5.1: New Files
N/A

- `H2IntegrateModel` in `h2integrate/core/h2integrate_model.py`
  - `collect_custom_models()`: added logic to check for custom resource models
  -  `create_custom_models()`: minor update to an error message
- `h2integrate/core/test/test_framework.py`: added test `test_custom_resource_model` for new functionality leveraging example 7
- `docs/resource/resource_index.md`: updated doc page to describe how to add a custom resource model

### DIFF
--- a/docs/resource/resource_index.md
+++ b/docs/resource/resource_index.md
@@ -6,8 +6,40 @@
 - [Tidal Resource Data](tidal_resource:models)
 
 
+## Setting resource data for a technology
 
-## Custom resource models
+There are two ways to supply resource data to a technology:
+1. Set resource for a technology data [using `set_val()`](#resource-data-specified-using-setval)
+2. Create a [custom resource model](#custom-resource-models) for the technology
+
+
+
+### Resource data specified using `set_val()`
+
+Resource data for a technology can be set using the `set_val()` command. In the [Run of River Example](https://github.com/NatLabRockies/H2Integrate/tree/develop/examples/07_run_of_river_plant/), the technology named `river` needs a resource input called `discharge`. An example of this is shown below:
+
+```python
+import pandas as pd
+from h2integrate.core.h2integrate_model import H2IntegrateModel
+# Create an H2I model
+h2i = H2IntegrateModel("07_run_of_river.yaml")
+
+# Suppose we load the resource data from a csv file
+resource_df = pd.read_csv("river_resource_data.csv")
+
+# Setup the h2i model
+h2i.setup()
+
+# Set the resource data for the river
+h2i.set_val("river.discharge", val=resource_df["discharge"].values, units="m**3/s")
+
+# Run the model
+h2i.run()
+```
+
+
+### Custom resource models
+The benefit of using a custom resource model is that the resource data can be made to vary for different inputs, which can be beneficial if running a design sweep or optimization where the resource location (specified by a latitude and longitude) is a design variable.
 
 A general resource model can be defined similarly to a custom technology model. A custom resource model should be defined in the plant configuration file within a site section under `sites`.
 
--- a/docs/resource/resource_index.md
+++ b/docs/resource/resource_index.md
@@ -12,7 +12,7 @@
 A general resource model can be defined similarly to a custom technology model. A custom resource model should be defined in the plant configuration file within a site section under `sites`.
 
 ```{note}
-Note that all custom resource models must have inputs of `latitude` and `longitude`
+Note that all custom resource models must have inputs of `latitude` and `longitude`. The outputs of your custom resource model should match the expected input to whatever model its connected to.
 ```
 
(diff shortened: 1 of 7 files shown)


docs/resource/resource_index.md
```diff
@@ -4,3 +4,34 @@
 - [Wind Resource Data](wind_resource:models)
 - [Solar Resource Data](solar_resource:models)
 - [Tidal Resource Data](tidal_resource:models)
+
+
+
+## Custom resource models
```
kbrunik: Do you think it would be worth highlighting that you could also include resource in your run script as:
```
t_vel = np.ones(8760)*30
prob.set_val("tidal_velocity", t_vel, units = "m/s")
```

I thought about using the custom resource models but I really just need to set a single input stream value so it felt like a lot of work.

h2integrate/core/h2integrate_model.py
```diff
@@ -250,15 +250,15 @@ def create_custom_models(self, model_config, config_parent_path, model_types, pr
...
+                        model_class_name = config[model_type].get(f"{prefix}model")
                         if (
                             model_class_name
                             != included_custom_models[model_name]["model_class_name"]
                         ):
                             raise (
                                 ValueError(
                                     "User has specified two custom models using the same model"
-                                    "name ({model_name}), but with different model classes. "
+                                    f"name ({model_name}), but with different model classes. "
```
kbrunik, line 261: nice

h2integrate/core/h2integrate_model.py
```diff
@@ -324,7 +324,7 @@ def collect_custom_models(self):
         self.create_custom_models(
             self.technology_config["technologies"],
             self.tech_parent_path,
-            ["performance_model", "cost_model", "finance_model"],
+            ["performance_model", "cost_model", "finance_model", "control_strategy"],
```
kbrunik: I have to be honest, I'm a little worried about allowing custom control strategies right now. With the team working to enable system level dispatch and no testing or safeties in place for a custom control strategy it feels a little risky. Do you think we could make an issue about this and leave this for another PR?

kbrunik (changes requested):
Thanks for adding the custom resource models functionality, I think it makes sense for someone to be able to use whatever they want for resource data. I could see adding some tooling to help take "custom" resource data and making it more compatible with some of the converter models (i.e., generating the meta data for the header in pysam resource files).

I really like the clean up to not require a nickname, since the info was redundant and it looks a little cleaner now.

I do feel pretty strongly about not enabling the ability for custom control strategies right now, but happy to talk about it. 

Can you also update the `CHANGELOG. md`?

docs/resource/resource_index.md
elenya-grant: I think it could be highlighted - but I'm also wondering if `resource_index.md` is the right place to note that? It almost feels like the general idea of using `set_val` should be noted in a user-guide doc page?

h2integrate/core/h2integrate_model.py
elenya-grant: fair fair - yes - I suppose I was just thinking that it may have been missed in the past but I totally see the reason for keeping it out for now.

elenya-grant: I removed it!

docs/resource/resource_index.md
kbrunik: hmmm, I do wonder if it would make sense to have it go in both places. I feel like if someone is trying to set the resource in a funky way they would go to the resource page

elenya-grant: okay! I updated the resource doc page!

kbrunik (approved):
Thanks for the changes!

PR #700, opened 2026-04-24
Add per-year pricing support for Grid and Feedstock cost models

### LINKED ISSUE #660: Update Grid and Feedstock so prices can be provided per-year
# Update Grid and Feedstock so prices can be provided per-year

@vijay092 requested this for the grid model, but I thought it was relevant for the feedstock model as well. Currently, both of these models can take price inputs in as either a single value or as a list/array that has a length equal to `n_timesteps`. It'd be nice if these options were expanded to include a case where the prices would be specified as a list/array with a length of `plant_life`. 

This can be done pretty easily by updating the logic in `GridCostModel` like below (but it should also be updated for the `electricity_buy_price` shape)
```python
if self.config.electricity_buy_price is not None:
    buy_price = self.config.electricity_buy_price
    if isinstance(buy_price, list | np.ndarray):
        if len(buy_price) != n_timesteps or len(buy_price) != plant_life:
            raise ValueError(
                f"electricity_buy_price length ({len(buy_price)}) "
                f"must match n_timesteps ({n_timesteps}) or plant_life ({plant_life})"
            )
        
        buy_price_shape = len(buy_price)
    else:
        buy_price_shape = 1
```

However, there would be some additional considerations to handle variable `dt` or varying simulation length cases. I will update this issue with that proposed solution later.

johnjasa:
# Add per-year pricing support for Grid and Feedstock cost models

Adds the ability to specify `electricity_buy_price`, `electricity_sell_price` (Grid), and `price` (Feedstock) as arrays of length `plant_life` for per-year pricing. Previously only scalar or per-timestep (length `n_timesteps`) arrays were supported.

A shared `determine_price_mode()` utility function handles validation and mode detection for reuse across models. When `n_timesteps == plant_life`, the per-year interpretation takes priority with a `UserWarning`.

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [x] Tools/Utilities
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues
Resolves #660

### Section 5.1: New Files
None.

- `h2integrate/core/utilities.py`
  - `determine_price_mode`: New utility function to classify price arrays as scalar, per-timestep, or per-year.
- `h2integrate/converters/grid/grid.py`
  - `GridCostModel.setup`: Refactored to use `determine_price_mode` for buy/sell price validation.
  - `GridCostModel.compute`: Updated VarOpEx calculation to produce per-year array for per-year pricing.
- `h2integrate/core/feedstocks.py`
  - `FeedstockCostModel.setup`: Refactored to use `determine_price_mode` for price validation.
  - `FeedstockCostModel.compute`: Updated VarOpEx calculation to produce per-year array for per-year pricing.
- `docs/technology_models/grid.md`
  - Updated input table and added Price Input Modes section.
- `docs/technology_models/feedstocks.md`
  - Updated price parameter docs and tip box.
- `h2integrate/converters/grid/test/test_grid.py`
  - Added 6 tests for per-year buy/sell pricing, invalid lengths, ambiguity warning, and non-hourly dt.
- `h2integrate/core/test/test_feedstocks.py`
  - Added 3 tests for per-year pricing, invalid lengths, and ambiguity warning.


h2integrate/core/utilities.py
```diff
@@ -104,6 +105,63 @@ def merge_shared_inputs(config, input_type):
...
+    - A scalar (constant for all timesteps and years).
+    - An array of length ``n_timesteps`` (per-timestep pricing).
+    - An array of length ``plant_life`` (per-year pricing).
+
+    When ``n_timesteps == plant_life`` and the price is an array of that length,
+    the per-year interpretation will be used and a :class:`UserWarning` is
+    raised for clarity.
+
+    Args:
+        price: The price value(s). Can be a scalar (int/float), list, or
```
elenya-grant: Nonblocking nitpick - the `price` arg doesn't have a type specified (although there are many options - as made clear in the description). But for some consistency I think it'd be nice to maybe have something simple like this:

```suggestion
        price (float | list[float] | np.ndarray): The price value(s). Can be a scalar (int/float), list, or
```

Again - this is really more a thought than a suggestion - up to you whether to take it or leave it!

h2integrate/converters/grid/grid.py
```diff
@@ -198,6 +198,7 @@ def setup(self):
         super().setup()
 
         n_timesteps = self.options["plant_config"]["plant"]["simulation"]["n_timesteps"]
+        plant_life = int(self.options["plant_config"]["plant"]["plant_life"])
```
elenya-grant, line 216: Do you think `plant_life` and `n_timesteps` should be set as attributes in the `CostModelBaseClass`? It may require updating some tests - but I could see it being more useful across other cost models (perhaps this could be a future PR)

h2integrate/converters/grid/grid.py
```diff
@@ -281,21 +277,30 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
...
         # Add buying costs if buy price is configured
         # electricity_out represents power flowing OUT of grid (buying)
         if self.config.electricity_buy_price is not None:
             electricity_out = inputs["electricity_out"]
             buy_price = inputs["electricity_buy_price"]
-            # Buying costs money (positive VarOpEx)
-            varopex += np.sum((self.dt / 3600) * electricity_out * buy_price)
+            if self._buy_price_mode == "per_year":
+                # Per-year price: total energy * price per year
+                total_energy_bought = np.sum(electricity_out) * (self.dt / 3600)
```
elenya-grant: Nonblocking: What if instead of always using `electricity_sold` and `electricity_out` as inputs to the cost model - what if the inputs were based on the `_buy_price_mode`? This would perhaps require adding more outputs to the grid performance model. But - it could be something like:

```python
if self._buy_price_mode == "per_year"
   self.add_input("annual_electricity_out", val=0.0, shape=plant_life, units="kW*h/yr")
else:
   self.add_input("electricity_out", val=0.0, shape=n_timesteps, units="kW")

if self._sell_price_mode == "per_year"
   self.add_input("annual_electricity_sold", val=0.0, shape=plant_life, units="kW*h/yr") # this would have to be added as an output to the grid performance model
else:
   self.add_input("electricity_sold", val=0.0, shape=n_timesteps, units="kW")

```

h2integrate/converters/grid/test/test_grid.py
```diff
@@ -698,3 +700,274 @@ def test_zero_interconnection_costs(plant_config, n_timesteps):
...
+    electricity_out = np.full(n_timesteps, 30000.0)  # 30 MW
+    prob.set_val("grid.electricity_out", electricity_out)
+
+    prob.run_model()
+
+    dt = plant_config["plant"]["simulation"]["dt"]
+    total_energy = np.sum(electricity_out) * (dt / 3600)
+    expected_varopex = total_energy * np.array(buy_prices)
+
+    varopex = prob.get_val("grid.VarOpEx", units="USD/year")
```
elenya-grant, line 746: I think I'm concerned because the simulation is only 24 timesteps but I'm not sure if the units make sense for the VarOpEx. For an hourly simulation of 24 timesteps. The total energy is units of `kW*d` (`kW*h*24`), then multiplied by buy price cost of `USD/kW*h` does not result in units of `USD/year`. Right? I don't think the grid cost model (and perhaps the grid performance model too) is flexible to shorter simulation times. 

The VarOpEx costs in the ProFAST finance models are divided by the total production of the commodity to get the variable O&M cost value in units of `USD/commodity_amount_units` (`commodity_amount_units` is for the commodity defined for the finance model). I have to think on this a bit more though...

h2integrate/core/utilities.py
```diff
@@ -104,6 +105,63 @@ def merge_shared_inputs(config, input_type):
         return config["shared_parameters"]
 
 
+def determine_price_mode(price, n_timesteps, plant_life, price_name="price"):
```
elenya-grant: Nice way to put this as a shared utility function - I think this will be helpful for other models in the future!

elenya-grant (commented):
I gave some minor suggestions but nothing that was blocking! I did have a question on whether the grid cost model is properly set-up to handle non-yearly simulations that I'd like your take on! 

I think you added a lot of great tests for the grid and feedstock model! I'm curious why didn't you add any tests to `h2integrate/core/test/test_utilities.py` that specifically only test the `determine_price_mode` function?

h2integrate/core/utilities.py
johnjasa: Great point, I've added type hints to the function itself and added your suggestion to the docstring

h2integrate/converters/grid/grid.py
johnjasa: I like this idea and think we could generally lean into the baseclasses more. I think it would make sense as a future PR since it'd touch so many models. I've made an issue here: https://github.com/NatLabRockies/H2Integrate/issues/708

h2integrate/converters/grid/grid.py
johnjasa: Thanks for drafting up how this could look! It's slightly unclear to me what the value-add would be here. The amount of electricity bought or sold doesn't vary year-to-year, so the `annual_electricity_out` and `annual_electricity_sold` arrays would have the same value repeated across the `plant_life`. Is your suggestion more to reduce the size of the arrays used in the calculations? I'm probably missing something!

h2integrate/converters/grid/test/test_grid.py
johnjasa: You're absolutely right about the grid components not being generalized for non-yearly production. I've updated the tests here. This would require a bigger set of changes beyond just this PR and should be done as part of the non-yearly simulation change work.

h2integrate/converters/grid/grid.py
elenya-grant: The `annual_electricity_out` is flexible to varying `dt` and different simulation lengths (output units are `kW*h/yr`). `electricity_out` and `electricity_sold` are output in rate units (`kW`) and not adjusted based on time-step or simulation length. My suggestion is so that the cost model doesn't have to do the adjustments for `dt` and `n_timesteps` when it could be done by the performance model.

johnjasa: I just implemented this! We're on a call. Could you please take a look and tell me if it's close to what you want?

docs/technology_models/grid.md
```diff
@@ -49,9 +49,9 @@ Multiple grid instances may be used within the same plant to represent different
 | `interconnection_opex_per_kw`   | scalar                       | $/kW/year | Annual O&M cost per kW of interconnection.                             |
 | `fixed_interconnection_cost`    | scalar                       | $         | One-time fixed cost regardless of size.                                |
 | `electricity_out`               | array[n_timesteps]           | kW        | Electricity flowing out of grid (buying from grid).                    |
-| `electricity_buy_price`         | scalar/array[n_timesteps]    | $/kWh     | Price to buy electricity from grid (optional, time-varying supported). |
+| `electricity_buy_price`         | scalar/array[n_timesteps]/array[plant_life] | $/kWh     | Price to buy electricity from grid (optional; supports scalar, per-timestep, or per-year). |
```
vijay092: Could this be expressed as a dictionary instead? e.g. `{"dimension": plant_life, "val": [...]}`. Would remove any ambiguity for the user.

h2integrate/converters/grid/grid.py
```diff
@@ -207,28 +221,14 @@ def setup(self):
...
-                if len(buy_price) != n_timesteps:
-                    raise ValueError(
-                        f"electricity_buy_price length ({len(buy_price)}) "
-                        f"must match n_timesteps ({n_timesteps})"
-                    )
-                buy_price_shape = n_timesteps
-
-            else:
-                buy_price_shape = 1
+            self._buy_price_mode, buy_price_shape = determine_price_mode(
```
vijay092: If you had the dictionary, you probably don't need this function.

vijay092 (commented):
Looking great, John! Just a couple of thoughts:
i) A dictionary format for the sell_price array (e.g. explicitly tagging it as a yearly or daily profile) would be nice. You shouldn't have to infer the type from the array size.
ii) Maybe for later: Is there a straightforward way to pull in price forecasts automatically, to reduce the burden on the user (asking for a friend ;))

docs/technology_models/grid.md
johnjasa: You're right and I like your suggestion! I've changed it so the buy price and sell price modes are set by the user in the config file to make it clear.

h2integrate/converters/grid/grid.py
johnjasa: I've removed this in favor of your approach -- thanks for the simplification!

johnjasa:
> Looking great, John! Just a couple of thoughts: i) A dictionary format for the sell_price array (e.g. explicitly tagging it as a yearly or daily profile) would be nice. You shouldn't have to infer the type from the array size. ii) Maybe for later: Is there a straightforward way to pull in price forecasts automatically, to reduce the burden on the user (asking for a friend ;))

Great suggestions, thanks Sanjana! I've made the change you suggested so users explicitly say what the buy or sell price is. Then I've made an issue for future work to pull the price forecasts: https://github.com/NatLabRockies/H2Integrate/issues/716. I will say we haven't historically prioritized grid-integrated systems, so as you continue to think up good features to add in for your work, please bring it up or create an issue!

vijay092 (approved):
Thanks for incorporating my suggestions, John! Looks great!

h2integrate/converters/grid/test/test_grid.py
```diff
@@ -611,14 +612,15 @@ def test_time_varying_buy_price(plant_config, n_timesteps):
 
 
 @pytest.mark.unit
-@pytest.mark.parametrize("n_timesteps", [24])
+@pytest.mark.parametrize("n_timesteps", [8760])
```
jaredthomas68, line 615: I'm not sure I understand why the full 8760 is needed for this test, can you explain?

h2integrate/converters/grid/grid.py
```diff
@@ -207,28 +223,20 @@ def setup(self):
...
-                buy_price_shape = n_timesteps
-
+            self._buy_price_mode = self.config.buy_price_mode
+            if self._buy_price_mode is None:
+                # Default: scalar if single value, per_timestep if array
+                if isinstance(self.config.electricity_buy_price, list | np.ndarray):
+                    self._buy_price_mode = "per_timestep"
+                else:
+                    self._buy_price_mode = "per_timestep"
+
```
jaredthomas68, line 235: why per_timestep for always? Is this a typo?

h2integrate/converters/grid/grid.py
```diff
@@ -237,28 +245,40 @@ def setup(self):
...
+                desc="Annual electricity flowing out of grid (buying from grid)",
+            )
+        else:
+            self.add_input(
+                "electricity_out",
+                val=0.0,
+                shape=n_timesteps,
+                units="kW",
+                desc="Electricity flowing out of grid (buying from grid)",
+            )
```
jaredthomas68, line 268: Is there a case where we would want both annual for the life and per timestep?

h2integrate/converters/grid/grid.py
```diff
@@ -237,28 +245,40 @@ def setup(self):
...
-                        f"must match n_timesteps ({n_timesteps})"
-                    )
-                sell_price_shape = n_timesteps
+            self._sell_price_mode = self.config.sell_price_mode
+            if self._sell_price_mode is None:
+                if isinstance(self.config.electricity_sell_price, list | np.ndarray):
+                    self._sell_price_mode = "per_timestep"
+                else:
+                    self._sell_price_mode = "per_timestep"
+
```
jaredthomas68, line 277: typo? same result either way

jaredthomas68 (commented):
Looks pretty good, but there are a couple of typos.

h2integrate/converters/grid/test/test_grid.py
johnjasa: Great question, Elenya brought this up here: https://github.com/NatLabRockies/H2Integrate/pull/700#discussion_r3138810137

h2integrate/converters/grid/grid.py
johnjasa: Good call, fixed

h2integrate/converters/grid/grid.py
johnjasa: Good call, fixed

h2integrate/converters/grid/grid.py
johnjasa: Perhaps. For now and the projects at hand, let's keep the behavior as-is. Once we run into an analysis where someone needs this, we can come back.

PR #701, opened 2026-04-24
Rename `pipe` and `cable` transport types to `pipe_passthrough` and `cable_passthrough`

### LINKED ISSUE #560: Rename Pass-Thru Cable and Pipe
This came up in PR #555 in [this comment](https://github.com/NatLabRockies/H2Integrate/pull/555#discussion_r2866715223) which says:
> maybe change the names of `pipe` and `cable` to something that indicates their "pass-through" natures, such as `pipe-passthrough` and `cable-passthrough`

`cable` and `pipe` are "protected" technology names in H2I - but as more transport components are added, it may be good to rename the "protected" technologies to something more specific.

johnjasa:
# Rename `pipe` and `cable` transport types to `pipe_passthrough` and `cable_passthrough`

Renames the protected transport type identifiers `pipe` to `pipe_passthrough` and `cable` to `cable_passthrough` to better indicate their pass-through (no-loss) nature. This makes the naming more specific and leaves room for future transport components that model actual losses.

Changes include:
- New files `pipe_passthrough.py` / `cable_passthrough.py` with renamed classes (`PipePassthroughPerformanceModel`, `CablePassthroughPerformanceModel`)
- Updated `supported_models` registry, `reserved_techs`, and `no_cost_models`
- Updated all 34 example YAML configs, tests, documentation, and notebooks

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [ ] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues
Resolves #560


elenya-grant:
Thanks for making these changes to resolve this issue!

I hate to be this person but - I kinda don't like it (ah! even though I made the issue!)

The reason(s) I don't like it - at least don't like it coming in now (perhaps it'll make sense in the future) is that:
- typing `cable` and `pipe` are so easy compared to the `_passthrough` versions of them. I can already imagine many mini "ugh" moments in my future because I had a typo in `_passthrough` and it throws an error
- Any future transport models that include some performance aspect can be easily named something more verbose because they'd be (likely) used less often. Take for example, example `21_iron_examples`, which use the `GenericTransporterPerformanceModel` for many different commodity streams, then these technologies are named things like `crude_ore_transport` or `iron_ore_transport`. For example, if we add a lossy cable model into H2I, then it's not crazy to name that technology in the tech config as `electricity_transport` or something.
- **The use of "cable" and "pipe" ARE case-sensitive**. Aka - if someone wants to use a lossy cable performance model and they name it "Cable" in their technology config - that would be OK
- We have "protected" name(s) in the finance model definition and likely other places in H2I - I think that having the very basic cable and pipe naming is ok since we've defined it as "protected"
- I think that these should only be renamed if we find that a certain cable or pipe model is being used more often than the pass-through lossless models OR re-evaluate immediately before a major release

johnjasa:
Thanks for you comment, @elenya-grant, this is very helpful! @jaredthomas68, based on what Elenya said, what do you think? Do we need this PR or no thank you?

jaredthomas68:
As I have given this more thought and the code has continued to progress, I lean towards not renaming them. I think it may be better to gradually transition to full tech models for transport with cost and performance models. The existing pipe and cable components could have a set cost and a simple efficiency that could be set to 0 and 1, respectively, to mimic the current behavior. Eventually it would be great to have transport models that can use start and end lat lon from two site specifications and open the way for routing (via reV perhaps). I think making the existing components slightly more general will be a better step forward than giving them pass through names.

johnjasa:
Thanks @elenya-grant and @jaredthomas68! I'll close this PR as not relevant.

PR #719, opened 2026-05-01
Feature: EIA Natural Gas Price Retrieval for Feedstocks

RHammond2:
# Create an EIA Natural Gas API Feedstock Model

The newly created `EIANaturalGasFeedstockConfig` and `EIANaturalGasFeedstockCostModel` allow users to directly download or rely on previously downloaded EIA API data for natural gas prices for the US or any of its 50 states, for each of the available price categories. Users can use a single cost year's annual average or monthly results to create an hourly price profile to match the consumption model.

- [x] Feature Enhancement
  - [ ] Framework
  - [x] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] The model is functionally complete, but is awaiting any amendments based on reviewer feedback.
- [x] Run some more local tests to ensure things work as anticipated since the model relies on external data and API keys

Any feedback is welcome from anyone who might be interested in this or impacted by it

**Structural feedback**: Are there any missing considerations for when this gets scaled to a national level or for prototyping purposes?

**Implementation feedback**: Do the naming conventions make sense?

**Other feedback**: Any thoughts on integrating tests relying on local data or API keys? I wasn't sure if we crossed this bridge with H2I yet or not, so held back on it.

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues

N/A

### Section 5.1: New Files
- `h2integrate/feedstocks/eia_ng_price.py`
  - `EIANaturalGasFeedstockConfig`: EIA natural gas price configuration model that will also retrieve the data upon initialization to ensure the `price` attribute is populated for the cost model.
  - `EIANaturalGasFeedstockCostModel`: EIA cost model attuned to the specifics of the configuration model.
- `h2integrate/feedstocks/test/test_eia_ng_feedstock.py`: Adds tests for all the basic functionality, except for any step requiring access to the EIA data (downloaded or API).
- `h2integrate/preprocess/eia.py`: provides general functionality for handling EIA natural gas API queries with an eye towards future integration of other EIA API access points.
- `h2integrate/preprocess/geospatial.py`: Adds a series of US-focused reverse geocoding and state name conversion functionality with an eye towards future expansion for other geographical regions.

- `h2integrate/core/feedstocks.py`: moved to the `h2integrate/feedstocks/feedstocks.py`, and updated some of the docstrings and inline commentary as I was reviewing it.
- `h2integrate/feedstocks/__init__.py`: Adds feedstock cost and performance models for simplified imports.
- `h2integrate/core/supported_models.py`: Adds the new EIA natural gas feedstock cost model.

This addition is geared towards the MPPOC model for a national scale data center modeling project, but has been made broadly applicable by allowing all the pricing data to be accessed.

Neither tests for the execution of the API retrieval nor examples have been added for this addition because it relies on either an API key or downloaded file. Please let me know if we should still provide something more than the documentation for users.

Tests pass.

- [x] **Model Structure**:
  - [] Used `attrs` class to define the `Config` to load in attributes for the model
  - [ ] Input file comments
  - [ ] Run file comments
- [x] **Documentation**:
    - [x] Model documentation page added to the appropriate `docs/` section
    - [x] `<model_name>.md` is added to the `_toc.yml`
  - [x] Run `generate_class_hierarchy.py` to update the class hierarchy diagram in `docs/developer_guide/class_structure.md`


h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +155,56 @@ def setup(self):
...
-        self.add_output("replacement_schedule", val=0.0, shape=plant_life, units="unitless")
-
     def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
-        # Capacity factor is the total amount consumed / the total amount available
+        """Calculates the following outputs:
+
+        - ``capacity_factor``: commodity_consumed / commodity_out
+        - ``total_commodity_consumed``: sum of commodity_consumed divided by number
+          of hours simulated.
+        - ``anual_commodity_consumed``: :py:attr:`total_commodity_consumed` * (1 / years simulated)
```
cfrontin: typo: annual

h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +155,56 @@ def setup(self):
...
-        # Capacity factor is feedstock_consumed/max_feedstock_available
         self.add_output(
             "capacity_factor",
             val=0.0,
             shape=self.plant_life,
             units="unitless",
             desc="Capacity factor",
         )
+        self.add_output(
+            "replacement_schedule",
```
cfrontin, line 174: this is an overall comment that's acute here; but I thought of EIA NG almost as a resource, like wind. I think you're right that it's more of a specialized feedstock. but at the same time it comes with some funkiness, like how "replacement schedule" doesn't really make that much sense as a feedstock.

cfrontin: really this is a question for a H2I pro and not me lol

RHammond2: I originally built it as a resource system, but realized there was nowhere to model price, which is what feedstocks are used for. The replacement schedule and other such costs are more for compatibility, so it could be worthwhile that they are set to zero and not allowed to have input values because we're simply connecting to a grid for natural gas. Then again, there may be costs associated with hooking up the power plant to the pipeline, so that could be the actual intent.

h2integrate/feedstocks/feedstocks.py
RHammond2: Thanks for flagging! Resolved with commit 78e18e96

cfrontin (approved):
looks great, well done.

bayc (approved):
Looks good to me, thanks for putting this together! Definitely useful for multiple workstreams right now. As discussed in the standup, my one request was that we can support conversion of lat/long coords to the states so it is not up to the user to change both when doing runs.

h2integrate/feedstocks/feedstocks.py
johnjasa: @elenya-grant and I are on a call right now, tagging her to chime in here as well as the whole PR!

h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +155,56 @@ def setup(self):
...
+        )
 
-        # The should be equal to the commodity_capacity input of the FeedstockPerformanceModel
+        # TODO: Update to the commodity_capacity input of the FeedstockPerformanceModel
+        # NOTE: Should I set this to rated_capacity if it's available?
         self.add_output(
             f"rated_{self.config.commodity}_production",
             val=0,
             units=self.config.commodity_rate_units,
         )
```
RHammond2, line 187: @elenya-grant and @johnjasa should this also get updated per the todo/note here and in the EIA feedstock?

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+        for line in f.readlines():
+            if ":" in line:
+                name, val = line.strip().split(":")
+                if name == "EIA_API_KEY":
+                    return val.strip()
+    raise ValueError(f"No 'EIA_API_KEY' defined in {api_key_file=}")
+
+
+@define(kw_only=True)
+class EIANaturalGasFeedstockPerformanceConfig(BaseConfig):
```
elenya-grant: I don't think the performance model or the config are necessary - aren't these identical to the `FeedstockPerformanceModel`?

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+        )[["state", "price"]]
+        df = convert_to_monthly(df)
+        df.price *= MCF_to_MMBTU
+
+        if filename is not None:
+            df.to_csv(filename, index_label="period")
+        return df[["price"]]
+
+
+class EIANaturalGasFeedstockCostModel(CostModelBaseClass):
```
elenya-grant:
```suggestion
class EIANaturalGasFeedstockCostModel(FeedstockCostModel):
```

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+            merge_shared_inputs(self.options["tech_config"]["model_inputs"], "cost"),
+            # merge_shared_inputs(self.options["tech_config"]["model_inputs"], "cost") | site_config,  # noqa: E501
+            additional_cls_name=self.__class__.__name__,
+            strict=False,
+        )
+        self.n_timesteps = int(self.options["plant_config"]["plant"]["simulation"]["n_timesteps"])
+
+        super().setup()
+
+        self.dt = self.options["plant_config"]["plant"]["simulation"]["dt"]
```
elenya-grant: I think the rest of this `setup()` method could be replaced with `super().setup()` if  this class inherits the FeedstockCostModel

h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +155,56 @@ def setup(self):
...
-
-        # Capacity factor is feedstock_consumed/max_feedstock_available
         self.add_output(
             "capacity_factor",
             val=0.0,
             shape=self.plant_life,
             units="unitless",
             desc="Capacity factor",
         )
+        self.add_output(
```
elenya-grant, line 173: why was `replacement_schedule` added as an output? It shouldn't be needed.

h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +155,56 @@ def setup(self):
...
-
-        # Sum the amount consumed
         outputs[f"total_{self.config.commodity}_consumed"] = inputs[
             f"{self.config.commodity}_consumed"
         ].sum() * (self.dt / 3600)
 
-        # Estimate annual consumption based on consumption over the simulation
-        # NOTE: once we standardize feedstock consumption outputs in models, this should
-        # be updated to handle consumption that varies over years of operation
+        # TODO: update to handle varying consumption levels when feedstock consumption is available
```
elenya-grant: I'm curious why this comment was changed? I thin the previous NOTE was a bit more clear to me

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+    Args:
+        state (str): Full state name in title casing or two-letter state abbreviation in upper case.
+
+    Returns:
+        str: Two-letter state abbreviation.
+    """
+    return STATE_MAP.get(state, state)
+
+
+def get_eia_api_key(api_key_file: Path | None) -> str:
```
elenya-grant: theres a lot of methods in `h2integrate/resource/utilities/nlr_developer_api_keys.py` that can handle different ways of setting API keys that may add flexibility to this.

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+    state: str = field(
+        default=None,
+        converter=attrs.converters.optional(
+            attrs.converters.pipe(convert_state_value, convert_state_to_code)
+        ),
+        validator=attrs.validators.optional(
+            attrs.validators.in_([*STATE_MAP, *STATE_MAP.values()])
+        ),
+    )
+    latitude: float = field(default=999.9, validator=attrs.validators.instance_of(float))
```
elenya-grant: I think latitude and longitude should look like this:
```python
    latitude: float = field(default=0.0, validator=range_val(-90.0, 90.0))
    longitude: float = field(default=0.0, validator=range_val(-180.0, 180.0))
```

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+    longitude: float = field(default=999.9, validator=attrs.validators.instance_of(float))
+    cost_year: int = field(default=CURRENT_YEAR)
+    annual_cost: float = field(default=0.0, converter=float)
+    start_up_cost: float = field(default=0.0, converter=float)
+    commodity: str = field(default="natural_gas", init=False)
+    commodity_rate_units: str = field(default="MMBtu/h", init=False)
+    commodity_amount_units: str = field(default="MMBtu", init=False)
+    filename: str = field(default=None)
```
elenya-grant: I got an `AttributeError` when the filename wasn't specified because `Path(None)` doesnt work. I think the default should be an empty string.

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+
+    def compute(self, inputs, outputs):
+        """Generates the feedstock array operating at full capacity for a full year."""
+        outputs[f"{self.config.commodity}_out"] = np.full(
+            self.n_timesteps, inputs[f"{self.config.commodity}_capacity"][0]
+        )
+
+
+@define
+class EIANaturalGasFeedstockConfig(BaseConfig):
```
elenya-grant: I think this should inherit the `FeedstockCostConfig`

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+
+            self.state = get_state_from_coords(self.latitude, self.longitude)
+
+        self.series = EIA_FACET[self.price_category].format(self.state)
+        if self.commodity_amount_units is None:
+            self.commodity_amount_units = f"({self.commodity_rate_units})*h"
+        self.url = self.create_eia_api_url()
+        self.price = self.get_data()
+
+    def create_eia_api_url(self):
```
elenya-grant: I think `create_eia_api_url()` and `get_data()` should be methods in `EIANaturalGasFeedstockCostModel` instead of the config class.

h2integrate/feedstocks/eia_ng_pricing.py
```diff
@@ -0,0 +1,554 @@
...
+
+        # TODO: Update to the commodity_capacity input of the FeedstockPerformanceModel
+        # NOTE: Should I set this to rated_capacity if it's available?
+        self.add_output(
+            f"rated_{self.config.commodity}_production",
+            val=0,
+            units=self.config.commodity_rate_units,
+        )
+
+    def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
```
elenya-grant: If this class inherits the FeedstockCostModel - then it would only need to have

```python
super().compute(inputs, outputs)
```

elenya-grant (changes requested):
Hi Rob! I think this is cool functionality - thanks for adding it in! A few higher-level thoughts:

1. I wonder if the API download and grabbing data should instead be a pre-processing method (since, as-is, the price data wouldn't change in a site-sweep like what's done in Example 22)
2. To get the price data to change per site (if lat/lon are being swept) would require the `latitude` and `longitude` to be inputs to this method. This would also require a lot of the functionality in the cost config class to be called in the `compute()` method. 
3. The H2IntegrateModel has some special handling for feedstocks - I think that if we want to make this a feedstock cost model then it should inherit the FeedstockCostModel and the logic within H2I should be updated to check if a model is or inherits the FeedstockCostModel rather than checking that the name of the model is FeedstockCostModel.

I am curious what the main purpose of this feedstock model is perhaps. Is the goal to get data downloaded once OR to make it so that the cost is updated in a sweep? My current suggestion is that someone wants to do a design sweep of site locations (like example 22) and have the natural gas price change per site, that would instead look like:
- use the EIA API calls and functions that you added here as functions available in the `preprocess` folder instead. A user can make a run-script that loops through lat/lons and get the natural gas price using those methods/functions. User basically makes a data frame with columns of site.latitude, site.longitude, and natural_gas_feedstock.price. 
- Save the data frame from step 1 as a csv. Run a DOE like what's done in Example 22 (but with natural_gas_feedstock.price as a design variable too). The feedstock model would still be using the FeedstockCostModel.

This suggestion would require no changes to H2IntegrateModel, or feedstocks.py. I think that integrating the feedstocks with the site information should perhaps be targeted in future work since, and I think the above suggestion would work well enough in the meantime.

I'd be curious what folks thoughts are on the proposed approach and I'm sorry to propose such a big change! I honestly hope that my suggestion would be the easiest to get this functionality in sooner rather than later. 

Please let me know if you want to talk through it at all! Happy to discuss over a call!

h2integrate/feedstocks/feedstocks.py
RHammond2: It was already there (https://github.com/NatLabRockies/H2Integrate/tree/develop/h2integrate/core/feedstocks.py#L183), I just moved the inline comment to the output's description.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: I originally used, this, but needed a value to trigger that no location data was provided. I think that we could simply riff on the original by using:

```python
    latitude: float | None = field(default=None, validator=attrs.validators.optional(range_val(-90.0, 90.0)))
    longitude: float | None = field(default=None, validator=attrs.validators.optional(range_val(-180.0, 180.0)))
```

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: This was actually an area I forgot I wanted feedback on because my main hangup with using the existing methods is that it was all built around the NLR API infrastructure and naming. I have thoughts about how we could improve that model to make it more general while still having all the NLR specifics, but that felt well out of scope for both the project and this PR. Basically, because this code doesn't use NLR naming conventions or APIs, it didn't feel like it fit well in terms of specifics or themes.

h2integrate/feedstocks/feedstocks.py
RHammond2: I converted from a note to a todo because of "this should be updated". Happy to revert it to its longer form, but I would still strongly argue this is a TODO for when the variable consumption is implemented.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Thanks for flagging, I think I fixed this (among other things) in some commits after you would've pulled this down. I also implemented a bunch of tests around much of the configuration class and functions.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: It originally did, but I need a couple pieces of the `FeedstockCostConfig.setup()` prior to being able to call setup, otherwise the price data won't be available. One remedy, and could work well for the next comment is to just set a default value, and get the actual data at the end of the setup. This could be ok because the price is automatically converted to monthly anyway for simplified handling of converting it to an 8760.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Similar to the previous comment, I originally did that as well, but the price data was required for determining the annual vs per timestep logic. Similar to the above, I think that if I use a dummy value for the price data prior to retrieval, then this could work even if the logic will be a bit out of order.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Same commentary as the above couple of responses.

h2integrate/feedstocks/eia_ng_pricing.py
elenya-grant: understood and heard! I agree and could work on making a separate PR for those environment variable methods to be generalizable

h2integrate/feedstocks/eia_ng_pricing.py
elenya-grant: yeah! I didn't even know about the optional attrs validators! Those are slick!

RHammond2:
Thanks for the feedback,  @elenya-grant! I think pretty much everything you're saying makes good sense to me. My own concern is that this isn't inherently built around multi-site analyses, so there is still some use in having a single site version be available. Otherwise, I agree that much of the actual API functionality could become preprocessing code that a single site or generic feedstock can use. I've also scheduled a meeting so we can iterate on these ideas a bit more effectively.

RHammond2:
The path forward based on a call with @elenya-grant is to make the following amendments:

1. Move all the API-specific functionality to a new preprocessing module
2. Expand the API to allow for downloading of multiple years and sites
3. Only use a specialized cost model and configuration that inherits the main feedstock model in place of an entirely separate model for each component.

h2integrate/core/h2integrate_model.py
elenya-grant, line 1: Thanks for updating this!

h2integrate/feedstocks/feedstocks.py
elenya-grant: I think this is fine as-is! It is set in the `compute()` method

h2integrate/feedstocks/feedstocks.py
```diff
@@ -155,48 +157,57 @@ def setup(self):
...
-
     def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
-        # Capacity factor is the total amount consumed / the total amount available
+        """Calculates the following outputs:
+
+        - ``capacity_factor``: commodity_consumed / commodity_out
+        - ``total_commodity_consumed``: sum of commodity_consumed divided by number
+          of hours simulated.
+        - ``annual_commodity_consumed``: :py:attr:`total_commodity_consumed` * (1 / years simulated)
+        - ``rated_commodity_production``: maximum input ``commodity_out``.
```
elenya-grant, line 195: thanks for adding a doc string!

h2integrate/feedstocks/eia_ng_price.py
```diff
@@ -0,0 +1,149 @@
...
+    ``longitude`` from the site configuration, otherwise the ``state`` will need to be provided
+    in the configuration.
+    """
+
+    def setup(self):
+        """Defines the inputs and outputs of the model and converts the
+        :py:attr:`EIANaturalGasFeedstockConfig.price` to an hourly timeseries for the
+        ``plant_life``.
+        """
+        site_config = self.options["plant_config"].get("site", {})
```
elenya-grant:
```suggestion
        site_config = self.options["plant_config"].get("sites", {})
```

The site portion of a plant config is formatted as:

```yaml
sites:
   wind_site:
      latitude:
      longitude:
   other_place:
      latitude:
      longitude:
```

So - It may be better to loop through the dictionaries under `sites` and see if any of them contain `latitude` or `longitude`. Or - it may be batter/easier to require a user to provide the name of the site to use if they don't provide a state or latitude or longitude. This could look like adding a config attribute of `site_name` and they could specify either `wind_site` or `other_place`.

h2integrate/preprocess/geospatial.py
elenya-grant, line 1: I wonder if this file should instead be named line `geospatial_us` or something to reflect that the functions within it are specific to the United States and not other countries.

h2integrate/feedstocks/test/test_eia_ng_feedstock.py
elenya-grant, line 1: would love to see some testing on `EIANaturalGasFeedstockCostModel`.

elenya-grant (commented):
Thanks for making these changes! I had a few suggestions to using `plant_config['sites']` because each key under `sites` is a user-defined site name. I think it'd be good to test the feedstock cost model using a plant config from one of the examples!

h2integrate/feedstocks/eia_ng_price.py
RHammond2: I'll add `site_name` as an input, otherwise I'll grab the first instance to avoid erroring out unnecessarily.

elenya-grant:
One other thought - since `price` is an openmdao input for the feedstock cost model, we should maybe throw a user warning if they change it.

I'm thinking like the following thing in the `compute()` statement
```python
def compute(self, inputs, outputs):
   if not np.isclose(inputs["price"], self.config.price, rtol=1e-6):
      warnings.warn("price has changed from EIA price", UserWarnings)
   super().compute(inputs, outputs)

```

h2integrate/preprocess/geospatial.py
RHammond2: I thought about this, but figured that for now there isn't a use case, so it was best not to overspecify things. If it's helpful, I could go through and make sure that the naming conventions on the functionality I've created isn't too broad.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Removed this class with the suggested refactor.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Resolved with the refactor.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Resolved with the refactor.

h2integrate/feedstocks/feedstocks.py
RHammond2: Original commentary was reinstated in a todo.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Resolved with refactor.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Suggestion incorporated with refactor.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Suggestion incorporated with refactor.

h2integrate/feedstocks/eia_ng_pricing.py
RHammond2: Suggestion incorporated with refactor.

h2integrate/feedstocks/eia_ng_price.py
RHammond2: Added in 38250e1d and 2577afd2

RHammond2:
> One other thought - since `price` is an openmdao input for the feedstock cost model, we should maybe throw a user warning if they change it.

Should this be a warning or an error? In theory this class should be relying on the EIA natural gas pricing data either through an API query or loading existing data, so it would seem counterproductive to override it.

johnjasa:
> > One other thought - since `price` is an openmdao input for the feedstock cost model, we should maybe throw a user warning if they change it.
> 
> Should this be a warning or an error? In theory this class should be relying on the EIA natural gas pricing data either through an API query or loading existing data, so it would seem counterproductive to override it.

I think it should be a warning if we do anything here. The use case would be if somebody has the EIA nat gas price coming in, then maybe does a parameter sweep on price with the same base config files, we'd want to warn them. I don't think the warning is _strictly_ necessary but I do think Elenya's suggestion is a user-friendly reminder, so probably worth adding imo.

RHammond2:
> > > One other thought - since `price` is an openmdao input for the feedstock cost model, we should maybe throw a user warning if they change it.
> > 
> > 
> > Should this be a warning or an error? In theory this class should be relying on the EIA natural gas pricing data either through an API query or loading existing data, so it would seem counterproductive to override it.
> 
> I think it should be a warning if we do anything here. The use case would be if somebody has the EIA nat gas price coming in, then maybe does a parameter sweep on price with the same base config files, we'd want to warn them. I don't think the warning is _strictly_ necessary but I do think Elenya's suggestion is a user-friendly reminder, so probably worth adding imo.

Thanks for the explanation, I'll add it as a warning.

h2integrate/feedstocks/test/test_eia_ng_feedstock.py
RHammond2: Added with 9dbd7d58

h2integrate/preprocess/geospatial.py
elenya-grant: thats a good point! I am good with leaving it as-is for now!

elenya-grant (approved):
Thanks for making these changes! Looks good to me! Very cool new functionality! Thanks for adding this in!

h2integrate/feedstocks/test/test_eia_ng_feedstock.py
```diff
@@ -0,0 +1,208 @@
...
+    CURRENT_YEAR,
+    EIANaturalGasFeedstockConfig,
+    EIANaturalGasFeedstockCostModel,
+)
+
+
+DUMMY_KEY = "xxxxxx"
+RG_NOT_INSTALLED = importlib.util.find_spec("reverse_geocoder") is None
+
+best_trailer_in_colorado_coords = (39.9140081, -105.2249155)
```
johnjasa, line 21: absolutely beautiful

johnjasa (approved):
Thanks for this, Rob! I briefly tried it locally and appreciate the added capabilities. I like that it's either a preprocessing step or a feedstock model; very flexible. Hope this is useful for your work, I know it will be for others!

h2integrate/feedstocks/test/test_eia_ng_feedstock.py
RHammond2: Glad you appreciated the easter egg!

Issue #724, opened 2026-05-04
Use time-series results for finance calcs


elenya-grant:
# Use time-series results for finance calcs

Before a lot of major development went into the finance models and performance models, the finance groups used to use the "commodity_out" from the technology specified as the commodity stream. ProFAST models used to set the `utilization` to 1 and set the rated capacity based on the `commodity_out` profile. Then, the ProFAST models were updated to instead take in `capacity_factor` (as the utilization) and `rated_commodity_production` (as the rated capacity). There may be cases where someone wants to calculate the finances for a technology but not based on its capacity factor or commodity production. For example, if someone wanted to use a splitter as the commodity stream or only wanted to use the electricity bought from the grid in the calculation. 

## Proposed solution

Add a new finance component, similar to `AdjustedCapExOpexComp`, that can handle this case and interface with the finance models as expected. The primary changes would be to H2IntegrateModel and the addition of this new component. For example:
```python
class CommodityStreamSpecifiedComp(om.ExplicitComponent):
   def setup(self):
      self.add_input(f"{commodity}_out", val=0.0, shape=self.n_timesteps, units=commodity_rate_units)
      self.add_output("capacity_factor", units="unitless", shape=self.plant_life)
      self.add_output(f"rated_{commodity}_production", units=commodity_rate_units)

```

And this component would be added to the finance subgroup internal to h2i. I will draft up a PR to show-case this.

elenya-grant:
Resolved with PR #725

PR #725, opened 2026-05-04
Ability to use timeseries profiles for finance calcs

elenya-grant:
# Ability to use timeseries profiles for finance calcs

This PR attempts to resolve Issue #724.

The basis of this PR is that sometimes a user may want to calculate the finances using a timeseries profile. For example, it was not previously possible to calculate the LCOE using the `electricity_out1` profile output from a splitter. This PR introduces the ability to use timeseries profiles with the finance models. This is done through an additional finance subsystem in the finance subgroup that is used upstream of the finance model. This new component is called `AdjustedCapacityFactorComp`. The user can opt to use one this functionality in a finance subgroup by specifying two variables, `use_commodity_stream_timeseries` and `commodity_stream_output` as shown below:
```yaml
  finance_subgroups:
    electricity_doc:
      commodity: electricity
      commodity_stream: electricity_splitter # name of technology that commodity is output from
      use_commodity_stream_timeseries: true # new variable to flag to use time-series profile
      commodity_stream_output: electricity_out1 # the output of `commodity_stream` tech
      technologies: [wind, doc]
```

This functionality is showcased in Example 17. Specific feedback for reviewers is outlined in Section 2.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [x] Other (please describe): added flexibility for commodity streams of finance models.
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Add unit and regression tests for `AdjustedCapacityFactorComp`
- [x] Add integration test to ensure that error is thrown if `use_commodity_stream_timeseries` is True but `commodity_stream_output` is not input.
- [x] Update finance documentation

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:
- Thoughts on variable names (`use_commodity_stream_timeseries` and `commodity_stream_output`)? <s>Should `commodity_stream_name` be `commodity_stream_output_name`?</s> I intentionally included `use_commodity_stream_timeseries` (although a lot of the logic in H2I could be handled with just `commodity_stream_output`) because I wanted users to have to be very clear with what they're doing if they want to use this option.
- Thoughts on the name of the class `AdjustedCapacityFactorComp`? Totally open to other naming suggestions.

- [x] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

This would resolve issue #724

### Section 5.1: New Files
- `h2integrate/finances/test/test_finance_components.py`: tests for `AdjustedCapacityFactorComp`

  - `create_finance_model()`: Updated to see if `AdjustedCapacityFactorComp` should be added to the finance subgroup.
  - `connect_technologies()`: Updated to handle connections to finance subgroup based on the commodity stream and whether to use the timeseries or the capacity factor.
- `h2integrate/finances/finances.py`: added new component `AdjustedCapacityFactorComp`
- `examples/17_splitter_wind_doc_h2/plant_config.yaml`: Added two finance subgroups to highlight this functionality.
- `pytest examples/test/test_all_examples.py::test_splitter_wind_doc_h2_example`: added subtests for the new finance subgroups, functions as an integration test.
- `h2integrate/core/test/test_framework.py`: added new test `test_use_commodity_stream_timeseries_finances_error`
- `docs/user_guide/specifying_finance_parameters.md`: updated documentation


h2integrate/finances/finances.py
```diff
@@ -85,3 +85,43 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
         outputs["total_capex_adjusted"] = total_capex_adjusted
         outputs["total_opex_adjusted"] = total_opex_adjusted
         outputs["total_varopex_adjusted"] = total_varopex_adjusted
+
+
+class AdjustedCapacityFactorComp(om.ExplicitComponent):
```
jaredthomas68, line 90: doc string

h2integrate/core/h2integrate_model.py
```diff
@@ -1002,6 +1007,21 @@ def create_finance_model(self):
                         # uniquely named outputs
                         commodity_output_desc = commodity_output_desc + f"_{finance_group_name}"
 
+                if finance_subgroups[subgroup_name]["use_commodity_stream_timeseries"]:
```
jaredthomas68, line 1011: nice error message!

jaredthomas68 (changes requested):
I like this, thanks for putting it together so quickly.

h2integrate/core/h2integrate_model.py
```diff
@@ -798,7 +798,7 @@ def create_finance_model(self):
             )
             tech_names = subgroup_params.get("technologies")
             commodity_stream = subgroup_params.get("commodity_stream", None)
-
+            (False if subgroup_params.get("commodity_stream_name", None) is None else True)
```
johnjasa: this line isn't used for anything? might've been debugging -- delete?

johnjasa:
```suggestion
```

johnjasa (approved):
Thanks for your quick and helpful response to @cfrontin's request, @elenya-grant! Really appreciate it. This is good to go from my end, I've just re-requested @jaredthomas68's review. Thanks again!

kbrunik:
I like the name `use_commodity_stream_timeseries`. I'd prefer `commodity_stream_output` to `commodity_stream_name` in the finance subgroup. 

Could you also add a test that ensures that `use_commodity_stream_timeseries` works for a converter `commodity_stream_output` as well? I'd assume that for the PEM it would make the LCOH be different values. And if we were really trying to dot our i's and cross our t's we'd add a test one for the LCOE from the wind and that it's the same value.

h2integrate/finances/finances.py
elenya-grant: done!

elenya-grant:
> I like the name `use_commodity_stream_timeseries`. I'd prefer `commodity_stream_output` to `commodity_stream_name` in the finance subgroup.
> 
> Could you also add a test that ensures that `use_commodity_stream_timeseries` works for a converter `commodity_stream_output` as well? I'd assume that for the PEM it would make the LCOH be different values. And if we were really trying to dot our i's and cross our t's we'd add a test one for the LCOE from the wind and that it's the same value.

@kbrunik - I updated the name to `commodity_stream_output`. Thanks for responding to that requested feedback! I also added your suggested tests into the `examples/test/test_all_examples.py::test_splitter_wind_doc_h2_example`, this test now tests that:
- LCOH using `hydrogen_out` from the electrolyzer is less than LCOH using the capacity factor (which reflects lifetime performance)
- LCOE using `electricity_out` from the wind is equal to the LCOE using the capacity factor (since the wind component does not reflect lifetime performance)

h2integrate/core/test/test_framework.py
```diff
@@ -14,6 +14,35 @@
...
+        "driver_config": driver_config,
+    }
+
+    with subtests.test("Commodity stream name is missing"):
+        with pytest.raises(ValueError) as excinfo:
+            H2IntegrateModel(top_level_config)
+        err = str(excinfo.value)
+        assert "`commodity_stream_output` is a required input" in err
+        assert "`use_commodity_stream_timeseries` is True" in err
+        assert "finance subgroup `electricity_doc`" in err
```
jaredthomas68, line 47: I'm not sure if these should be collapsed into a single assert or protected in three subtests.

jaredthomas68 (changes requested):
One minor test adjustment needed.

The following is non-blocking:

I also think shifting to a nested approach for the "use_commodity_stream_timeseries" etc could help extend this capability to allow for degradation adjustment over time:

```yaml
use_commodity_stream_timeseries:
    commodity_stream_output: "name" # should default to None to make the boolean unnecessary
    commodity_stream_degradation: 0.05 # decrease by 5% each year - note that I'm not set on naming or usage, just some thoughts.
```

This would also make the boolean unnecessary.

h2integrate/core/test/test_framework.py
elenya-grant: updated to be 3 separate subtests.

jaredthomas68 (approved):
Looks good. Thanks

Issue #748, opened 2026-05-11
SLC: User-specified custom system level controller


elenya-grant:
# SLC: User-specified custom system level controller

currently, users can specify a custom resource model, technology model, or system-level finance model. In the future, user's should be able to have their own/custom technology-specific control strategy and their own/custom system level controller model. 

For a user to add a custom technology-specific control strategy, this can be done simply in H2IntegrateModel with the following change in `collect_custom_models()` from this:

```python
self.create_custom_models(
            self.technology_config["technologies"],
            self.tech_parent_path,
            ["performance_model", "cost_model", "finance_model"],
        )
```

to this:

```python
self.create_custom_models(
            self.technology_config["technologies"],
            self.tech_parent_path,
            ["performance_model", "cost_model", "finance_model", "control_strategy"],
        )
```

A more complex change would be needed for a user to specify a custom **system level control** strategy.

Issue #747, opened 2026-05-11
SLC: Add logic in for connecting/handling splitters


elenya-grant:
# SLC: Add logic in for connecting/handling splitters

Currently, splitters are not connected to the SLC. These would likely be very important control points in the system and would perhaps require some special handling and connection to the SLC and the components that are connected to it. We should think about adding in connections to the SLC for splitters and logic in the SLC baseclass to handle splitters.

elenya-grant:
Thoughts on this: to keep H2I flexible to the broad set of potential use-cases, user's should be able to specify which (if any) splitters can be controlled by the SLC vs which ones the SLC should not have access to change. Aka - if a user should be able to specify which splitters the SLC can change the inputs for vs which ones it can't.

PR #746, opened 2026-05-11
Bugfix on electrolyzer outputs if no hydrogen is produced

elenya-grant:
# Bugfix on electrolyzer outputs if no hydrogen is produced

Updates to some electrolyzer outputs in case the electrolyzer is never turned on. 

If the electrolyzer was never turned on, the `efficiency` output was `np.nan`. This has been updated so that the efficiency output is zero if the efficiency value in `H2_Results` is nan.

The `refurb_period` (used to calculate the `replacement_schedule` output) was set to a hard-coded 80,000 value if the Time Between Replacement output from H2_Results was nan. Now it's been updated to be whatever was set in the config as `uptime_hours_until_eol`. The `time_until_replacement` output has been updated similarly (before it was always output to the value from H2_Results, regardless of it was nan or not).

## Section 1: Type of Contribution

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [x] Fill out TODO list steps
- [x] Describe requested feedback from reviewers on draft PR
- [x] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [ ] Step 1
- [ ] Step 2

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Added tests for new functionality or bug fixes
- [x] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [x] Examples have been updated (if applicable)
- [ ] `CHANGELOG.md`
  - [ ] At least one complete sentence has been provided to describe the changes made in this PR
  - [ ] After the above, a hyperlink has been provided to the PR using the following format:
    "A complete thought. [PR XYZ]((https://github.com/NatLabRockies/H2Integrate/pull/XYZ)", where
    `XYZ` should be replaced with the actual number.

### Section 5.2: Modified Files
[...]

Issue #741, opened 2026-05-08
SLC: Add methods to the `SystemLevelControlBase` to improve handling of multi-commodity systems


elenya-grant:
# SLC: Add methods to the `SystemLevelControlBase` to improve handling of multi-commodity systems 

The current methods in the baseclass can be useful for getting important information about multi-commodity systems:
- `_get_commodity_for_tech()`: return a list of the commodities output from a given tech (based on the technology graph)
- `get_upstream_techs_for_commodity()`: returns a list of technologies upstream of a given tech that output the specified commodity
   - **this should be updated to only return technologies that have a path to the given tech**
- `find_converter_techs()`: returns a set of tuples formatted as `(input_commodity, tech_name, output_commodity)` for only technologies that have a different output commodity than the input commodity.

But perhaps one of the most useful methods (which may require other methods within in) would be a method that can return the aggregated conversion ratio of each energy flow stream to the demand technology. 

If we use `examples/35_system_level_control/yes_hydrogen` to highlight this:

```yaml
technology_interconnections:
  - [ng_feedstock, natural_gas_plant, natural_gas, pipe]
  # connect NG feedstock to NG plant
  - [wind, battery, electricity, cable]
  # wind output available for battery charging (electricity_in)
  - [wind, elec_combiner, electricity, cable]
  # wind to combined output
  - [battery, elec_combiner, electricity, cable]
  # battery net output to combined output
  - [natural_gas_plant, elec_combiner, electricity, cable]
  # electricity to electrolyzer
  - [elec_combiner, electrolyzer, electricity, cable]
  # electrolyzer to storage
  - [electrolyzer, h2_storage, hydrogen, pipe]
  # combine hydrogen streams
  - [electrolyzer, h2_combiner, hydrogen, pipe]
  - [h2_storage, h2_combiner, hydrogen, pipe]
  # send hydrogen to load
  - [h2_combiner, h2_load_demand, hydrogen, pipe]
  # combined supply to demand
```

It'd be nice to have a method that can get us the conversion ratio of electricity to hydrogen (`wind + battery + natural_gas_plant` electricity production divided by the electrolyzer hydrogen production), but also the conversion ratio of natural gas to electricity or natural gas to hydrogen. This method should be made to account for more complex cases, such as:
- natural_gas_plant actually used to provide electricity to an ammonia plant instead of the electrolyzer
- hydrogen from the electrolyzer is then converted back into electricity using a fuel cell
[...]

PR #751, opened 2026-05-12
System level control framework

elenya-grant:
# System level control framework

**This work was also done by @kbrunik and @johnjasa :)**

Introducing the initial framework for system level control. This PR introduces 3 system level controllers, which are simple controllers. The goal of this PR is to introduce the framework for system level control - which will be further developed in future PRs that address the issues listed in Section 4.

The three system level controllers are:
- `DemandFollowingControl`
- `CostMinimizationControl`
- `ProfitMaximizationControl`

Many of the file changes in this PR are related to adding control classifiers for each technology performance model.

PR's that were pre-requisites to this are:
- PR #664 
- PR #746 

The information that the system level controller is input for each technology is:
- the rate

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [ ] Open draft PR

### TODO:
- [ ] Step 1
- [ ] Step 2

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [ ] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

**Issues related to this PR**
- Discussing technology classifications [Issue #687](https://github.com/NatLabRockies/H2Integrate/issues/687)

**Issues for planned follow-on work**
- Require user to specify demand component in control_parameters ([Issue #739](https://github.com/NatLabRockies/H2Integrate/issues/739))​
- Only connect technologies upstream of the demand technology to the SLC ([Issue #740](https://github.com/NatLabRockies/H2Integrate/issues/740))​
- Add methods to baseclass to improve handling of multi-commodity systems ([Issue #741](https://github.com/NatLabRockies/H2Integrate/issues/741))​
- Handling of splitters ([Issue #747](https://github.com/NatLabRockies/H2Integrate/issues/747))​
- Custom control strategies ([Issue #748](https://github.com/NatLabRockies/H2Integrate/issues/748))​

### Section 5.1: New Files
- `h2integrate/control/control_strategies/system_level/solver_options.py`
  - `SLCSolverOptionsConfig`: configuration class for the non-linear solver options to specify for the system level controller. This is used to avoid usage of hard-coded defaults in `H2IntegrateModel`.
- `h2integrate/control/control_strategies/system_level/system_level_control_base.py`
  - `SystemLevelControlBase`: baseclass for system level controllers. Includes methods to setup inputs and outputs and methods that may be useful for the controllers. The `setup()` method ensures that all commodity streams are defined with consistent units (aka - all hydrogen commodity streams are defined in kg/h and all electricity commodity streams are defined in kW, etc)
- `h2integrate/control/control_strategies/system_level/cost_minimization_control.py`
  - `CostMinimizationControl`: 
- `h2integrate/control/control_strategies/system_level/demand_following_control.py`
  - `DemandFollowingControl`: 
- `h2integrate/control/control_strategies/system_level/profit_maximization_control.py`
  - `ProfitMaximizationControl`: 
  - `ProfitMaximizationControlConfig`
  
### Section 5.2: Modified Files (of importance)
  - `__init__()`: Added attribute `slc` which is used to decide whether to call `_classify_slc_technologies()` and `add_system_level_controller`.
  - `_classify_slc_technologies()`: new method that creates the `slc_config` dictionary used in the system level controller base class and is an input to `add_system_level_controller()`
  - `add_system_level_controller()`: new method that creates the system level control model and connects the corresponding inputs and outputs between technologies and the system level controller
  - `create_technology_models()`: updated to populate the attribute `self.tech_control_classifiers` which is a dictionary with keys as the technology names and values of the corresponding technology control classifier.
- `h2integrate/core/supported_models.py`: added `strategy_map` dictionary which is the equivalent of `supported_models` but specifically for system level controllers
- `h2integrate/core/model_baseclasses.py`
   - `PerformanceModelBaseClass`
      - `apply_curtailment()`: new method that applies curtailment to the output of any technology that is classified as "flexible". This logic only happens if "system_level_control" is found as a top-level key in the plant configuration file.
      - `setup()`: checks if control classifier is flexible and if so, adds input of `f"{self.commodity}_set_point"` and output of `f"uncurtailed_{self.commodity}_out"`
   - `CostModelBaseClass`
      - `setup()`: added output of "marginal_cost". 
- `h2integrate/finances/profast_lco.py`
   - `ProFastLCO.compute()`: updated to not run profast if the capacity factor is zero and if `system_level_control` is found in the plant config. Warns the user that its setting the LCO output to a large number but that it was not actually calculated if the the capacity factor is zero and `system_level_control` is found in the plant config

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+:width: 70%
+:align: center
+```
+
+## Storage
+Storage is a unique control classifier because it assumes that within the model that energy isn't created or destroyed (minus some efficiency losses). While it's technically "dispatchable" in that it can receive and change its performance based on a set point it's handling within H2I is unique because it's attached to storage performance models, which is handled differently than converter performance models. A converter model only has positive (or zero) `{commodity}_out`, whereas a storage model can have positive or negative `{commodity}_out`.
+
+There are two types of cases for the storage control classifier:
+1. **with a storage controller**
+When the storage performance model is controlled with a storage-level controller (open-loop or feedback controlled), the system-level controller outputs combined demand, that is always positive to the storage-level controller. The demand is `{commodity}_in` from the technologies upstream of the storage that output the same commodity to the storage performance model and the `remaining_demand`.
```
jaredthomas68, line 68: As discussed, I think the SLC should always communicate set point to the storage controller and the current storage control methods should likely be moved to SLCs, I think this is being worked on.

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,72 @@
...
+# Cost Minimization System Level Controller
+
+The cost minimization controller, `CostMinimizationControl`, meets demand at minimum variable cost using merit-order dispatch.
+Unlike the {ref}`demand following controller <slc-demand-following>`, which splits demand evenly across dispatchable technologies, this controller dispatches the cheapest technologies first.
+
+## Dispatch Logic
+
+The controller follows a three-step dispatch process:
+
+1. **Curtailable technologies** run at their full rated capacity (assumed zero marginal cost). Their output is subtracted from the demand.
```
jaredthomas68: Maybe available capacity instead of rated?

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,72 @@
...
+2. **Storage technologies** absorb any surplus (charging) or provide the deficit (discharging). Residual demand is split evenly across storage technologies producing the demanded commodity.
+3. **Dispatchable technologies** are dispatched by cheapest marginal cost first, each up to its rated capacity, until the remaining demand is met.
+
+## Marginal Cost Configuration
+
+Marginal costs are specified per dispatchable technology in the `cost_per_tech` dictionary under `system_level_control.control_parameters` in the plant config. Each entry can be:
+
+| Value | Description |
+| --- | --- |
+| Numeric (e.g. `0.05`) | Constant marginal cost in `$/(commodity_unit*h)` |
```
jaredthomas68: should this be either `$/(commodity_rate_unit*h)` or `$/commodity_amount_unit`? I don't think `$/(commodity_unit*h)` is correct. I think users will expect it to be `$/commodity_amount_unit`

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,72 @@
...
+| `"feedstock"` | Sums upstream feedstock `VarOpEx` values and divides by the technology's total production |
+
+```{note}
+The dispatch order is determined by sorting dispatchable technologies by their **mean** marginal cost across all timesteps (cheapest first).
+```
+
+### Example Configuration
+
+```yaml
+system_level_control:
```
jaredthomas68, line 46: I'm wondering if the system level control should be called "dispatch" and the local control called "control", where actual control decisions are made. I'm not sure we won't have true control at the system level or dispatch at the tech level, so this is just a thought to gather comments from others. It would simplify naming conventions: dispatch making set point decisions, control making path decisions to achieve those set points.

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,72 @@
...
+
+- `f"{tech_name}_{tech_output_commodity}_out"`
+- `f"{tech_name}_rated_{tech_output_commodity}_production"`
+
+The outputs for `curtailable`, `dispatchable`, or `storage` technologies *without* a storage controller are:
+- `f"{tech_name}_{tech_output_commodity}_set_point"`
+
+The outputs for `storage` technologies *with* a storage controller are:
+- `f"{tech_name}_{tech_output_commodity}_demand"`
+
```
jaredthomas68: As noted previously, I think we should avoid this distinction if it is reasonable and have all techs accept set_point into their local control (which would be required)

docs/control/system_level_control/slc_demand_following.md
```diff
@@ -0,0 +1,108 @@
...
+    )
+)
+```
+## Dispatch Logic
+
+The demand is satisfied in a fixed three-step priority order, and each step's shortfall or surplus is passed to the next:
+
+1. **Curtailable techs** run at their full rated capacity. Their total output is subtracted from the demand, which may drive the residual demand negative (surplus).
+
+2. **Storage techs** receive the residual demand (which may be positive or negative). When demand is positive the storage is commanded to discharge; when negative it is commanded to charge. If multiple storage techs produce the demanded commodity, the residual demand is
```
jaredthomas68:
```suggestion
2. **Storage techs** receive the residual demand (which may be positive or negative). When residual demand is positive the storage is commanded to discharge; when negative it is commanded to charge. If multiple storage techs produce the demanded commodity, the residual demand is
```

docs/control/system_level_control/slc_demand_following.md
```diff
@@ -0,0 +1,108 @@
...
+
+This framework provides a starting point for hybrid energy system control but is intended to be extended with more sophisticated strategies for complex multi-commodity systems.
+
+## Limitations
+
+- No cost awareness: The controller dispatches technologies purely to meet demand without considering operational costs, commodity prices, or economic optimization.
+- Even splitting across storage: When multiple storage technologies produce the demanded commodity, the residual demand is divided evenly among them (`demand / n_storage`), regardless of differences in capacity, state of charge, or efficiency.
+- Even splitting across dispatchable technologies: Similarly, any remaining demand after storage dispatch is split evenly across all dispatchable technologies (`remaining_demand / n_dispatchable`), without accounting for marginal costs or capacity constraints.
+- Fixed priority order: The dispatch order (curtailable → storage → dispatchable) is fixed in the current implementation.
```
jaredthomas68, line 102: @cfrontin ran into a case where (curtailable -> dispatchable -> storage) was needed. I don't think any changes are needed here now, but I expect the order will need to be flexible.

jaredthomas68: This happened due to intentionally undersized grid interconnect

docs/control/system_level_control/slc_profit_max.md
```diff
@@ -0,0 +1,121 @@
...
+        '</div>'
+    )
+)
+```
+
+## Dispatch Logic
+
+The controller follows a three-step dispatch process:
+
+1. **Curtailable technologies** run at full rated capacity - they are always profitable to produce (zero marginal cost).
```
jaredthomas68: available instead of rated?

docs/control/system_level_control/slc_profit_max.md
```diff
@@ -0,0 +1,121 @@
...
+This is the key difference from the {ref}`cost minimization controller <slc-cost-min>`: unprofitable dispatch is skipped entirely, so demand may go unmet.
+```
+
+## Commodity Sell Price
+
+The sell price can be configured in two ways in `system_level_control.control_parameters`:
+
+| Value | Description |
+| --- | --- |
+| Numeric (e.g. `0.06`) | Constant sell price in `$/(commodity_unit*h)` |
```
jaredthomas68: `$/(commodity_amount_unit)`?

docs/control/technology_level_control/technology_control_overview.md
```diff
@@ -1,6 +1,6 @@
...
+There are two different systematic approaches, or frameworks, in H2Integrate for technology-level control: [open-loop](#open-loop-control) and [pyomo](#pyomo-control). These two frameworks are useful in different situations and have different impacts on the system and control strategies that can be implemented. Both control frameworks are focused on technology-level dispatching. The open-loop framework has logic that is applicable to both storage technologies and converter technologies **IS THIS STILL TRUE?** and the pyomo framework is currently applicable to storage technologies. The technology-level storage controllers may curtail/discard commodity amounts exceeding the needs of the storage technology and the specified demand. However, any unused commodity may be connected to another down-stream component to avoid actual curtailment.
```
jaredthomas68: Yes, the open-loop framework does have converter relevant logic, as seen used in #773

h2integrate/control/control_strategies/system_level/test/test_slc_examples.py
```diff
@@ -0,0 +1,304 @@
...
+        assert wind_out.sum() > 0
+
+    with subtests.test("solar farm generates power"):
+        assert solar_out.sum() > 0
+
+    with subtests.test("natural gas dispatched"):
+        assert ng_out.sum() > 0
+
+    with subtests.test("grid used when needed"):
+        assert grid_out.sum() > 0
```
jaredthomas68, line 308: I'm loving all the tests! Nice job.

I think these tests should be `=` though instead of just `>` for the final PR

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,821 @@
...
+
+    Controller-specific configuration parameters may be read from
+    ``plant_config["system_level_control"]["control_parameters"]``
+    """
+
+    def initialize(self):
+        self.options.declare("driver_config", types=dict)
+        self.options.declare("plant_config", types=dict)
+        self.options.declare("tech_config", types=dict)
+        self.options.declare("slc_config", types=dict)
```
jaredthomas68, line 65: Would it make sense to merge the SLC config with the driver config or the plant config? I'm not sure it should have its own file.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,821 @@
...
+
+    def setup(self):
+        plant_config = self.options["plant_config"]
+        slc_config = self.options["slc_config"]
+
+        self.n_timesteps = plant_config["plant"]["simulation"]["n_timesteps"]
+
+        # Read pre-computed classification from plant_config
+        self.commodity = slc_config["demand_commodity"]
+        self.commodity_units = slc_config.get("demand_commodity_rate_units", None)
```
jaredthomas68: I think the `rate` specifier is important for clarity throughout, can we not drop it?

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,821 @@
...
+        self.techs_to_commodities = slc_config["tech_to_commodity"]
+
+        # There are multiple commodities being produced by technologies in the system
+        self.multi_commodity_system = (
+            True if len({e[-1] for e in self.techs_to_commodities}) > 1 else False
+        )
+
+        self.commodities_to_units = {self.commodity: self.commodity_units}
+        self.commodities_to_ref_var = {}
+        self._setup_fixed_category(self.fixed_techs)
```
jaredthomas68, line 119: why is "fixed" not included in `_setup_tech_category`?

h2integrate/control/control_strategies/passthrough_controller.py
jaredthomas68, line 1: I like this

h2integrate/core/h2integrate_model.py
```diff
@@ -462,6 +476,357 @@ def create_plant_model(self):
...
+                    f"system_level_controller.{tech_name}_{commodity}_storage_duration",
+                )
+
+            # Connect the controller's output back to the technology.
+            classifier = slc_config["tech_control_classifiers"][tech_name]
+            if slc_config["storage_techs_to_control"].get(tech_name, False):
+                # Storage tech with its own sub-controller: provide a demand
+                # signal that the sub-controller translates into
+                # charge/discharge commands.
+                self.plant.connect(
```
jaredthomas68: This approach makes sense to me and I think could be extensible to converters with their own controllers as well.

h2integrate/core/h2integrate_model.py
```diff
@@ -603,6 +976,22 @@ def create_technology_models(self):
...
+
+                        # Collect control classifier for system-level control
+                        if model_type == "performance_model" and self.slc:
+                            perf_cls = self.supported_models.get(perf_model)
+                            if perf_cls is not None:
+                                classifier = getattr(perf_cls, "_control_classifier", None)
+                                if classifier is not None:
+                                    self.tech_control_classifiers[tech_name] = classifier
+
+                if perf_om_object is not None:
```
jaredthomas68, line 988: It looks like this is intended to add a controller to techs without controllers, but from what I see it just adds a pass through controller to every tech, what am I missing?

jaredthomas68: NM, line 1084

jaredthomas68 (changes requested):
I am very impressed, really excellent work! Thank you for putting this together and so well. I appreciate how general, flexible, and yet robust it seems to be (having not actually tried using it myself yet). Thank you for adding in the pass through control, it looks like that needs to be included in the docs still (see comments).

We are going to need a path forward for pyomo optimal dispatch at the plant level. Is that path there are and I just don't see it?

docs/control/system_level_control/control_classifier.md
johnjasa: Understood and agreed. I envision this as follow-on work in separate PRs. #749 and #750 are probably the most relevant issues, with a few others coming into play

docs/control/system_level_control/slc_cost_min.md
johnjasa: Changed "full rated" to "available" throughout

docs/control/system_level_control/slc_cost_min.md
johnjasa: Good catch, changed throughout

docs/control/system_level_control/slc_cost_min.md
johnjasa: I could see that delineation being useful, primarily to be more precise when discussing system-level control vs. tech-level control. That being said, I do like how broad and inclusive "control" is as a term. @elenya-grant and @kbrunik and others, what do you think of Jared's suggested delineation here?

h2integrate/control/control_strategies/system_level/test/test_slc_examples.py
johnjasa: Are you suggesting that each of these output streams should be 0? Or that we should be testing different values? To be clear, each of these values would not be 0 as there would be some generation from each technology.

docs/control/system_level_control/slc_demand_following.md
johnjasa: Understood, great use case to consider for the future!

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: This is a good question, I revisited the possibility of combining the methods but I think they're worth keeping separate here because there fixed method is dramatically simpler than the more general one, and I think combining them would remove clarity for both.

I've added to the docstring for the `_setup_fixed_category`:
```
        This method is separate from the more general ``_setup_tech_category`` because the logic
        for fixed techs is dramatically simpler
        (no demand or rated inputs, only production inputs).
```

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: Good call, I've renamed this throughout to be `commodity_rate_units`

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: The SLC config is not actually its own yaml file or dict, but it is included as part of the `plant_config`, mostly through the logic of the system level controller being a property of the plant. This dict that's passed into the SLC is a lightly modified version of this that comes in through `H2IntegrateModel`. Are you okay with this delineation or do you prefer a different organization?

johnjasa:
> I am very impressed, really excellent work! Thank you for putting this together and so well. I appreciate how general, flexible, and yet robust it seems to be (having not actually tried using it myself yet). Thank you for adding in the pass through control, it looks like that needs to be included in the docs still (see comments).
> 
> We are going to need a path forward for pyomo optimal dispatch at the plant level. Is that path there are and I just don't see it?

Thanks for your review and suggestions, Jared! Very helpful stuff. I've went through and addressed all of your comments, mostly through changing the code, docs, or comments. I like your meticulousness on the naming. Please let me know if you want to see any more changes.

For the SLC with optimization pathway, I agree that is not included in this PR and is a necessary future step. I imagine some portions of this framework will need to be updated and expanded. I don't see the lack of optimal SLC as a hindrance for this PR. As you keep thinking of future work, please see the issues that already exist and continue to create your own!

docs/control/system_level_control/slc_cost_min.md
elenya-grant: I like the use of `system_level` as a descriptor. I don't have strong feelings whether it's called `system_level_dispatch` or `system_level_control`. I like @jaredthomas68 definition of "dispatch make set point decisions" and controllers trying to achieve those set point (although I'm struggling to understand what the output(s) of the controllers would be). If we go with that definition - then do all existing storage "controllers" in H2I fall under the "dispatch" definition?

.github/workflows/ci.yml
```diff
@@ -1,5 +1,10 @@
 name: Testing
 on:
+  push:
```
genevievestarke: Does this need to be adjusted when we merge in?

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
+# System Level Control Technology Performance Classifiers
+
+To enable a generic system level control framework we need to classify each technology based on how the model, that is included in H2I, can operate within the system.
+
+```{note}
+While in real life there are a lot of controllable parameters allowing for ramping production up or down for a particular technology (e.g., turbine yaw). The particular model in H2I might not be capable of simulating a modulated response based on an input signal.
```
genevievestarke:
```suggestion
While in real life there are a lot of controllable parameters allowing for ramping production up or down for a particular technology (e.g., wind or solar curtailment), the model of that technology in H2I might not be capable of the same response behavior to input signals.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+These classifications are for how the models in H2I are implemented, **not** how the actual physical subsystem might operate.
```
genevievestarke:
```suggestion
These classifications are for specific H2I dispatch formulations and are based on how the models in H2I are implemented, **not always** on how the actual physical subsystem might operate.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
+# System Level Control Technology Performance Classifiers
+
+To enable a generic system level control framework we need to classify each technology based on how the model, that is included in H2I, can operate within the system.
```
genevievestarke:
```suggestion
To enable a generic system level control framework we need to classify each technology based on how the model that is included in H2I can operate within the system.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+These classifications are for how the models in H2I are implemented, **not** how the actual physical subsystem might operate.
+This is a useful and necessary distinction that delineates different model capabilities clearly.
+```
+
+We have identified five key classifiers that are able to represent the different behaviors that we can expect from the models. Each performance model includes a parameter setting the classifier `_control_classifier`.
+
+Classifier | Meaning | Example Technology Models
+-- | -- | --
+fixed | Always produces commodity and cannot be controlled or reduced; does not receive a set-point | classical nuclear
+flexible | Produces based on resource; can only reduce (curtail) | wind, solar
```
genevievestarke:
```suggestion
flexible | Produces based on resource; can only reduce (curtail) from the maximum generation based on available resource | wind, solar
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+dispatchable | Can modulate consumption/production within bounds; receives a commodity set-point | grid, electrolyzer, NG turbine
+storage | Can modulate consumption/production within bounds while tracking SOC | battery, h2 storage, any storage
+feedstock | Are not directly controlled, but useful for SLC to know about to make dispatch decisions | feedstocks
```
genevievestarke:
```suggestion
feedstock | Are not directly controlled, but useful for SLC to make dispatch decisions | feedstocks
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+
+To add a classifier for a particular model it would look something like this in the class:
```
genevievestarke, line 21: This is in case someone is adding a new model, correct? All of the control classifiers have been set for current models?

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+```{python}
+_control_classifier = "flexible"
+```
+
+## Fixed
+A fixed performance model represents anything that always produces at its rated capacity and cannot be controlled or reduced by the system level controller. The SLC reads the output from a fixed technology and subtracts it from the demand, but does not send a set-point back to the technology. A good example of this is a classical nuclear plant model — it produces a constant output that the rest of the system must accommodate.
+
+## Flexible
+A flexible performance model represents anything that can have the output reduced based on a given set point from the system level controller. A good example of this is the PVWatts PySAM solar plant in H2I, the performance of the system is based on the input solar resource. The solar performance does not change based on, for example, an updated set point to the tracking software, but we could limit the power output from the solar performance model based on a given demand set point. To simplify the implementation of applying this curtailment or reduction based on a set point we added a method, `apply_curtailment()` to the `PerformanceBaseClass`.
```
genevievestarke:
```suggestion
A flexible performance model represents anything that can have the output reduced based on a given set point from the system level controller. A good example of this is the PVWatts PySAM solar plant in H2I; the performance of the system is based on the input solar resource. The solar performance does not change based on, for example, an updated set point to the panel tracking software. However, we could limit the power output from the solar performance model based on a given demand set point. To simplify the implementation of applying this curtailment or reduction based on a set point we added a method, `apply_curtailment()` to the `PerformanceBaseClass`.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+:width: 70%
+:align: center
+```
+
+### Apply curtailment based on set_point
+Within the `compute()` method in the performance model you can apply the curtailment using the `apply_curtailment()` method.
+```
+self.apply_curtailment(outputs)
+```
+which, applies curtailment to `{commodity}_out` based on `{commodity}_set_point`. There is then `uncurtailed_{commodity}_out` and `{commodity}_out` as outputs from the performance model.
```
genevievestarke:
```suggestion
which, applies curtailment to `{commodity}_out` based on `{commodity}_set_point`. This adds `uncurtailed_{commodity}_out` and `{commodity}_out` as outputs from the performance model.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Dispatchable
+A dispatchable performance model represents anything that can receive a set point. Any model that has the "dispatchable" control classifier tag is able to receive a set point and change it's behavior based on that set point. There aren't additional special methods to handle this because it's internal to each performance model.
```
genevievestarke: This is a slightly confusing definition since in the previous section we say that flexible technologies also receive a set point, and change their behavior based on it. 
Is flexible a special case of dispatchable?

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+
+```{figure} figures/dispatchable.png
+:width: 70%
+:align: center
+```
+
+## Storage
+Storage is a unique control classifier because it assumes that within the model that energy isn't created or destroyed (minus some efficiency losses). While it's technically "dispatchable" in that it can receive and change its performance based on a set point it's handling within H2I is unique because it's attached to storage performance models, which is handled differently than converter performance models. A converter model only has positive (or zero) `{commodity}_out`, whereas a storage model can have positive or negative `{commodity}_out`.
```
genevievestarke:
```suggestion
Storage is a unique control classifier because it assumes that within the model that energy isn't created or destroyed (minus some efficiency losses). While it's technically "dispatchable" in that it can receive and change its performance based on a set point, its handling within H2I is unique because it's attached to storage performance models, which is handled differently than converter performance models. A converter model only has positive (or zero) `{commodity}_out`, whereas a storage model can have positive or negative `{commodity}_out`.
```

docs/control/system_level_control/control_classifier.md
```diff
@@ -0,0 +1,69 @@
...
+The system-level controller outputs set points to the storage performance model which can be considered charge (negative) and discharge (positive) commands (storage-level set points) to the storage performance model, directly.
+
+
+```{figure} figures/storage.png
+:width: 85%
+:align: center
+```
+
+## Feedstock
+Another category of control classifiers are feedstocks. The unique thing about feedstocks is that they are considered outside of the controllable system within H2I. While they can't be controlled it can be helpful for controllers to know how much feedstock is available within the system, hence their classification.
```
genevievestarke:
```suggestion
The feedstocks category is unique in that they are considered outside of the controllable system within H2I. While they can't be controlled, it can be helpful for controllers to know how much feedstock is available within the system, hence their classification.
```

genevievestarke: Why are then considered outside the system? Does the controller make decisions based on the availability of a feedstock? I think they could be better classified as inputs to the system.

docs/control/system_level_control/slc_cost_min.md
genevievestarke: I agree that it would clarify things to go with `system_level_dispatch`. In my head, the `apply_curtailment` method for flexible technologies is a good example of technology-level control.

genevievestarke (commented):
Here are some high-level comments based on how far I've gotten!

docs/control/system_level_control/slc_demand_following.md
genevievestarke: In Hercules, we've implemented the curtailment order as a user input, so I think that could be added in a future PR!

docs/control/system_level_control/control_classifier.md
elenya-grant: Feedstocks are not controllable but we could imagine that the feedstock availability may be useful for more advanced controllers. Aka - the use-case that motivated this was something like, trying to meet some ammonia production demand and that demand cannot be met because of an insufficient supply of nitrogen feedstock. In that case - the controller may adjust the set points for the hydrogen and electricity technologies differently if it knows that the ammonia demand cannot be met due to the nitrogen supply (which is not controllable)

h2integrate/core/h2integrate_model.py
```diff
@@ -462,6 +476,336 @@ def create_plant_model(self):
...
 
+    def _classify_slc_technologies(self):
+        """Classify technologies for system-level control.
+
+        Uses ``self.tech_control_classifiers`` (populated by ``create_technology_models()``)
+        to partition technologies into fixed, flexible, dispatchable, and storage lists.
+        Also identifies the single demand technology and its commodity.
+
+        Returns:
+            dict: Classification dictionary (``slc_config``) with keys:
```
genevievestarke, line 491: Could we add better descriptions of what the outputs are here? (e.g. the description that starts on line 665 would be nice here, too)

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,825 @@
...
+    - ``demand_commodity``: the commodity being controlled (e.g. "electricity")
+    - ``demand_commodity_rate_units``: units string (or None) of the demand commodity
+    - ``demand_tech``: name of the demand technology
+    - ``storage_techs_to_control``: dictionary with keys of the technology names. The value is True
+        if the technology is classified as "storage" and has an attached controller.
+        Otherwise the value is False.
+    - ``technology_graph``: directional graph object representation of the
+        technology_interconnections found in the ``plant_config``
+    - ``tech_to_commodity``: set of tuples formatted as (tech_name, tech_output_commodity)
+    - ``tech_control_classifiers``: dictionary of technologies with keys as the technology names the
```
genevievestarke:
```suggestion
    - ``tech_control_classifiers``: dictionary of technologies with key-value pairs of each technology name and its corresponding control classifier. 
```
Does this make sense?

h2integrate/core/h2integrate_model.py
```diff
@@ -462,6 +476,336 @@ def create_plant_model(self):
...
+        slc_config["demand_commodity_rate_units"] = demand_commodity_rate_units
+        slc_config["tech_to_commodity"] = tech_to_commodity
+        slc_config["storage_techs_to_control"] = storage_tech_to_control
+        slc_config["technology_graph"] = self.technology_graph
+
+        slc_config["tech_control_classifiers"] = self.tech_control_classifiers
+
+        return slc_config
+
+    def add_system_level_controller(self, slc_config):
```
genevievestarke, line 620: More for my own understanding: Why do we add the system level controller and pre-compute this dictionary before handing off to the system level controllers in the control folder?

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,825 @@
...
+    def _setup_tech_category(self, category, tech_list):
+        """Create OpenMDAO I/O variables for all technologies in a given category.
+
+        This single method handles flexible, dispatchable, and storage
+        technologies. The logic is identical for all three categories —
+        iterate over each technology's commodities and register the
+        appropriate inputs (production output, rated capacity) and output
+        (per-tech demand).
+
+        All initial demand values are ``1.0``; the solver converges from there
```
genevievestarke, line 218: I don't love this, but it's something we can fix later.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,825 @@
...
+        using the connected rated-production inputs at run time.
+
+        After this method returns, four lists are stored on ``self`` under
+        names produced by the *category* prefix:
+
+            ``self.{category}_input_names``
```
genevievestarke, line 224: What does input mean here? input to the controller or the technology?

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -0,0 +1,825 @@
...
+                    in_name = f"{tech_name}_{commodity}_out"
+                    meta_data = self.add_input(
+                        in_name,
+                        val=0.0,
+                        shape=self.n_timesteps,
+                        units=None,
+                        units_by_conn=True,
+                        desc=f"{commodity} output from {tech_name}",
+                    )
+                    if meta_data["units"] is None:
```
genevievestarke, line 279: Could you explain what's happening here? Does the in_name always include units?

h2integrate/control/control_strategies/system_level/cost_minimization_control.py
```diff
@@ -0,0 +1,84 @@
...
+    2. Flexible techs run at rated capacity (assuming zero marginal cost).
+    3. Storage absorbs surplus / provides deficit.
+    4. Dispatchable techs are dispatched in ascending marginal-cost order,
+       each up to its rated capacity, until remaining demand is met.
+
+    Marginal costs are configured via ``cost_per_tech`` in the
+    ``system_level_control["control_parameters"]`` section of ``plant_config``.  Each
+    dispatchable technology's entry can be:
+
+    - A numeric value ($/commodity_unit, e.g. 0.05 for $0.05/kWh)
```
genevievestarke: It seems that `feedstock` is also an option for marginal costs?

examples/35_system_level_control/battery_with_controller/run_wind_ng_demand.py
```diff
@@ -0,0 +1,65 @@
...
+h2i.run()
+
+# Post-process the results
+h2i.post_process()
+
+# Plot the first 168 hours (1 week)
+n_hours = 168
+hours = np.arange(n_hours)
+
+wind_out = h2i.prob.get_val("plant.wind.electricity_out")[:n_hours]
```
genevievestarke, line 21: I'm getting an error when I run the examples because `plant.wind.electricity_out` doesn't exist. It seems to be because they don't exist on the plant object.

examples/35_system_level_control/battery_with_controller/run_wind_ng_demand.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+from h2integrate.core.h2integrate_model import H2IntegrateModel
+
+
+##################################
+# Create an H2I model with a fixed electricity load demand
+h2i = H2IntegrateModel("wind_ng_demand.yaml")
```
genevievestarke, line 9: There appears to be some funky stuff going on here where the battery is discharging above the demand. 

Image

genevievestarke: I think it might be a plotting issue filling in the area between two points

examples/35_system_level_control/yes_hydrogen/run_multi_commodity.py
```diff
@@ -0,0 +1,19 @@
...
+
+from h2integrate import EXAMPLE_DIR
+from h2integrate.core.h2integrate_model import H2IntegrateModel
+
+
+os.chdir(EXAMPLE_DIR / "35_system_level_control" / "yes_hydrogen")
+
+##################################
+# Create an H2I model with a fixed electricity load demand
+h2i = H2IntegrateModel("wind_ng_demand.yaml")
```
genevievestarke, line 11: I know that this concept is pretty new, so we might have to have some discussions to hash this out. Is the controller controlling electricity and hydrogen to meet a hydrogen demand? 
I find it pretty hard to parse the tech_config + system_level_control config, and I don't see where the commodity being controlled is set. Could we set that at the system_level_control config level instead of having it be inferred?

genevievestarke:
@johnjasa, I'm done with my initial review! This took a lot of work and it's a really cool functionality! I think it will take me some time to get caught up on all the details, but I think my big suggestion is that we are more clear what technologies this does and does not work for. For example, there is an apply_curtailment function for two of the wind models, but not for the ard model, and not for the tidal or hydro models, even though they're classified as flexible. Sorry if I'm not caught up on all of this, but where are we tracking what models can be used in this framework and which ones can't (if there are limitations)?

docs/control/system_level_control/slc_demand_following.md
johnjasa: Great call, I've made #775 to track this!

.github/workflows/ci.yml
johnjasa: Good catch, I've removed the `push` section! Thanks

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: What don't you love about this and what would you change it to? Should the initial demand values be something else?

docs/control/system_level_control/control_classifier.md
johnjasa: This is correct, yep! I'll add a note in the `adding_a_new_technology.md` page to make this more clear

docs/control/system_level_control/slc_cost_min.md
johnjasa: Good catch, I hadn't fully updated these docs since adding controllers for all techs. I just updated this throughout, thanks!

docs/control/system_level_control/control_classifier.md
johnjasa: Great Q @genevievestarke and thanks for chiming in here @elenya-grant! I've updated the doc page based on these comments; please give it a looksie and see if it's more clear

docs/control/system_level_control/control_classifier.md
johnjasa: Thanks for asking this, I know there's a fine line between these classifications. I've updated the doc page to better reflect the differences and similarities between them. Please give it a look and see if it's more in-line with your understanding, or if we can change this more!

h2integrate/core/h2integrate_model.py
johnjasa: Thanks for this, I've gone through and updated this with more info for each entry! In the process, I noticed a simplification we could do now that every technology always has a controller, so I appreciate your comment!

h2integrate/core/h2integrate_model.py
johnjasa: The short of it is for code readability and error-checking. If we can break out some of the methods and make them more atomic, it allows us to check the user inputs a bit sooner. I'd want most of that checking or processing to happen in the `h2integrate_model.py` level as some of the pre-computed values may be useful outside of the literal SLC system at some point (though maybe not yet).

examples/35_system_level_control/battery_with_controller/run_wind_ng_demand.py
johnjasa: Good call, I do think it's a plotting artifact. I redid all these plots so they're stacked bar charts instead to avoid this red herring: 
image

examples/35_system_level_control/yes_hydrogen/run_multi_commodity.py
johnjasa: @elenya-grant I'm tagging you in here so you can weigh in as well!

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: It means the literal variable inputs (read: OpenMDAO `add_input(<input_name>, ...`) used by the system level controller. So it'd be the commodity flows from the different technologies being used to meet the demand. Would you like to see changes to the code or comments to clarify this?

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: Basically, to get the units handling across different cases (i.e. we have commodity flows that match the demand units, or they have disparate units) I had to use some special handling here. Sometimes OpenMDAO is able to pass units via connection (introspect the connected variables and use the passed-in units), but other times these variables might not have units defined at this point in the model instantiation.

`in_name` is always just a string with the input name and does not include units definitions.

h2integrate/control/control_strategies/system_level/cost_minimization_control.py
johnjasa: Good call, updated this docstring and others to correctly reflect that

examples/35_system_level_control/battery_with_controller/run_wind_ng_demand.py
johnjasa: That's related to the OpenMDAO version; I've bumped up the minimum OM version as part of this PR. If you install OM 3.44.0 it should work, but please let me know if it doesn't!

examples/35_system_level_control/yes_hydrogen/run_multi_commodity.py
genevievestarke: Chatted with Elenya! I think it makes sense what it's doing, but maybe adding a bit more clarification around using the demand components with the system level control would be good.
I didn't realize that that's what determines where the control is applied (my fault for not reading the docs fully, but I think it could be more prominent), and that you could only have one demand component.

elenya-grant: I made issues about these to be handled in future PRs:
- #739 
- #740 
- #741

h2integrate/control/control_strategies/system_level/system_level_control_base.py
genevievestarke: Ok, chatted with Elenya about this, too, and I think it makes sense! This is just holding a space for the units so that they can be connected later when the connection information propagates down.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
genevievestarke: Maybe I'm not understanding the demand values, and those are external inputs? I was just thinking if this is the initial condition for the set points, we could probably find a way to pick better initial conditions.

examples/35_system_level_control/battery_with_controller/run_wind_ng_demand.py
genevievestarke: I love this new plot!

h2integrate/control/control_strategies/system_level/system_level_control_base.py
genevievestarke: Ah, ok, inputs to the controller and not the technologies. I think that makes sense!

genevievestarke (approved):
I think this all looks really good! I like the updated docs and the functionality is awesome! My only recommendation is that we make the mechanism of setting SLC more clear upfront in the docs: You set it with a demand component and you can only use one demand component when you're using SLC (currently).
I'd love for that to come in on this PR, but ultimately it's not blocking!

johnjasa:
> I think this all looks really good! I like the updated docs and the functionality is awesome! My only recommendation is that we make the mechanism of setting SLC more clear upfront in the docs: You set it with a demand component and you can only use one demand component when you're using SLC (currently). I'd love for that to come in on this PR, but ultimately it's not blocking!

Great idea Gen, I've added this stipulation (only one demand component allowed) in multiple places in the docs and docstrings. Thanks for your in-depth review and iteration with Elenya and me!

johnjasa (approved):
After much effort by many, multiple meetings and discussions, and iterations, I'm glad this is ready to come in! We can and will continue to modify and extend these capabilities. Great job here, team!

h2integrate/control/control_strategies/system_level/test/test_slc_examples.py
jaredthomas68: I would like to have us testing that the output is as expected instead of just non-zero. Just a tighter check to flag when things change.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
jaredthomas68: I think it works. I was confused about not having the user input configs mirror the input configs expected by the class. I think it is ok, but we may want to reconsider in the future for keeping things clean/clear for users.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -7,28 +7,52 @@ class SystemLevelControlBase(om.ExplicitComponent):
...
-    and technology classification reading from ``plant_config`` and ``slc_config``
+    demand input, fixed/flexible/dispatchable/storage/feedstock technology I/O
+    creation, and technology classification reading from ``plant_config`` and
+    ``slc_config``.
 
     Subclasses must implement ``compute()`` with their dispatch strategy.
 
+    Each technology group is expected to contain a controller (either user-defined or an
+    auto-injected ``PassthroughController``) that consumes a ``{commodity}_demand`` input and
+    produces the ``{commodity}_set_point`` actually fed to the performance/cost models. The
```
jaredthomas68: I'm not sure where the comments on this went, but I am a little confused on the naming here. I thought we had decided to have the SLC take *in* `demand` and *output* `set_point`. The tech control would take *in* `set_point` and *output* `control_command` or something like that.

jaredthomas68 (dismissed):
This is looking really good. I just have one outstanding comment we can likely resolve in dev meeting.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: > I'm not sure where the comments on this went, but I am a little confused on the naming here. I thought we had decided to have the SLC take _in_ `demand` and _output_ `set_point`. The tech control would take _in_ `set_point` and _output_ `control_command` or something like that.

I've tried doing this change for a few hours but couldn't get everything working well. I've captured this in an issue for future work: #778.

docs/control/system_level_control/system_level_control.md
```diff
@@ -0,0 +1,65 @@
+# System-Level Control
+
+System-level control (SLC) within H2I is meant to operate to control the entire plant with performance and cost feedback driving the operation of the plant or system in a closed-loop. It acts as a supervisory controller meaning that it can work to coordinate the entire system and can work with other technology level controllers.
+
+```{note}
+The SLC framework is *technology-agnostic* and works with any H2I technology (converters, storage, feedstocks, demand components, etc.). It only cares about a technology's [`_control_classifier`](control_classifier.md) and the commodity it produces. To opt a technology in, set `_control_classifier` on its performance model; for `flexible` models, also call `self.apply_curtailment(outputs)` at the end of `compute()`. See the [developer guide on adding a new technology](../../developer_guide/adding_a_new_technology.md) for the full checklist.
```
vijay092, line 6: Just to clarify, what does `flexible` mean here?

docs/control/system_level_control/system_level_control.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+The most basic SLC is shown in the figure below, where the SLC receives a demand. Based on that demand it emits a per-technology `{tech_name}_{commodity}_set_point` signal to each controlled technology. Each technology group contains a controller that converts that set-point into the `{commodity}_command_value` consumed by the technology's performance model. From each technology block there is `{commodity}_out` (potentially changed by the control signal) that is connected via feedback to the SLC. The SLC will then attempt to converge the system where it will loop through changing the set-points in attempts to meet the system demand until the overall system stops changing how much `{commodity}_out` each technology is outputting.
```
vijay092: This figure is not rendering correctly

vijay092: is the system controller for that block only?

vijay092: and this happens every timestep?

docs/control/system_level_control/system_level_control.md
```diff
@@ -0,0 +1,65 @@
...
+```{figure} figures/slc_basic.png
+:width: 70%
+:align: center
+```
+
+The SLC control strategy and solver options are set within `plant_config.yaml` under the `"system_level_control"` section.
+
+```yaml
+system_level_control:
+  control_strategy: DemandFollowingControl
```
vijay092, line 28: is the solver finding the optimal setpoint?

docs/control/system_level_control/slc_profit_max.md
```diff
@@ -0,0 +1,116 @@
...
+This is the key difference from the {ref}`cost minimization controller <slc-cost-min>`: unprofitable dispatch is skipped entirely, so demand may go unmet.
+```
+
+## Commodity Sell Price
+
+The sell price can be configured in two ways in `system_level_control.control_parameters`:
+
+| Value | Description |
+| --- | --- |
+| Numeric (e.g. `0.06`) | Constant sell price in `$/(commodity_amount_units)` |
```
vijay092, line 75: I thought there was also an option to specify a time-varying sell price?

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,67 @@
...
+# Cost Minimization System Level Controller
+
+The cost minimization controller, `CostMinimizationControl`, meets demand at minimum variable cost using merit-order dispatch.
+Unlike the {ref}`demand following controller <slc-demand-following>`, which splits demand evenly across dispatchable technologies, this controller dispatches the cheapest technologies first.
+
+## Dispatch Logic
+
+The controller follows a three-step dispatch process:
+
+1. **Curtailable technologies** run at their available capacity (assumed zero marginal cost). Their output is subtracted from the demand.
```
vijay092: A bit confused why this is called curtailable because we are not actually curtailing them?

docs/control/system_level_control/slc_cost_min.md
```diff
@@ -0,0 +1,67 @@
...
+  display_name: Python 3.11.13 ('h2i_env')
+  language: python
+  name: python3
+---
+
+(slc-cost-min)=
+# Cost Minimization System Level Controller
+
+The cost minimization controller, `CostMinimizationControl`, meets demand at minimum variable cost using merit-order dispatch.
+Unlike the {ref}`demand following controller <slc-demand-following>`, which splits demand evenly across dispatchable technologies, this controller dispatches the cheapest technologies first.
```
vijay092, line 18: Is there a reason why one would choose `demand following` over this controller?

vijay092 (commented):
Really nice work everyone! I mostly had some clarifying questions to make sure I understand things correctly.

docs/control/system_level_control/slc_cost_min.md
kbrunik: Perhaps not, but implementing the demand following was the simplest way to develop on the system level controller because it's not considering very many variables. Since we already had it made figured we could include it and it's a simple example to look at for other folks looking to develop controllers

docs/control/system_level_control/slc_cost_min.md
kbrunik: Fixed! This was a holdover from when we were calling tech like wind and solar "curtailable", we're now calling them "flexible". In this very simple cost min strategy the "flexible" tech is assumed to have zero marginal cost to operate so it's always allowed to run at full capacity. There's probably instances where there's a trade off in VarOpex between a tech like wind and running a natural gas plant and in some instances the wind could be curtailed in favor of the natural gas plant, but you're correct it's not being curtailed in this strategy as is

docs/control/system_level_control/slc_profit_max.md
kbrunik: Yes you can add time-varying sell price to the specific technologies. Right now the SLC profit max dispatch only uses single value floats. It probably wouldn't be too big of a lift to fully sync up the already existing time-varying sell price logic but a lot of the work in this PR was to showcase the connection points and how to consider different outputs from technology models not to have really good controllers, that's certainly the goal in the long run though!

docs/control/system_level_control/system_level_control.md
kbrunik: Yes, it's adjusting the set point based on the outputs it has access to such as the commodity_production and rated capacity and in the cost-aware strategies VarOpex and then looking at the differences in the outputs between runs. @johnjasa could probably elaborate a bit more.

docs/control/system_level_control/system_level_control.md
kbrunik: `flexible` means that the model outputs can be curtailed but otherwise cannot responded to a set point. The control classifiers are specific to the model as implemented in H2I so in reality a wind farm can respond to set points to adjust the turbines to better capture the wind, in H2I the pysam model will not change based on a setpoint received and the outputs of the wind farm simulation could be curtailed but that's it. Happy to chat more about this if that would be helpful. The classifiers were a big piece of discussion so it's certainly a bit confusing since they don't fully match with intuition.

docs/control/system_level_control/system_level_control.md
kbrunik: I updated the figure because it was out-of-date. The system level controller is for all technologies in the system. I changed the name from block to technology to try to make it slightly clearer.

Right now the solver operates over the full 8760 simulation not on a per-timestep basis.

docs/control/system_level_control/system_level_control.md
johnjasa: Well-said, Kaitlin! I will clarify the language slightly that it's solving for the set-points that reduce the residual between the demand and what's produced to 0, but isn't necessarily always finding the "optimal" set-point. Depending on the controller logic, it might just be meeting the demand without an optimization objective. There nominally could be many solutions and the solver is finding _one_ of them based on the controller used.

johnjasa (approved):

kbrunik (approved):

PR #755, opened 2026-05-18
Headroom/reserve rate connections

cfrontin:
# Headroom/reserve accounting for energy generation and storage

For certain types of analyses, there is value in redundant power generation, where power that _could_ be supplied at short notice but not currently online is valuable if not required.
One major place this becomes very important is for powering large loads like datacenters off the grid, where grid service reliability levels have to be recreated with on-site power.
This PR is an attempt to add "headroom" power outputs, which represent offline power (or other rate flow) capacity that could be used to provide redundancy, peak absorption, etc.

Headroom is different for different technologies: for a grid interconnect, say, it's simply the difference between the rated power output and the current power output.
For a natural gas plant, it's the power in excess of the current output that is limited by either the natural gas supply _or_ the rated output of the capital equipment, whichever is less.

For storage it's a bit more complicated. The internal headroom is:
- no less than zero
- no more than the discharge rating
- no more than the current available SOC divided by the storage dt

But then, this internal headroom must have the battery output subtracted (recall: discharge positive, charge negative), since, on one hand, active discharge counts against the capacity & rating that is offline-but-available and, on the other, active charge is excess power generation that could be diverted to power delivery, and therefore represents a form of headroom provided in some sense indirectly by the battery above and beyond its discharge capability.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [ ] clean up pre-commit formatting feedback

**Structural feedback**:

- noted above, battery charge flows are accounted as a battery-induced headroom, but there's probably cases where charging flows are not necessarily delivered through to overall system output loads when not used for charging; is there a better handling for this?

**Implementation feedback**:

- where should these features ultimately live in terms of the abstraction spectrum?
    - I made these changes at the level of the technologies, at the lowest level of specialization rather than the base classes
    - I didn't want to go too far up the abstraction chain, and when I explored doing so there were some ambiguities about how to define things (e.g. `_out` vs. `_sold` suffixes in `grid` aren't necessarily generalizable in higher levels of abstraction
- I use `headroom` to describe available but offline rate output from different technologies: is this the right terminology?
    - I find this useful but it's inspired largely by music amplification engineering (how much can i turn this amp up before it reaches rated power somewhere in the signal chain and distorts the signal) and may have a more appropriate energy and power nomenclature
    - it _was_ the preferred terminology in the storage model before I made any changes; [see this line](https://github.com/NatLabRockies/H2Integrate/blob/14623c36353feb5434ca433bd42259c8058c527c/h2integrate/storage/storage_baseclass.py#L334)

**Other feedback**:
N/A

- [ ] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

N/A.

### Section 5.1: New Files
- `path/to/file.extension`

- `path/to/file.extension`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/converters/grid/grid.py
```diff
@@ -99,10 +99,26 @@ def setup(self):
             "electricity_sold",
             val=0.0,
             shape=n_timesteps,
-            units="kW",
+            units=self.commodity_rate_units,
             desc="Electricity sold to the grid",
         )
 
+        self.add_output(
+            "electricity_headroom_sold",
```
genevievestarke: I think this naming convention is a little confusing. the current name makes it sound like this is electricity that has already been sold. Maybe `electricity_headroom_to_sell` or something would make sense?

h2integrate/converters/grid/test/test_grid.py
```diff
@@ -220,6 +235,13 @@ def test_selling_electricity(plant_config, n_timesteps):
     actual_in = prob.get_val("grid.electricity_in")
     np.testing.assert_array_almost_equal(actual_in, electricity_in)
 
+    # The headroom should be the difference between electricity_in and the rating
+    headroom = prob.get_val("grid.electricity_headroom_sold")
```
jaredthomas68: I think headroom is a good name. Could refer to https://docs.nrel.gov/docs/fy19osti/73590.pdf for discussion of the different forms headroom might take.

h2integrate/converters/grid/grid.py
jaredthomas68: I think `_to_sell` makes sense

h2integrate/converters/grid/grid.py
```diff
@@ -99,10 +99,26 @@ def setup(self):
             "electricity_sold",
             val=0.0,
             shape=n_timesteps,
-            units="kW",
+            units=self.commodity_rate_units,
```
jaredthomas68: The units of `electricity_sold` needs to be kW because it is interconnected with sell price in the cost model and comes from the same input in the config. That said, you could add the unit conversion in the cost model to make it work.

jaredthomas68 (commented):
I think this is fine and could come in with a few changes

h2integrate/converters/grid/grid.py
cfrontin: accidentally reverted before but fixed in https://github.com/NatLabRockies/H2Integrate/pull/755/commits/af59add5f315e2db5d629ebac499537f1c953713

h2integrate/converters/grid/grid.py
cfrontin: changed to _sell_headroom

cfrontin: in https://github.com/NatLabRockies/H2Integrate/pull/755/commits/31e43195baab5cba704a5c0b3202649d64ce547f

h2integrate/converters/grid/grid.py
johnjasa: Chiming in here, the performance model's output should be `self.commodity_rate_units` to match the rest of the compute; the conversion happens behind-the-scenes in OM when going from the performance to the cost model. The cost model always uses `units="kW"`; I'll push up this change now

johnjasa (approved):
Thanks for this, Cory! Nice atomistic improvement to the code. I pushed up some changes directly and will merge this in now.

Issue #742, opened 2026-05-09
Reduce GH actions testing costs


johnjasa:
We got an unofficial request through the grapevine to scale back the amount of GH actions tests we're running.
This is probably because we have a swath of tests across different Python versions that run for every PR and commit pushed to the PRs.

We could reduce the minutes used by having the full testing suite only run on non-draft PRs or those marked "ready for review," per @RHammond2's suggestion.
Me thinking this through, we could also have just a subset of the tests run (e.g. just unit tests for one version of Python) for PRs not marked ready, or something like that.

Another obvious but more time-consuming way to reduce our testing computational cost is to reduce the scope and depth of the tests we run.
We could examine the cases that are run, cut out the unnecessary ones or simplify them, and see where that gets us.

Definitely worth more thought.

RHammond2:
As discussed in previous standups, another course of action to take is to pare back the examples in the repo, or design specific tests around the use case. Similarly, it would be ideal if instead of rerunning the same case many times, we use fixtures to store the commonly reused run cases and test those more thoroughly.

PR #762, opened 2026-05-19
Defer heavy imports for faster package load times

johnjasa:
# Defer heavy imports for faster package load times

Dramatically reduces `import h2integrate` and `H2IntegrateModel` import times by deferring heavy transitive dependencies (OpenMDAO, HOPP, PySAM, pyomo, scipy, matplotlib, jsonschema, etc.) to the point of first use rather than loading them at module scope.

Old timing:
```
===========================================================================
H2Integrate Import Timing (existing timing)
===========================================================================
  import h2integrate                                        11.36s  [OK]
  from h2integrate import H2IntegrateModel                  10.33s  [OK]
  from h2integrate import load_yaml                         10.14s  [OK]
  Load supported_models + resolve 1 model                   12.63s  [OK]
  Load supported_models + resolve ALL models                11.46s  [OK]
```

New timing:
```
===========================================================================
H2Integrate Import Timing (with just-in-time loading)
===========================================================================
  import h2integrate                                         1.77s  [OK]
  from h2integrate import H2IntegrateModel                   1.53s  [OK]
  from h2integrate import load_yaml                          1.46s  [OK]
  Load supported_models + resolve 1 model                    6.23s  [OK]
  Load supported_models + resolve ALL models                11.68s  [OK]
```

No behavioral changes - all deferred imports are resolved on first use and cached by Python's import system.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Defer heavy imports in `h2integrate_model.py`
- [x] Add ruff per-file E501 ignore for `supported_models.py`
- [x] Verify all examples still run correctly

**Structural feedback**: Is the `_ModelRegistry` dict subclass approach acceptable, or would you prefer a different loading pattern?

**Implementation feedback**: Are there any methods in `h2integrate_model.py` where the local `import openmdao.api as om` pattern feels problematic?

**Other feedback**: N/A

- [ ] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

N/A

### Section 5.1: New Files
N/A

- `h2integrate/__init__.py`
  - Replaced imports with `__getattr__` lazy loader to avoid triggering the full dependency tree on `import h2integrate`.
- `h2integrate/core/supported_models.py`
  - Introduced `_ModelRegistry`; values don't include the `h2integrate.` prefix (prepended automatically in `_resolve`). 
  - Moved `import openmdao.api as om` and several other heavy imports from module-level to local method-level imports.
- `pyproject.toml`
  - Added `*/supported_models.py` to `[tool.ruff.lint.per-file-ignores]` for E501.

The primary bottleneck was `supported_models.py`, which imported all ~120 model classes and their transitive dependencies (HOPP, PySAM, pyomo, scipy, matplotlib, etc.) at module scope. The `_ModelRegistry` subclass stores import paths as strings and resolves them on first dict access, so only the models actually used in a given run are imported. Python's import caching means each module is only loaded once regardless of how many methods contain local import statements.


h2integrate/core/h2integrate_model.py
```diff
@@ -453,12 +456,16 @@ def create_plant_model(self):
         the same for each technology. This includes site information, project parameters,
         control strategy, and finance parameters.
         """
+        import openmdao.api as om
```
elenya-grant: does the `om` import really slow stuff down? If its used in multiple methods then I kind of think it should be left at the top of the file instead of in each method.

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,189 @@
...
+
+    def copy(self):
+        """Return a new _ModelRegistry with the same (still-deferred) entries."""
+        new = _ModelRegistry()
+        for k in super().keys():
+            new[k] = super().__getitem__(k)
+        return new
+
+
+supported_models = _ModelRegistry(
```
elenya-grant, line 47: I think this is slick! I'll be bummed that we won't be able to right click on a package object and have it open the file in VSCode anymore - but I will survive! If this reduces the import time then I'm down with it!

elenya-grant (approved):
Thanks for working on this! I think these changes are good!

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,189 @@
...
+        "TidalResource": "resource.tidal:TidalResource",
+        "RiverResource": "resource.river:RiverResource",
+        "WTKNLRDeveloperAPIWindResource": "resource.wind.nlr_developer_wtk_api:WTKNLRDeveloperAPIWindResource",
+        "OpenMeteoHistoricalWindResource": "resource.wind.openmeteo_wind:OpenMeteoHistoricalWindResource",
+        "OpenMeteoHistoricalSolarResource": "resource.solar.openmeteo_solar:OpenMeteoHistoricalSolarResource",
+        "GOESAggregatedSolarAPI": "resource.solar.nlr_developer_goes_api_models:GOESAggregatedSolarAPI",
+        "GOESConusSolarAPI": "resource.solar.nlr_developer_goes_api_models:GOESConusSolarAPI",
+        "GOESFullDiscSolarAPI": "resource.solar.nlr_developer_goes_api_models:GOESFullDiscSolarAPI",
+        "GOESTMYSolarAPI": "resource.solar.nlr_developer_goes_api_models:GOESTMYSolarAPI",
+        "MeteosatPrimeMeridianSolarAPI": "resource.solar.nlr_developer_meteosat_prime_meridian_models:MeteosatPrimeMeridianSolarAPI",
```
RHammond2, line 59: This might be fodder for a carry-on PR, but I think that we should make some of these models available at higher levels. For instance instead of every solar resource file having to be imported from its module, have all the primary models and functionality available at `from h2integrate.resource.solar`. The two areas where I could see this not catching on is: 1) when you import from `h2integrate.resource.solar` you end up with all the transitive imports required by each model, so it does increase some of the overhead, and 2) with the `supported_models` framework, it's a bit of a moot point because we rarely need to import specific models outside of the testing framework.

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,189 @@
...
+    strings, where the module path is relative to the ``h2integrate``
+    package (the ``h2integrate.`` prefix is prepended automatically).
+    On first ``__getitem__``, ``get``, or iteration over values, the class
+    is imported, cached, and returned.  This avoids importing every model
+    module (and their heavy transitive dependencies) at package-load time.
+    """
+
+    _PKG_PREFIX = "h2integrate."
+
+    def _resolve(self, key):
```
RHammond2, line 18: Could you add a docstring to this method? The info is contained at the class level, but is also applicable here. I think its easier to have duplicated descriptions so it's clear where something is happening.

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,189 @@
...
+    def __getitem__(self, key):
+        return self._resolve(key)
+
+    def get(self, key, default=None):
+        if key in self:
+            return self._resolve(key)
+        return default
+
+    def values(self):
+        return [self[k] for k in super().keys()]
```
RHammond2: I'm not sure I understand why we wouldn't return `list(super().values())` or not even override `values` in the first place. Could you provide a bit more context on this decision making?

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,189 @@
...
+
+    def items(self):
+        return [(k, self[k]) for k in super().keys()]
```
RHammond2: Same general comment as the one I left for `values`.

RHammond2 (changes requested):
I like the simplicity of this approach, and the massive improvements in import speedup. This should actually help a bit with #742 because of the sheer number of tests that import the model, which is a really nice secondary impact.

I have a couple of specific questions before I hit the approve button, and a couple of non-blocking general que

RHammond2 (changes requested):
I really like the simplicity of this approach and the all the added speedups. This should actually help a reasonable amount with #742 because of how many times the model is imported throughout the tests.

There are a couple spots in `_ModelRegistry` where some docstrings would be appreciated, a couple questions about the reimplementation of `dict` methods, and a non-blocking general question about import structures.

h2integrate/core/h2integrate_model.py
johnjasa: It doesn't really; I was trying out a few different things and this was leftover. I removed all of these and put the import at the top, thank you!

h2integrate/core/supported_models.py
johnjasa: I agree this obfuscates things a little bit, but I do think the reduced import time is worthwhile. This might also point us towards having more standardized naming conventions for files so you can easily find them in VSCode, e.g. `TidalResource` could be defined in `tidal_resource.py` instead of the (as-is) `tidal.py`, and maintain this same naming throughout H2I.

Do you think that would help quickly find what you're looking for? What other solutions are possible to help here?

h2integrate/core/supported_models.py
johnjasa: I like your suggestion overall! I do think with point 2) it becomes mildly moot unless there are use cases I'm not imagining. You mention that we should do this, can you help me see the value-add a bit more? Is it to simplify imports for users or devs, reduce `supported_models.py` complexity, or something else? Lastly, I do think this would fit well in a follow-on PR -- could you please make an issue for it? Could just be your comment copy-pasted, but also feel free to add more info/context/the answer to my value-add question there.

h2integrate/core/supported_models.py
johnjasa: I'm so glad we have PR reviews because this is firmly not needed, good catch! I've removed the `values` and `items` overriding.

h2integrate/core/supported_models.py
RHammond2: There likely isn't much of a use case for it other than to reign in the more nested subpackage pathways on top of it being a convenience functionality for devs. I'll add an issue and link this comment chain as well.

h2integrate/core/supported_models.py
RHammond2: Glad to be of service!

h2integrate/core/supported_models.py
johnjasa: Done, good call!

johnjasa:
> I really like the simplicity of this approach and the all the added speedups. This should actually help a reasonable amount with #742 because of how many times the model is imported throughout the tests.
> 
> There are a couple spots in `_ModelRegistry` where some docstrings would be appreciated, a couple questions about the reimplementation of `dict` methods, and a non-blocking general question about import structures.

Thanks, Rob! This is ready for re-review, I've implemented some good changes based on your comments.

h2integrate/core/supported_models.py
RHammond2: Created #767

h2integrate/core/supported_models.py
```diff
@@ -1,333 +1,187 @@
...
+    On first ``__getitem__``, ``get``, or iteration over values, the class
+    is imported, cached, and returned.  This avoids importing every model
+    module (and their heavy transitive dependencies) at package-load time,
+    reducing computational cost of imports and running tests.
+    """
+
+    _PKG_PREFIX = "h2integrate."
+
+    def _resolve(self, key):
+        """Import and cache the class for *key*, returning the resolved class."""
```
RHammond2:
```suggestion
        """Import and cache the class for :py:attr:`key`, returning the resolved class."""
```

RHammond2 (approved):
Thanks for the updates, @johnjasa!

PR #771, opened 2026-05-27
more robust api calls, suggested by copilot after all my api calls we…

ptrbortolotti:
Hi H2I team, my API calls kept failing despite having the correct keys. I then entertained a lengthy conversation with GH copilot that suggested some edits to the download_tools.py script to improve robustness, essentially switching from `r.text.replace` to `r.content.decode`. The fix works nicely for me and for Cory, despite Cory not having any issue previously. Can someone please test it out? If all looks good, it's probably a good idea to merge this in


h2integrate/resource/utilities/download_tools.py
```diff
@@ -43,7 +51,9 @@ def download_from_api(url, filename):
                 raise RuntimeError("Maximum API request rate exceeded!")
             else:
                 n_tries += 1
-        except requests.exceptions.Timeout:
+        except (requests.exceptions.Timeout, TypeError) as e:
+            # Catch Timeout and TypeError (from charset_normalizer issues)
+            print(f"Retry {n_tries + 1}/5 due to: {type(e).__name__}")
```
elenya-grant: We don't have print statements elsewhere in the code unless it's in post-processing. I wonder if this should either raise a user-warning or if the print statement should be removed:

```suggestion
            warnings.warn(f"Retry {n_tries + 1}/5 due to: {type(e).__name__}", UserWarning)
```

elenya-grant: Actually - I guess I'm lying - it does print out the filename if theres an error in some above logic (perhaps that should be re-done too)

h2integrate/resource/utilities/download_tools.py
```diff
@@ -23,18 +23,26 @@ def download_from_api(url, filename):
...
-                text_json = json.loads(r.text)
-                if "errors" in text_json.keys():
-                    err = text_json["errors"]
+                err = r.content.decode("utf-8", errors="replace")
+                try:
+                    text_json = json.loads(err)
+                    if "errors" in text_json.keys():
+                        err = text_json["errors"]
+                except json.JSONDecodeError:
+                    pass
```
elenya-grant: why `pass` if theres a `JSONDecodeError`?

h2integrate/resource/utilities/download_tools.py
```diff
@@ -23,18 +23,26 @@ def download_from_api(url, filename):
             r = requests.get(url)
             if r:
                 localfile = Path(filename).open("w+")
-                txt = r.text.replace("(Â°C)", "(C)").replace("(Â°)", "(deg)")
+                # Use r.content.decode() to avoid charset_normalizer issues
+                txt = (
```
elenya-grant, line 27: does `r` always have the `content` attribute?

elenya-grant (commented):
Howdy Pietro! I think this looks like a clean fix for whatever problem you were running into! Thanks for this! I'm curious about what prompted these changes (good changes - just want to know more about what was happening for you). 

Do you think in the PR description you could show/share more details about the error message you were getting (prior to this fix)? I haven't tried to run this locally yet - but I can check that it works for me too. What dataset were you making an API call to when you ran into the errors?

h2integrate/resource/utilities/download_tools.py
elenya-grant: Aka - no changes needed on this bit!

h2integrate/resource/utilities/download_tools.py
ptrbortolotti: The code is trying to extract the error message when it is json formatted. If it isn't json formatted, it just keeps going instead of crashing

h2integrate/resource/utilities/download_tools.py
ptrbortolotti: from CoPilot: Yes, r always has the content attribute when it's a valid requests

ptrbortolotti:
I have been trying to replicate the error that I was running into but now I cannot... frankly we could simply replace this line 'txt = r.text.replace("(Â°C)", "(C)").replace("(Â°)", "(deg)")'. this is what was failing. the rest is CoPilot doing stuff...

h2integrate/resource/utilities/download_tools.py
```diff
@@ -23,18 +24,26 @@ def download_from_api(url, filename):
...
-                text_json = json.loads(r.text)
-                if "errors" in text_json.keys():
-                    err = text_json["errors"]
+                err = r.content.decode("utf-8", errors="replace")
+                try:
+                    text_json = json.loads(err)
+                    if "errors" in text_json.keys():
+                        err = text_json["errors"]
+                except json.JSONDecodeError:
+                    pass
```
johnjasa:
```suggestion
                err = r.text
                text_json = json.loads(r.text)
                if "errors" in text_json.keys():
                    err = text_json["errors"]
```

h2integrate/resource/utilities/download_tools.py
```diff
@@ -43,7 +52,9 @@ def download_from_api(url, filename):
                 raise RuntimeError("Maximum API request rate exceeded!")
             else:
                 n_tries += 1
-        except requests.exceptions.Timeout:
+        except (requests.exceptions.Timeout, TypeError) as e:
+            # Catch Timeout and TypeError (from charset_normalizer issues)
+            warnings.warn(f"Retry {n_tries + 1}/5 due to: {type(e).__name__}", UserWarning)
```
johnjasa:
```suggestion
        except requests.exceptions.Timeout:
```

h2integrate/resource/utilities/download_tools.py
```diff
@@ -1,5 +1,6 @@
 import json
 import time
+import warnings
```
johnjasa:
```suggestion
```

PR #777, opened 2026-06-09
Fix: Minor update to EIA NG pricing

RHammond2:
# Fix: Minor update to EIA NG pricing

This fix provides a series of minor patches to the EIA natural gas feedstock cost model and preprocessing functionality.
- Updated control flow for reading from an already downloaded file, and not requiring the API
- Add a missing conditional in the dataframe filtering
- Adds the missing `discrete_inputs` and `discrete_outputs`
- Ensures a truth array is not passed directly to an `if`

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [ ] Documentation
- [x] `CHANGELOG.md`

### Section 5.1: New Files
N/A

- `h2integrate/feedstocks/eia_ng_price.py`
  - `EIANaturalGasFeedstockConfig.price`: Converted to the expected NumPy array type.
  - `EIANaturalGasFeedstockCostModel.setup`: Passes the `price` array to the configuration instead of a dataframe.
  - `EIANaturalGasFeedstockCostModel.compute`: Includes previously unnecessary `discrete_inputs` and `discrete_outputs`, and fixes the `if` condition to an `array.all()` in place of a truth array.
- `h2integrate/preprocess/eia.py`
  - `get_eia_ng_data`: URL creation is moved to after checking and using an existing file, a missing conditional is added, and missing function arguments are added.
- `h2integrate/feedstocks/test/test_feedstocks.py`: Adds a test to run the EIA NG feedstock model from a downloaded file, and updates for duplicated `plant_config` usage.

Tests pass.


h2integrate/feedstocks/test/test_feedstocks.py
```diff
@@ -169,14 +204,83 @@ def test_single_feedstock_natural_gas():
...
+def test_eia_natural_gas_feedstock(plant_config, eia_ng_feedstock_tech_config):
+    """Test a single natural gas feedstock with basic parameters."""
+    tech_config = eia_ng_feedstock_tech_config
+    _, driver_config = create_basic_feedstock_config()
+
+    # Test performance model
+    perf_model = FeedstockPerformanceModel()
+    perf_model.options["tech_config"] = tech_config
+    perf_model.options["plant_config"] = plant_config
+    perf_model.options["driver_config"] = driver_config
```
elenya-grant:
```suggestion
    perf_model = FeedstockPerformanceModel(
       tech_config = tech_config, 
       plant_config = plant_config, 
       driver_config = {}
       )
```

h2integrate/feedstocks/test/test_feedstocks.py
```diff
@@ -169,14 +204,83 @@ def test_single_feedstock_natural_gas():
...
+    # Check that output is generated correctly
+    ng_output = prob.get_val("feedstock_perf.natural_gas_out")
+    assert len(ng_output) == 8760
+    assert np.all(ng_output == 100.0)
+
+    # Test cost model
+    cost_model = EIANaturalGasFeedstockCostModel()
+    cost_model.options["tech_config"] = tech_config
+    cost_model.options["plant_config"] = plant_config
+    cost_model.options["driver_config"] = driver_config
```
elenya-grant:
```suggestion
    cost_model = EIANaturalGasFeedstockCostModel(
       tech_config = tech_config,
       plant_config = plant_config,
       driver_config = {}
       )
```

h2integrate/feedstocks/eia_ng_price.py
```diff
@@ -164,11 +159,11 @@ def setup(self):
             filename=self.config.filename,
         )
         price = eia.convert_to_hourly(price)
-        self.config.price = price
+        self.config.price = price.price.to_numpy()
```
johnjasa, line 162: I think this is throwing off the tests, maybe should be this?
```suggestion
        self.config.price = price.to_numpy()
```

RHammond2: This is only partially related. The code was updated to expect a NumPy array, but the test was left unmodified to handle this change. I have since pushed a commit addressing that change.

elenya-grant (approved):
Hi Rob! I think this looks great! Had a few minor suggestions but nothing blocking. 

One consideration/thought I want to run by ya! In the resource models, it downloads data from the API and each resource model has a specified file-naming convention. So, even if a user does not provide a filename, it will look to see if that filename already exists in the RESOURCE_DIR. If that filename does not exist, the resource data will be downloaded to the RESOURCE_DIR using the user-supplied filename instead of the default. This ends up having a bit of complicated logic but my main thought is would a built-in file naming convention be useful here so that duplicate files aren't downloaded? I also think that the EIA grid data you've uploaded for testing here could be saved in the `resource_files` folder in a subfolder named `natural_gas_prices` or similar (but maybe with a filename like `EIA_ng_prices_MN_2001.csv`). (I'll admit that I haven't run this myself so I'm not sure if there is already a file naming convention that I'm missing)

Again - nothing blocking from me at the moment! Thanks!

RHammond2:
> One consideration/thought I want to run by ya! In the resource models, it downloads data from the API and each resource model has a specified file-naming convention. So, even if a user does not provide a filename, it will look to see if that filename already exists in the RESOURCE_DIR. If that filename does not exist, the resource data will be downloaded to the RESOURCE_DIR using the user-supplied filename instead of the default. This ends up having a bit of complicated logic but my main thought is would a built-in file naming convention be useful here so that duplicate files aren't downloaded? I also think that the EIA grid data you've uploaded for testing here could be saved in the `resource_files` folder in a subfolder named `natural_gas_prices` or similar (but maybe with a filename like `EIA_ng_prices_MN_2001.csv`). (I'll admit that I haven't run this myself so I'm not sure if there is already a file naming convention that I'm missing)

I think this is a good idea in general. Is there a model example to base this off of? Based on perusing the current usage, I think the biggest lift will be adding the `resource_dir` into the existing `eia.py` workflows, but I think we could just pawn that off to the cost configuration class to create the expected `filename`. If we're mostly using this to set the default `resource_dir` to `DEFAULT_RESOURCE_DIR`, then I think this update would especially make sense.

johnjasa (approved):
Thanks for this and your patience, Rob! Really glad to see these bug fixes.

PR #788, opened 2026-06-22
Adding inflation rate to numpy NPV calc

### LINKED ISSUE #339: Feature request: allow user to provide inflation rate to numpy financial model
# Allow user to provide inflation rate to numpy financial model

While numpy financial NPV only accepts a single rate, the rate that the user should provide depends on what the user wants out. The user may want to include inflation in the rate, or not, or even provide only the inflation as the rate, and I think we can help simplify this for the user by allowing the user to include inflation separately and doing the adjustment for them. 

## Proposed solution
The could provide the real discount rate and the inflation rate (which would be summed internally), or provide the nominal discount rate and set inflation to zero.

## Alternatives considered
Just leave it as is and let the user worry about the math. However, once we start including more details in the numpy financial model we may have to add the ability to specify inflation separately.

## Additional context
See discussion here: https://github.com/NREL/H2Integrate/pull/310#discussion_r2474263888

johnjasa:
# Allow user to provide inflation rate to NumPy Financial NPV model

Resolves #339.

1. Added an optional `inflation_rate` input to `NumpyFinancialNPVFinanceConfig`. Defaults to `0.0`, validator-constrained to `[0, 1]`.
2. Combined it with `discount_rate` via the Fisher equation `(1 + r_eff) = (1 + discount_rate) * (1 + inflation_rate)` to match how ProFAST combines its real discount rate and `general_inflation` inputs, keeping the two finance backends consistent.
3. Default behavior is preserved: with `inflation_rate=0` the effective rate reduces to `discount_rate`, so existing configs and tests don't change.

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

**Structural feedback**:
- The original issue suggested summing the two rates. I switched to the Fisher equation `(1 + r_eff) = (1 + discount_rate) * (1 + inflation_rate)` because it's the exact relationship between real and nominal rates and is what ProFAST uses internally.

**Implementation feedback**:
- `inflation_rate` is validated with `range_val(0, 1)` to mirror `discount_rate`. Let me know if a different range (e.g. negative inflation) should be allowed.

**Other feedback**:
- N/A

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues

Fully resolves [#339](https://github.com/NatLabRockies/H2Integrate/issues/339).

### Section 5.1: New Files
- N/A

- `h2integrate/finances/numpy_financial_npv.py`
- `h2integrate/finances/test/test_numpy_financial_npv.py`

N/A

For typical project finance rates (real 5%, inflation 4%), the multiplicative form gives `r_eff = 0.092` vs. the additive `0.09` results in a 0.2 pp difference.

```
$ python -m pytest h2integrate/finances/test/test_numpy_financial_npv.py -v
...
5 passed, 8 subtests passed in 0.37s
```

Ruff lint + format clean on the two modified files.

N/A


docs/finance_models/numpy_financial_npv.md
```diff
@@ -17,7 +17,8 @@ Implements validation and default handling using the `attrs` library.
 | Attribute                         | Type             | Description                                           | Default     |
 | --------------------------------- | ---------------- | ----------------------------------------------------- | ----------- |
 | `plant_life`                      | `int`            | Operating life of the plant in years. Must be ≥ 0.    | —           |
-| `discount_rate`                   | `float`          | Discount rate (0–1).                | —           |
+| `discount_rate`                   | `float`          | Discount rate (0–1). Can be either a real or nominal rate depending on how `inflation_rate` is specified. | —           |
```
jaredthomas68: It might be more clear to specify if we want real or nominal if we are always adjusting with the Fisher equation. So, instead of `discount_rate`, call it `real_discount_rate` if we are always adding in inflation to the discount rate.

docs/finance_models/numpy_financial_npv.md
```diff
@@ -30,7 +31,8 @@ An example of what to include in the `plant_config` to use the `NPVFinance` mode
 npv:
   finance_model: "NumpyFinancialNPV"
   model_inputs:
-    discount_rate: 0.09 # each period is discounted at a rate of `discount_rate`
+    discount_rate: 0.09 # each period is discounted at a rate of `(1 + discount_rate) * (1 + inflation_rate) - 1`
+    inflation_rate: 0.0 # optional, defaults to 0; provide e.g. 0.025 if `discount_rate` is a real rate
```
jaredthomas68:
```suggestion
 inflation_rate: 0.0 # optional, defaults to 0; provide e.g. 0.025 if inflation rate is 2.5%. 
```

docs/finance_models/numpy_financial_npv.md
```diff
@@ -64,7 +66,7 @@ npv:
     - Technologies with `replacement_cost_percent` and a refurbishment period incur periodic capital costs.
 
 3. Discounting
-    - Each series of cash flows is discounted using NumPy Financial’s `npf.npv(discount_rate, values)`.
+    - Each series of cash flows is discounted using NumPy Financial’s `npf.npv(effective_rate, values)`, where `effective_rate = (1 + discount_rate) * (1 + inflation_rate) - 1` is the Fisher-equation combination of the user-supplied real (or nominal) discount rate and inflation rate. When `inflation_rate` is left at its default of 0, this reduces to `discount_rate`, so the user can supply either a nominal discount rate alone or split it into a real discount rate plus an inflation rate. The multiplicative (Fisher) form matches the way ProFAST combines its real discount rate and `general_inflation` inputs.
```
jaredthomas68: I think this discussion is slightly confusing. See my comment above about calling the input `real_discount_rate`.

docs/finance_models/numpy_financial_npv.md
jaredthomas68: The way we are using `discount_rate`, I think it should probably be the weighted average cost of capital (WACC) unless we are also getting the debt rate and debt/equity ratio and the tax rate as inputs.

jaredthomas68 (changes requested):
I think this is a good start. However, there are a few other related values that we should be aware of in our implementation and documentation to be clear and help users properly analyze the results. The rate used as shown should probably be the WACC, but that is not mentioned anywhere in the documentation. If we want discount rate to be the expected rate of return on equity, then we need to also request the debt rate the, the tax rate, and the debt/equity ratio (or debt-equity split).

docs/finance_models/numpy_financial_npv.md
elenya-grant: numpy NPV is a very simple NPV calculation - finance parameters like debt to equity ratio or debt rate are not included.

johnjasa: Understood on your views here, Jared! I think we should do this renaming of discount rate across all models simultaneously outside of this PR so I've made an issue here: #812 

I agree with Elenya that the numpy NPV calc is so simplistic that other values like WACC might not come into play. Are you suggesting that we have a different input into the numpy NPV model, either in value or name?

jaredthomas68: @elenya-grant, this is more a question of using the tool properly than whether something is included in the model. The model accepts a single rate. The single rate is technically WACC, which would reduce to just debt or just equity (possibly with tax applied depending on your use case) if you have only a single finance stream. If we use the numpy financial models incorrectly, then we will get an incorrect result. If we expect the user to supply the WACC as the discount rate, then we need to make that clear in the docs. Right now, it looks like the user is supplying the equity rate as discount rate because that is what we do in ProFAST. Adding a WACC method would be simple and make the user's life easier.

jaredthomas68: I added WACC, and clarification that this finance model is pre-tax

PR #792, opened 2026-06-24
Hooking up electricity buy price and more correctly

### LINKED ISSUE #790: Correctly connect `buy_price` to system level controllers that use it
# Correctly connect `buy_price` to system level controllers that use it

Right now, the `buy_price` variable within a technology (e.g. grid) is not connected to the SLC controller's `buy_price` value.
If it's a scalar price, we use the config values from the tech.
However, if you're setting the `buy_price` via `set_val` (usually because it varies in time, e.g. is an array) then you have to set it in both the grid tech _and_ the SLC because it's not currently connected.

## Proposed solution

Connect the `buy_price` directly to the SLC.
I think I didn't do this before because there was some coupling introduced that negatively affected convergence, but that was a while ago and my memory is fuzzy.
I can't think of a legit reason that wouldn't work now.

johnjasa:
# Connect `buy_price` from technology cost models to the SLC controller

Resolves the linked issue. There were some complexities because of the flexibility afforded by grid electricity price inputs (can be a scalar, `plant_life`, or an 8760) as well as the feedstocks setup.

1. Wired each cost-aware system-level controller's `{tech_name}_buy_price` input directly to the corresponding technology's own buy-price input via an OpenMDAO 3.44 input-to-input connection in `H2IntegrateModel.add_system_level_controller`, so a single `prob.set_val()` on (for example) `plant.grid.electricity_buy_price` now propagates to the SLC instead of silently using only the YAML default.
2. The connection is created on `self.model` (not `self.plant`), because OpenMDAO 3.44 only resolves input-to-input connections registered on the top-level group; using promoted names works because the plant subgroup is added with `promotes=["*"]`.
3. Added a small `_get_tech_buy_price_input_name` helper in `system_level_control_base.py` (shared by both files) that returns the tech's actual input name (`electricity_buy_price` for Grid, `price` for Feedstock).
4. Added `_get_buy_price_default_and_shape` so the SLC's `{tech}_buy_price` input is created with the same shape as the tech's input (driven by `buy_price_mode` for Grid; by price-array length for Feedstock). This is required because input-to-input connections need matching shapes; per-timestep, per-year, constant, and scalar configurations are all handled.
5. Updated `_buy_price_marginal_cost` to broadcast scalar / per-year arrays up to `n_timesteps` so the existing merit-order dispatch code is unchanged.
6. Fixed the `test_slc_complex_profit_max` test, which was implicitly relying on the bug (the example's `grid_buy_price = sell_price + 0.02` makes grid never profitable, but the SLC was using the constant `0.06` default, which sometimes beat the sell-price spikes). The corrected assertion verifies grid is never dispatched, which is now the right behavior.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [x] Bug Fix
- [x] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

**Implementation feedback**:
- `_get_tech_buy_price_input_name` currently recognizes `electricity_buy_price` (Grid) and `price` (Feedstock). If other techs grow buy-price inputs, this is the spot to extend. It may be worthwhile to standardize the naming now or in the future?

**Other feedback**:
- N/A

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues
Fully resolves #790

### Section 5.1: New Files
- N/A

- `h2integrate/core/h2integrate_model.py` adds the input-to-input connection for `cost_per_tech: buy_price` techs and updates the relevant docstring.
- `h2integrate/control/control_strategies/system_level/system_level_control_base.py` adds the two helpers, sizes the SLC `{tech}_buy_price` input to match the tech, and broadcasts in `_buy_price_marginal_cost`.
- `h2integrate/control/control_strategies/system_level/test/test_slc_examples.py` replaces the buggy "grid used when needed" assertion with the now-correct "grid not dispatched when always unprofitable" check.

The user explicitly requested using the OpenMDAO 3.44 input-to-input connection feature here. The 25 unit tests in `test_slc_controllers.py` keep using `set_val("slc.{tech}_buy_price", ...)` because those tests instantiate the SLC component in isolation (no tech subgroup to connect from), so they validate the controller behavior independently of the new wiring.

N/A

N/A


h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -648,9 +725,22 @@ def _buy_price_marginal_cost(self, inputs, tech_name):
...
-        return np.broadcast_to(inputs[f"{tech_name}_buy_price"], self.n_timesteps).copy()
+        buy_price = np.asarray(inputs[f"{tech_name}_buy_price"])
+
+        if buy_price.shape == (self.n_timesteps,) or buy_price.shape == (1,):
+            return np.broadcast_to(buy_price, self.n_timesteps).copy()
+
+        if buy_price.shape == (int(self.options["plant_config"]["plant"]["plant_life"]),):
+            # Per-year price: use the first year's value as a representative
+            # per-timestep marginal cost for dispatch decisions.
+            return np.full(self.n_timesteps, buy_price[0])
```
genevievestarke, line 743: This is because we are only simulating the first year?

genevievestarke (approved):
This looks good to me, thank you for adding it @johnjasa ! 
I just had a small clarifying question, but nothing to block merging!

h2integrate/control/control_strategies/system_level/system_level_control_base.py
johnjasa: That's right, yep!

Issue #765, opened 2026-05-20
Request: Better/more generic handling of API credentials


RHammond2:
# Simplify and make API credential handling more consistent

The current handling of API credentials is geared exclusively towards NLR-maintained APIs,  however, there are useful components within the tooling that could be made available for the integration of other API endpoints. For example in #719, a custom implementation of the EIA API credential handling was implemented (see comments: https://github.com/NatLabRockies/H2Integrate/pull/719#discussion_r3196889805) to work around the over-specified nature of the `h2integrate/resource/utilities/nlr_developer_api_keys.py`. Ideally, this would be a non-resource-specific module called `api_handling.py` (or similar) in `h2integrate/tools/` that generically specificies where and how API keys should be stored that works with NLR and other third-party tooling.

## Proposed solution

- [ ] Move `h2integrate/resource/utilities/nlr_developer_api_keys.py` to `h2integrate/tools/api_handling.py` (or similar) for API handling access availability outside of the resource package.
- [ ] Move API handling functions in `h2integrate/preprocess/eia.py` to the same generic API handling module.
- [ ] Create generic API credential querying from common locations:
  - [ ] home directory dot file (e.g. `~/.env`, `~/.nlrrc`, `~/.h2integraterc`
  - [ ] custom locations (i.e., project directories, API-specific required file locations)
  - [ ] environment variable
- [ ] Have endpoint-specific access handling utilize generic methods as much as possible to avoid repetitive functionality.

## Alternatives considered

The only other alternative that works currently is having each new API implementation handling whenever new functionality is introduced. This works from the #719 implementation, however it was brought up in both the prototyping and review stages that it would be nice to use the existing functionality, though it was mutually agreed upon that it's out of scope to refactor the existing API handling at the same time.

elenya-grant:
I started to work on generalizing the functions in `h2integrate/resource/utilities/nlr_developer_api_keys.py` to be compatible for other API keys and calls. I will push up it up to a branch and link it here shortly! I have not added in functionality to get variables from other files listed (like `.nlrrc` or `.h2integraterc`)

elenya-grant:
@RHammond2 - do you think this issue was resolved in PR #798?

elenya-grant:
Resolved with PR #798 and #834 (unless @RHammond2 has other thoughts)

PR #798, opened 2026-07-02
Generalized functions for getting, setting, and finding environment variables

elenya-grant:
# Generalized functions for getting, setting, and finding environment variables

Generalized the functions in `h2integrate/resource/utilities/nlr_developer_api_keys.py` for getting, finding, and setting environment variables. The functions in `nlr_developer_api_keys.py` were only usable for getting, finding, and setting environment variables for downloading resource data from NLR API calls. 

The generalized functions now live in `h2integrate/core/env_tools.py`. The functions for NLR API calls have been updated to use these generalized functions. 

This PR will make it easier for new environment variables to be introduced and used across models and tools without having to duplicate a bunch of code. Another benefit of this PR is that is has introduced more thorough testing for these methods.

A future PR could utilize this functionality to check if `RESOURCE_DIR` is specified as an environment variable.

Major shoutout to @RHammond2 who helped a lot with making this code an improvement on the existing API functionality and gave really great feedback throughout!

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [x] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Add tests for cases where some environment variables are set as environment variables and some need to be loaded from a file
- [x] finish making changes based on Rob's feedback

**Structural feedback**:

**Implementation feedback**:
- Is there a case where an environment variable will not be a string or wouldn't be usable in string format? All the methods have logic checking if the length of the environment variable value is zero, but this logic would break if the environment variable is set to a float, integer, or boolean. I think that I'd prefer to keep the assumption that environment variables are strings because most of the loading or getting environment variable methods will return the environment variable as a string. (Depending on the responses to this, I may make a follow-on issue but will not integrate the ability for these methods to handle non-strings in this PR)

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues

This is intended to resolve Issue #765

Issue #802 was created to note possible follow-on work that could be done once this PR is merged in!

Issue #811 was created to document some broader reaching reviewer feedback that could be resolved in a follow-on PR and should be discussed within the team.

### Section 5.1: New Files
- `docs/misc_resources/debugging_environment_variables.md`: new doc page that outlines ways to debug environment variables or set them prior to running H2I
- `h2integrate/core/env_tools.py`
  - `get_environment_variables()`: main method that is used to return the key/value pairs of specified environment variables where the environment variable value was found (and set any environment variables if specified).
  - `load_env_vars_from_file()`: loads all environment variables from a file, flexible to different separators
  - `set_env_var()`: method that sets values of environment variables from a dictionary
- `h2integrate/core/test/test_env_tools.py`: new test file for the functions in `h2integrate/core/env_tools.py`

- `h2integrate/resource/utilities/nlr_developer_api_keys.py`: removed old getter/setter methods and global variables, totally updated to use `get_environment_variables()`
  - `get_nlr_developer_api_credential`: general method to get either nlr api key or email
  - `get_nlr_developer_api_email`: updated to wrap `get_nlr_developer_api_credential` to get API email
  - get_nlr_developer_api_key`: updated to wrap `get_nlr_developer_api_credential` to get API key
- `docs/getting_started/environment_variables.md`: reformatted/updated to include the EIA API info.


RHammond2 (changes requested):
@elenya-grant I think this is a great step in the right direction, especially with moving much of the functionality out of the resource subpackage!

I have two significant hangups with the current draft: 1) we should generally avoid using global variables as a best practice, and 2) much of the refactored code is only lightly refactored, leaving it still highly geared towards an NLR credential getting/setting process. The second point results in this solution still being overly complex for general use.

I think that the core code could be something more like the following example that largely mirrors your setup (I've got a few mismatched names for expediency), but is more flexible to non-NLR credential getting and setting in addition to simplifying the NLR setup. I would also imagine using the module as `from h2integrate.core.env_tools as api` so we could invoke usage such as `api.get_credentials("NLR_API_EMAIL", "NLR_API_KEY")`.

Another added benefit is that instead of a developer being required to do 3 steps to interact with API settings, the below reduces is it to simply calling `api.set_env_var_from_file(".newapirc", "NEW_EMAIL", "NEW_KEY")`. Later, if needed the credentials could be extracted as `new_email = os.environ["NEW_EMAIL"]` or `new_email = api.get_credentials("NEW_EMAIL")`. Alternatively, they could simply be retrieved as part of the module setup.

To be clear, I don't believe my code needs to be taken exactly as written as it's a prototype, but the straightforwardness
of the solution should be considered to resolve the heart of #765.

```python
import os
import warnings
from pathlib import Path
from h2integrate import ROOT_DIR

HOME_DIR = Path.home()

NREL_DEPRECATION_MSG = (
    "The '{old}' environment variable is deprecated and will be removed in a future release. "
    "Please use '{new}' instead. The nrel.gov API domain has moved to nlr.gov."
)


def read_config(file_path: Path) -> dict:
    """Load any dictionary-like key, value pairs from a configuration file (e.g. .env or .cdsapirc)
    that uses either a ``key:value` or `key=value` format for storing data.

    Args:
        file_path (Path): The full file path and name containing configuration details to be
            extracted.

    Returns:
        dict: Dictionary of key, value pairs found in :py:attr:`file_path`.
    """
    if isinstance(file_path, str):
        file_path = Path(file_path).resolve()
    config = {}
    with file_path.open("r") as f:
        for line in f.readlines():
            if ":" in line:
                sep = ":"
            elif "=" in line:
                sep = "="
            k, v = line.strip().split(sep, 1)
            config[k.strip()] = v.strip()
    return config


def get_credentials(*args: str, file_name: str | None = None) -> dict:
    """Retrieve a series of credentials from a :py:attr:`file_name` in either the home directory
    or H2Integrate root directory.

    Args:
        args (str): Name(s) of the credential(s) that should be retrieved from either
            :py:attr:`file_name` or environment variables.
        file_name (str, optional): The name of a configuration file found in either the H2Integrate
            root directory or the user's home directory that should contain the credential(s) in
            :py:attr:`args`.

    Returns:
        dict: Dictionary of all :py:attr:`args` with values of either a found value or None.
    """
    if file_name is not None:
        if (file_path := (ROOT_DIR / file_name)).is_file():
            config = read_config(file_path)
        elif (file_path := (HOME_DIR / file_name)).is_file():
            config = read_config(file_path)
        return {name: config.get(name) for name in args}
    return {name: os.environ.get(name) for name in args}


def set_env_var_from_file(file_name: str, *args: str):
    """Sets a series of environment variables from a file base in the H2Integrate root directory
    or the user's home directory.

    Args:
        file_name (str): Name of the file found in either the H2Integrate root directory or the
            user's home directory that contains the variables to be set for the computational
            environment.
        args (str): Name(s) of the credentials to retrieve from :py:attr:`file_name`.

    Raises:
        KeyError: Raised if any of :py:attr:`args` wasn't found in :py:attr:`file_name`.
    """
    credentials = get_credentials(*args, file_name=file_name)
    for name, value in credentials.items():
        if value is None and os.environ.get(name) is None:
            raise KeyError(f"No credential for {name} was found in '{file_name}'")
        os.environ[name] = value


def get_nlr_api_credential(which: str) -> str:
    """Get either the NLR API email or key with a fallback for the NREL credentials.

    Args:
        which (str): One of "email" or "key" to indicate which NLR API credential should be
            retrieved.

    Raises:
        ValueError: Raised if an invalid value was passed to :py:attr:`which`.
        KeyError: Raised if neither of the NLR or NREL credentials could be found.
    """
    if which not in ("email", "key"):
        raise ValueError("`which` must be one of 'email' or 'key'.")
    new_name = f"NLR_API_{which.upper()}"
    old_name = f"NREL_API_{which.upper()}"
    credentials = get_credentials(new_name, old_name, file_name=".env")
    if (email := credentials[new_name]) is None:
        if (email := credentials[old_name]) is None:
            msg = (
                f"No credential matching {new_name} or {old_name} found as an environment variable"
                " nor the '.env' configuration file in the user home directory or H2Integrate root"
                " directory."
            )
            raise KeyError(msg)
        warnings.warn(
            NREL_DEPRECATION_MSG.format(old=old_name, new=new_name),
            FutureWarning,
            stacklevel=3,
        )
    return email


def set_nlr_api_credential(which: str):
    """Set an NLR API email or key with a fallback for the NREL credentials.

    Args:
        which (str): One of "email" or "key" to indicate which NLR API credential should be
            retrieved.

    Raises:
        ValueError: Raised if an invalid value was passed to :py:attr:`which`.
    """
    if which not in ("email", "key"):
        raise ValueError("`which` must be one of 'email' or 'key'.")
    os.environ[f"NLR_API_{which.upper()}"] = get_nlr_api_credential(which=which)
```

To answer one of your questions, I think the `env_tools.py` should be `api_credentials.py` or something with either "api" or "credentials" or similar in the name to indicate we're working specifically with API credentials. I think "env_tools.py" would be assumed to be coding environments more often than environment variables. The environment variables is also only one piece of the puzzle API credentials puzzle.

elenya-grant:
Responding to [this comment](https://github.com/NatLabRockies/H2Integrate/pull/798#pullrequestreview-4675433103) from @RHammond2:

First of all, thank you for the very helpful and thoughtful comment, this is very helpful in understanding your vision for resolving Issue #765. I like some of the added functionality that your approach introduces and I think I could get that into this PR (such as getting variables from a dictionary type formatted file). If I'm understanding your proposed code correctly, then I think that some of the existing functionality will be lost and other changes to the code may be needed (and some of this is related to the use of global variables). I'm also concerned about any lost functionality because it could mess with people's workflows. My understanding is that your proposed code would basically be used instead of the functions I have in `env_tools.py` - is that right?

I'm going to try to be concise and I'm sorry if it doesn't make sense! When I originally made the NLR API methods for the resource models, I was trying to make it so that  finding/loading/setting/getting environment variables was as user-friendly as possible and would work for a variety of workflows and ways of setting environment variables while minimizing complexity within the resource models themselves.

Quick summary of the current (before this PR) functionality with the NLR API credentials and my understanding of it:
- If a user sets the environment variable using [the preferred method](https://h2integrate.readthedocs.io/en/latest/getting_started/environment_variables.html#save-environment-variables-with-conda-preferred) then we wouldn't need to use global variables, this is easy because it doesnt require a read-file thing. As long as a user has done this type of set-up, the environment variable value is easily accessed with `os.getenv("NLR_API_KEY")` and the user does not have to run `set_nlr_key_dot_env()` at the start of each script. So when `get_nlr_developer_api_key()` is called within a resource model, it returns the environment variable without ever calling `set_nlr_key_dot_env()`
- If a user has the environment variables in a .env file thats in one of the possible locations (`cwd/".env"`, `ROOT_DIR/".env"`, or `ROOT_DIR.parent/".env`) and the environment variables haven't already been loaded/set (which is usually the case) then the **`set_nlr_key_dot_env()` is called within `get_nlr_developer_api_key()`** to first set the variable before returning it. This also wouldn't require the use of global variables.

The reason we're using global variables (and maybe we could use environment variables instead or something) is to accommodate for other user-workflow/preference cases. For example, a user is using a .env file to define the environment variables and that .env file is not in one of the "default" locations. If they were to try to run H2I with a resource model that needs the NLR_API_KEY, it would not be able to find the API key and the run would fail if their run script looked like this:
```python
from h2integrate import H2IntegrateModel
h2i = H2IntegrateModel("config.yaml")
h2i.setup()
h2i.run()
```

Since the path to their ".env" file cannot be input to the resource models themselves, they need to instead call the setter method ahead of time so that the environment variables are set before the resource models call `get_nlr_developer_api_key()`
```python
from h2integrate import H2IntegrateModel
from h2integrate.resource.utilities.nlr_developer_api_keys import set_nlr_key_dot_env
env_path = "Users/my/nonstandard/place/for/.env"
set_nlr_key_dot_env(path=env_path)
h2i = H2IntegrateModel("config.yaml")
h2i.setup()
h2i.run()
```

Basically, the `get_api_key()` style functions that are called in resource models are set-up so that a user doesnt have to call `set_nlr_key_dot_env` in every run-script if they do the preferred way of setting environment variables or they do it in the way thats instructed in the documentation (maybe except for the .yaml way - Idk how that works). But the `set_nlr_key_dot_env` is available so that folks can run resource models in H2I even if their workflow isn't one of the default supported ways of doing it. 

### Responses to other stuff:

I imagine that environment variables may be used to include more than just API credentials. Environment variables allow for folks to set a) info that shouldn't be shared and b) info that is specific to their workflow. For example, the RESOURCE_DIR can also be set as an environment variable and that'd be used instead of `DEFAULT_RESOURCE_DIR` (Issue #802). This is why I named things with `env` instead of `api`. 

Related to above, I think that the `read_config` method may have issues if someone is using a Windows folder path as the environment variable value (``RESOURCE_DIR = C:\\Users\whatever``) because it'd split the filepath at `:`.

Perhaps some of my push-back is because I'm misinterpreting your comment "it still highly geared towards an NLR credential getting/setting process". Do you mean specific to the NLR_API_KEY and NLR_API_EMAIL variables used for resource data downloads OR do you mean like some workflow that is common at NLR that isn't common elsewhere? I think the only NLR-specific part of my proposed changes is the depreciation method warning.

Let me know if you want to hop on a call and chat about through some of this! I would be down to integrate a form of the `read_config` method you shared! Thanks again for the thoughtful comment!

RHammond2:
Thanks for the all the additional context, @elenya-grant, it's quite helpful to understand where some of the naming is coming from! I'm going to respond below to all your comments/responses as encountered, so apologies if this reads a little akwardly. I'll immediately contradict that sentiment by addressing an item from the end first: let's keep you original thinking with an "env"-focused name instead of my thoughts on the "api"-based name.

The naming of functions in my example code could easily be changed, they are merely what made sense of demonstrating the level of simplicity/limited scope I meant when discussing this refactor previously.

As for global variables, I don't believe they should be used at all, particularly because they are generally discouraged as a best practice. I also didn't realize the existing implementation used them at all until I just double checked. I think if we're trying to be able to just trying to make access easier, then at the bottom of `env_tools.py`, we could do something similar to the following (using my prototyped code naming for consistency). However, I also think that based on a later comment about custom paths, any module needing to ensure the environment variables are set should actually just be running that code rather than tacking an initial setter onto the bottom of `env_tools.py`. This also means importing from `env_tools.py` doesn't run any setters or getters, rendering the following unnecessary, but I also think it's best if no NLR environment variables are set on import.

```python
# env_tools.py
# all preceding code, including function definitions

set_nlr_api_credential()
NLR_API_EMAIL = get_nlr_api_credential("email")
NLR_API_KEY = get_nlr_api_credential("key")

# some other module
from h2integrate import NLR_API_EMAIL, NLR_API_KEY
```

My opinion is also that we can take the lead from how most other APIs enforce the placement of their API credential files (when environment variables are not set) and simply error out as opposed to allowing for arbitrary environment files without users explicitly defining this in a custom setup. As you suggested, users can easily avoid this by configuring their environments using the already provided directions, or simply setting them independently as is shown in your response.

If that's less than desirable, then I think the credential getter I proposed could simply be updated to the following.

```python
def get_credentials(*args: str, file_name: str | None = None, file_path: str | Path | None = None) -> dict:
    """Retrieve a series of credentials from a :py:attr:`file_name` in either the home directory
    or H2Integrate root directory. If `:py:attr:`file_path` is provided, then :py:attr:`file_name`
    and already set environment variables will be ignored. If :py:attr:`file_name` is provided, then
    already set environment variables will be ignored. If neither file options are used, then an
    existing environment variable will be retrieved.

    Args:
        args (str): Name(s) of the credential(s) that should be retrieved from either
            :py:attr:`file_name` or environment variables.
        file_name (str, optional): The name of a configuration file found in either the H2Integrate
            root directory or the user's home directory that should contain the credential(s) in
            :py:attr:`args`.
        file_path (str | Path, optional): The full file path for where the configuration file can be
        found if not using the H2Integrate root directory or user home directory

    Returns:
        dict: Dictionary of all :py:attr:`args` with values of either a found value or None.
    """
    if file_path is not None:
        file_path = Path(file_path).resolve()
        if file_path.is_file():
            config = read_config(file_path)
            return {name: config.get(name) for name in args}
        raise FileNotFoundError(f"Provided `file_path` is invalid: {file_path}")
    if file_name is not None:
        if (file_path := (ROOT_DIR / file_name)).is_file():
            config = read_config(file_path)
        elif (file_path := (HOME_DIR / file_name)).is_file():
            config = read_config(file_path)
        return {name: config.get(name) for name in args}
    return {name: os.environ.get(name) for name in args}
```

To extend this, we could also add an `overwrite` flag to the environment variable setter, so that if the variable already exists, no further code will be run.

```python
def set_env_var(*, overwrite: bool = False, **kwargs: str):
    """Set or overwrite environment variables.
    Args:
        overwrite (bool, optional): Indicator to overwrite existing environment variables provided
            in :py:attr:`kwargs`. Defaults to False.
        kwargs (str): name and value of environment variables to set. If :py:attr:`overwrite` is
            False, the value will be skipped.
    """
    for name, value in kwargs.items():
        if os.environ.get(name) is not None and not overwrite:
            continue
        os.environ[name] = value


def set_env_var_from_file(file_name: str, *args: str, *, overwrite: bool = False):
    """Sets a series of environment variables from a file base in the H2Integrate root directory
    or the user's home directory.

    Args:
        file_name (str): Name of the file found in either the H2Integrate root directory or the
            user's home directory that contains the variables to be set for the computational
            environment.
        args (str): Name(s) of the credentials to retrieve from :py:attr:`file_name`.

    Raises:
        KeyError: Raised if any of :py:attr:`args` wasn't found in :py:attr:`file_name`.
    """
    credentials = get_credentials(*args, file_name=file_name)
    for name, value in credentials.items():
        if value is None and os.environ.get(name) is None:
            raise KeyError(f"No credential for {name} was found in '{file_name}'")
    set_env_var(overwrite=overwrite, **credentials)
```

`read_config` shouldn't have any issues as [Pathlib.home()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.home) is used to create the `HOME_DIR`, which is also showcased in [this SO answer on a Windows environment](https://stackoverflow.com/questions/22947427/getting-home-directory-with-pathlib#33244431).

Regarding my comment about the code being heavily geared towards NLR credentials, this is coming from the fact that
every function takes `varname_new` and `varname_old`, a use case only really needed for the NLR developer API credentials. This need was also why I used `*args` and `**kwargs` (above not in original prototype) so that the NLR-specific getters and setters could still check for both.

elenya-grant:
Responding to [this comment](https://github.com/NatLabRockies/H2Integrate/pull/798#issuecomment-4963900740) from @RHammond2 

Thanks for the additional clarification and more context on your thoughts. Based on your feedback, I see 3-4 immediate changes that I could make to the code to align with your comments:
1. Add the `read_config` function to read environment variables from a file
2. Remove usage of global variables and use environment variables instead
3. Utilize args and kwargs to make the code more generalized (so not as geared towards allowing multiple variable names)
4. Integrate something like the `get_credentials` function 


My thoughts on the generalize flow are listed below:
- There are 2 different ways to load environment variables from a file, one uses `load_dotenv()` and one uses a version of `read_config` (basically a read file). Maybe the use of `load_dotenv` can be removed (`load_dotenv` basically reads a file and then sets the environment variables). These don't throw errors. Lets assume we just use a version of `read_config()` for simplicity and no longer use `load_dotenv`
   - Should this be called for a specific environment variable or should it load all environment variables in that file
   - Should the environment variable(s) be set by calling `set_env_var()` within `read_config()` or should `read_config` just return the dictionary of variables?
- The main `get_credentials()` method does the following logic:
   - checks if the environment variable already exists, if it does, return the value for it
   - call `read_config` for "default" file locations of the environment file. If the environment variable is found, then set the environment variable if it's not set in `read_config()`. 
   - I think `get_credentials()` should have an input toggle to turn on/off error throwing so that it can be used for deprecated key names. For example:
```python
api_key = get_credentials("NLR_API_KEY", strict=False) # don't throw error, just try
if api_key is None:
   api_key = get_credentials("NREL_API_KEY", strict=False) # don't throw error, just try
   if api_key is not None: 
      # the deprecated one exists, so throw depreciation warning
      warnings.warn(DEPRECIATION_MSG.format(old="NREL_API_KEY", new="NLR_API_KEY"), FutureWarning, stacklevel=3)
if api_key is None:
   # API key has not been found - have it throw an error in `get_credentials` call
   api_key = get_credentials("NLR_API_KEY", strict=True)   
 ```

Finally - if setting environment variables is done in `read_config()` then I don't think that `get_credentials()` needs a filepath input argument. If a user has the environment file in a non-default location, then they can set their environment variable(s) in their run script using `read_config()`. If environment variables are set in `get_credentials()` and not set in `read_config` then a user would call `get_credentials()` in their runscript and `get_credentials()` should still have the filepath inputs.

Last question: The default folders we look in for the `.env` file are the CWD, ROOT_DIR, and ROOT_DIR.parent. Should we also check these folders for files named `.nlrrc` and `.h2integraterc`?

File: h2integrate/core/env_tools.py [hunk 1]
```diff
@@ -0,0 +1,111 @@
...
+    if not file_path.is_file():
+        return env_vars
+    with file_path.open("r") as f:
+        for line in f.readlines():
+            if "=" in line:
+                sep = "="
+            elif ":" in line:
+                sep = ":"
+            else:
+                # skip this line
```
RHammond2: I would update this comment to "skip line if invalid separator" or similar (or remove it altogether since the docstring explains the valid separators), otherwise the comment is confusing and makes me think we're skipping `continue`.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    *args: str,
+    file_name: str | None = None,
+    file_path: str | None = None,
+    set_variables: bool = True,
+):
+    """Retrieve a series of credentials from a :py:attr:`file_name` in either the home directory
+    or H2Integrate root directory. If `:py:attr:`file_path` is provided, then :py:attr:`file_name`
+    and already set environment variables will be ignored. If :py:attr:`file_name` is provided, then
+    already set environment variables will be ignored. If neither file options are used, then an
+    existing environment variable will be retrieved.
```
RHammond2:
```suggestion
    """Retrieve a series of credentials from a fully resolved :py:attr:`file_path`,
    a :py:attr:`file_name` located in either the home directory or H2Integrate root directory, or
    from the environment variables.

    If `:py:attr:`file_path` is provided, then :py:attr:`file_name` and already set environment
    variables will be ignored. If :py:attr:`file_name` is provided, then already set environment
    variables will be ignored. If neither file options are used, then an existing environment
    variable will be retrieved.
```
I didn't fully update the description when I added `file_path` to the example`, so here is a more complete description.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+        file_path (str | Path, optional): The full file path for where the configuration file can be
+        found if not using the H2Integrate root directory or user home directory
+        set_variables (bool, optional): If True, set the environment variables if they
+            haven't already been set.
+
+    Returns:
+        dict: Dictionary of all :py:attr:`args` with values of either the value if found.
+    """
+    # Check if the environment variables have already been set
+    env_vars = {name: os.environ.get(name) for name in args if os.environ.get(name) is not None}
```
RHammond2:
```suggestion
    full_env_vars = {arg: None for arg in args}
    env_vars = {name: var for name in args if (var := os.environ.get(name)) is not None}
```

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    remaining_vars = set(args) - set(env_vars)
+    if len(remaining_vars) == 0:
```
RHammond2:
```suggestion
    if not set(args).difference(env_vars):
```

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+        raise FileNotFoundError(f"Provided `file_path` is invalid: {file_path}")
+
+    default_folders = [Path.cwd(), Path.home(), ROOT_DIR, ROOT_DIR.parent]
+    if file_name is None:
+        # If a file_name isn't provided, look for a .env file
+        file_name = ".env"
+
+    for folder in default_folders:
+        if (file_path := (folder / file_name)).is_file():
+            env_vars |= load_env_vars_from_file(file_path)
```
RHammond2:
```suggestion
            env_vars = load_env_vars_from_file(file_path) | env_vars
```
One thing I didn't consider previously, is if we are going to prioritize existing environment variables (suggestions) or any variables contained in a file. The assumption to use the file's variables over the environment variables is probably correct since the user provided the file as opposed to not, but I wanted to raise this just in case.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    env_vars = {name: os.environ.get(name) for name in args if os.environ.get(name) is not None}
+    remaining_vars = set(args) - set(env_vars)
+    if len(remaining_vars) == 0:
+        # All environment variables have already been set
+        return env_vars
+
+    if file_path is not None:
+        file_path = Path(file_path).resolve()
+        if file_path.is_file():
+            env_vars = load_env_vars_from_file(file_path)
```
RHammond2:
```suggestion
            env_vars |= load_env_vars_from_file(file_path)
```
See my comment on `file_name` in case the update order needs to match my suggestion.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    file_name: str | None = None,
+    file_path: str | None = None,
+    set_variables: bool = True,
+):
+    """Retrieve a series of credentials from a :py:attr:`file_name` in either the home directory
+    or H2Integrate root directory. If `:py:attr:`file_path` is provided, then :py:attr:`file_name`
+    and already set environment variables will be ignored. If :py:attr:`file_name` is provided, then
+    already set environment variables will be ignored. If neither file options are used, then an
+    existing environment variable will be retrieved.
+
```
RHammond2, line 115: After reading some more of the implementation, the description should be updated to address that environment variables will always be checked first, and then any file-based environment variables.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    if file_path is not None:
+        file_path = Path(file_path).resolve()
+        if file_path.is_file():
+            env_vars = load_env_vars_from_file(file_path)
+            env_vars_subset = {name: env_vars.get(name) for name in args if name in env_vars}
+            if set_variables:
+                # Set the environment variables
+                set_env_var(overwrite=True, **env_vars_subset)
+            return env_vars_subset
```
RHammond2, line 153:
```suggestion
            return env_vars
```

Wouldn't we want to return all the variables if this is a getter? I agree with the setting logic to avoid overwrites, but this could have a side effect of an existing environment variable not being returned when it's needed simply because a file was also checked for a missing variable.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+        file_name (str, optional): The name of a configuration file found in either the H2Integrate
+            root directory or the user's home directory that should contain the credential(s) in
+            :py:attr:`args`.
+        file_path (str | Path, optional): The full file path for where the configuration file can be
+        found if not using the H2Integrate root directory or user home directory
+        set_variables (bool, optional): If True, set the environment variables if they
+            haven't already been set.
+
+    Returns:
+        dict: Dictionary of all :py:attr:`args` with values of either the value if found.
```
RHammond2:
```suggestion
        dict: Dictionary of all :py:attr:`args` with values of either the value, if found, or None.
```
Sorry, I butchered this one in my comment.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    # Check if the environment variables have already been set
+    env_vars = {name: os.environ.get(name) for name in args if os.environ.get(name) is not None}
+    remaining_vars = set(args) - set(env_vars)
+    if len(remaining_vars) == 0:
+        # All environment variables have already been set
+        return env_vars
```
RHammond2, line 137:
```suggestion
        return full_env_vars | env_vars
```
Accounting for the need to ensure a matching value is returned for all input arguments.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if file_path is not None:
+        file_path = Path(file_path).resolve()
+        if file_path.is_file():
+            env_vars = load_env_vars_from_file(file_path)
+            env_vars_subset = {name: env_vars.get(name) for name in args if name in env_vars}
+            if set_variables:
+                # Set the environment variables
+                set_env_var(overwrite=True, **env_vars_subset)
```
RHammond2, line 149: I don't think the `env_vars_subset` is necessary because we're already doing the overwrite, plus we would want to return all the desired environment variables if checking the existing ones first.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for folder in default_folders:
+        if (file_path := (folder / file_name)).is_file():
+            env_vars |= load_env_vars_from_file(file_path)
+
+    env_vars_subset = {name: env_vars.get(name) for name in args if name in env_vars}
+    if set_variables:
+        # Set the environment variables
+        set_env_var(overwrite=True, **env_vars_subset)
+    return env_vars_subset
```
RHammond2, line 189: Same comment here that I don't believe the subsetting is required.

h2integrate/core/env_tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Args:
+        args (str): Name(s) of the credential(s) that should be retrieved from either
+            :py:attr:`file_name` or environment variables.
+        file_name (str, optional): The name of a configuration file found in either the H2Integrate
+            root directory or the user's home directory that should contain the credential(s) in
+            :py:attr:`args`.
+        file_path (str | Path, optional): The full file path for where the configuration file can be
+        found if not using the H2Integrate root directory or user home directory
+        set_variables (bool, optional): If True, set the environment variables if they
```
RHammond2, line 125: There is a bit of inconsistency in this as written because variables retrieved from `file_name` will update `env_vars`, but not `file_path`. Depending on how you think this should be handled, some of my suggestions should be rejected, or all, and make a fresh batch of changes in a separate commit if that's easier to follow (likely).

h2integrate/resource/utilities/nlr_developer_api_keys.py
```diff
@@ -1,268 +1,81 @@
...
-    value = os.getenv(old_name)
-    if value is not None:
+    if which.lower() not in ("email", "key"):
+        raise ValueError("`which` must be one of 'email' or 'key'.")
+    old_name = f"NREL_API_{which.upper()}"
+    new_name = f"NLR_API_{which.upper()}"
+    nlr_api_vars = get_environment_variables(
+        new_name, old_name, file_path=env_path, set_variables=True
+    )
+    if not bool(nlr_api_vars):
```
RHammond2, line 42: This check may need to be updated based on commentary in `env_tools.py`.

h2integrate/resource/utilities/nlr_developer_api_keys.py
```diff
@@ -1,268 +1,81 @@
...
 
+_ENV_MISSING_MSG = (
+    "{new} (or {old}) has not been set. " "Please set the {new} environment variable."
+)
 
-def _get_env_with_fallback(new_name, old_name):
-    """Get an environment variable by its new name, falling back to the deprecated old name.
 
-    If only the old name is set, a deprecation warning is issued.
+def get_nlr_developer_api_credential(which: str, env_path: str | Path | None = None) -> str:
```
RHammond2: I might just add the `set_path` boolean keyword argument here too, just so the `conftest.py` update makes sense without thinking too hard.

test/conftest.py
```diff
@@ -21,7 +21,7 @@ def pytest_sessionstart(session):
 
     # Set a dummy API key
     os.environ["NLR_API_KEY"] = "a" * 40
-    set_nlr_key_dot_env()
+    get_nlr_developer_api_credential(which="key")
```
RHammond2:
```suggestion
    _ = get_nlr_developer_api_credential(which="key")
```
Just a best practice for unneeded return values.

RHammond2 (commented):
I think this a lot better, @elenya-grant! I left a handful of comments addressing a couple pieces of inconsistency in how found variables in `file_path` vs `file_name` are treated. Mostly this is a question of should a getter return a value for all input arguments or just those with a match? My comments clearly lean towards yes, though there's nothing wrong with that not being the case so long as the documentation matches.

The other major piece is the inconsistency of updating the existing environment variables with those found in a file. When using `file_path` this doesn't occur, but it does in `file_name`. I would lean towards it making more sense to update the existing `env_vars` vs using just those found in the file even though that wasn't my original thought on the matter.

h2integrate/core/env_tools.py
elenya-grant: I don't see I see where `full_env_vars` is used?

h2integrate/core/env_tools.py
elenya-grant: Maybe I don't understand the side-effect. Is this the case you're talking about: an existing environment variable isn't returned because a file was checked for that environment variable and that environment variable wasn't found in the file?

My main reason for not returning a value if it's not found is that I think it's easy to check if a key is in a dictionary rather than checking the value of a key. So - when the getter method is used, its easy to do this:
```python
res = get_environment_variables("MY_KEY","MY_KEYB")
if "MY_KEYB" not in res:
   raise KeyError("missing MY_KEYB")
```

which is my personal preference over this:

```python
res = get_environment_variables("MY_KEY","MY_KEYB")
if (m:=res.get("MY_KEYB",None)) is None:
   raise KeyError("missing MY_KEYB")
```

h2integrate/core/env_tools.py
elenya-grant: updated comment to your suggestion!

h2integrate/core/env_tools.py
elenya-grant: I re-did the doc string for this function to try to be more precise about the step-by-step logic and update it to be more accurate.

h2integrate/core/env_tools.py
elenya-grant: For now, I'm leaving the logic so that it only returns environment variables that were found. (Missing key vs None value). this means that I haven't added the `full_env_vars` bit but I updated the env_vars bit to use the `:=` operator like you suggested.

h2integrate/core/env_tools.py
elenya-grant: I changed the `if` thing to check `if not remaining_vars` but left the `remaining_vars` logic as-is because I think its easier to follow (personally)

h2integrate/core/env_tools.py
elenya-grant: how I've updated it is that:
- If `file_path` is input, then environment variables from the file path are prioritized
- If reading environment variables from one of the default locations, the preexisting environment variables are prioritized

h2integrate/core/env_tools.py
elenya-grant: updated the description!

h2integrate/core/env_tools.py
elenya-grant: I think `env_vars_subset` just makes sure any extraneous environment variables aren't set.

h2integrate/core/env_tools.py
elenya-grant: I think I updated it?

h2integrate/resource/utilities/nlr_developer_api_keys.py
elenya-grant: do you mean the `set_variables` boolean?

test/conftest.py
elenya-grant: done!

docs/getting_started/environment_variables.md
```diff
@@ -1,18 +1,25 @@
...
-To use resource datasets from the NLR developer network, you will need an NLR API key, which can be obtained from:
-    [https://developer.nlr.gov/signup/](https://developer.nlr.gov/signup/).
-
-You will need to set the API key and the email you used to get the API key for downloading resource data from the NLR developer network. The 40 character API key is referred to in following sections as the value for the `NLR_API_KEY` environment variable. The email used to get the API key is referred to in the following sections as the value for the `NLR_API_EMAIL` environment variable.
+(environment_variables:environment-variables)=
+# Environment Variables
+H2Integrate can pull data from datasets accessible with API keys or user-specific tokens. Since API keys and tokens are unique to each user, these are accessed in H2Integrate as environment variables. These environment variables only need to be set to utilize their corresponding functionality. Some environment variables can also used to customize your workflow. The list of environment variables for different types of functionality are listed below:
```
genevievestarke: This feels a little vague to me (how do you get user specific tokens?) but if it's something people generally should know, then I would leave it!

genevievestarke:
```suggestion
H2Integrate can pull data from datasets accessible with API keys or user-specific tokens. Since API keys and tokens are unique to each user, these are set individually and accessed by H2Integrate as environment variables. These environment variables need to be set to utilize their corresponding functionality. Some environment variables can also be used to customize your workflow. The list of environment variables that can be used by H2Integrate are listed below:
```

docs/getting_started/environment_variables.md
```diff
@@ -1,18 +1,25 @@
...
+
+- [NLR Developer Network](environment_variables:nlr_developer)
```
genevievestarke, line 5: I think it would be nice to add what kind of things you would need this for (i.e. wind and solar resource data)

docs/getting_started/environment_variables.md
```diff
@@ -106,3 +113,27 @@ The ".env" file will be looked for in all of the following locations:
     NLR_API_EMAIL='email-for-api-key'
     ```
 3. Save and close the ".env" file.
+
+
+(environment_variables:nlr_developer)=
```
genevievestarke, line 118: This part seems to be in python code formatting (highlighting for and in) rather than markdown formatting. I don't known if that's just a GitHub thing here?

docs/getting_started/environment_variables.md
```diff
@@ -106,3 +113,27 @@ The ".env" file will be looked for in all of the following locations:
...
+# NLR Developer Network Environment Variables
+
+H2Integrate can pull weather resource datasets (e.g. data needed for wind or solar generation) automatically for a user-provided location.
+To use resource datasets from the NLR developer network, you will need an NLR API key, which can be obtained from:
+    [https://developer.nlr.gov/signup/](https://developer.nlr.gov/signup/).
+
+You will need to set the API key and the email you used to get the API key for downloading resource data from the NLR developer network. The 40 character API key is referred to in following sections as the value for the `NLR_API_KEY` environment variable. The email used to get the API key is referred to in the following sections as the value for the `NLR_API_EMAIL` environment variable.
```
genevievestarke:
```suggestion
You will need to set the API key and the email you used to get the API key to download resource data from the NLR developer network. The 40 character API key is referred to in following sections as the value for the `NLR_API_KEY` environment variable. The email used to get the API key is referred to in the following sections as the value for the `NLR_API_EMAIL` environment variable.
```

docs/misc_resources/debugging_environment_variables.md
```diff
@@ -0,0 +1,197 @@
...
+
+
+## Non-standard environment file location or name
+
+If your getting an error for a missing environment variable and your environment variables are stored in a `.env` file, then the `.env` file may not be found in the expected location.
+
+(env_var_debug:get_default_fpaths)=
+### Get the supported default filepaths
+
+The code below shows how to print a list of the default filepaths that H2I will look for the environment file.
```
genevievestarke, line 35: Nice, I love this!!

docs/getting_started/environment_variables.md
genevievestarke, line 81: is it a yaml file for yml file here?

genevievestarke: I know yaml and yml files are the same, but I think we should be consistent throughout this section

genevievestarke (commented):
I think the docs look great @elenya-grant, thank you for adding an amazing debugging section!!

docs/getting_started/environment_variables.md
elenya-grant: its in the H2I root directory. I don't know much about setting environment variables with this method - I think that this method may require that the environment variables are set every time you re-activate your conda environment.

elenya-grant: good catch - I will clean it up!

docs/getting_started/environment_variables.md
```diff
@@ -1,18 +1,25 @@
...
+- [EIA Natural Gas Cost Data](environment_variables:eia_ng)
+    - `EIA_API_KEY`
+- Customized Workflow
+    - `RESOURCE_DIR`
 
 ```{note}
-The old environment variable names ``NREL_API_KEY`` and ``NREL_API_EMAIL`` are still supported
-for backward compatibility, but are deprecated and will be removed in a future release.
-Please migrate to ``NLR_API_KEY`` and ``NLR_API_EMAIL``.
+Tips on debugging environment variable related errors or issues can be found [here](https://h2integrate.readthedocs.io/en/develop/misc_resources/debugging_environment_variables.html)
```
kbrunik: I built the docs locally and this link didn't work, why not use write it like: 

`[here](#env_var_debug:intro)`

and then add `(env_var_debug:intro)=` to the `debugging_environment_variables.md`? This would be more resilient to other branches that folks might be working on/looking at

elenya-grant: the link is broken because its not yet merged into develop, but I could give your way a try! I've had varying success with those types of links when they're in different files but I'll try it!

h2integrate/core/env_tools.py
RHammond2: I'm not sure how I feel about having separate logics, but I'm open to it, especially if there's some context I'm missing.

h2integrate/core/env_tools.py
RHammond2: Fair enough!

h2integrate/core/env_tools.py
RHammond2: Dropping this in favor of an incoming comment that addresses this better.

h2integrate/core/env_tools.py
RHammond2: My bad, I confused myself with this one.

h2integrate/resource/utilities/nlr_developer_api_keys.py
RHammond2: Yes, thank you!

docs/getting_started/environment_variables.md
RHammond2: This is a myst markdown thing for creating cross-referencing tags that are able to be used easily from external files.

docs/getting_started/environment_variables.md
elenya-grant: Updated first sentence to this: "H2Integrate can pull data (such as wind and solar resource data, feedstock prices, etc) from public datasets accessible with API keys or user-specific tokens"

docs/getting_started/environment_variables.md
elenya-grant: changed section header to "Set Environment Variables with .yml file"

docs/getting_started/environment_variables.md
elenya-grant: I think that saying H2I sets the environment variables could be misleading in some weird way. With some of the methods to set environment variables, they're set with conda and accessed by H2I. With the .env file method, environment variables are loaded and set by H2I then accessed with H2I. I did make your other suggested changes.

docs/getting_started/environment_variables.md
elenya-grant: I updated it with your suggestion @kbrunik

docs/getting_started/environment_variables.md
elenya-grant: I suppose the individual sections on API keys and such should explain that but idk - I guess I'm use "token" as a catch-all for any user-specific code/key/etc that may be needed to access publicly available datasets.

h2integrate/resource/utilities/nlr_developer_api_keys.py
elenya-grant: done!

h2integrate/core/env_tools.py
elenya-grant: yeah - I suppose that I don't know either and I don't have strong feelings. I can update it so that the logic is the same (will update this comment when I push that change up)

docs/misc_resources/debugging_environment_variables.md
```diff
@@ -0,0 +1,198 @@
...
+kernelspec:
+  display_name: native-ard-h2i
+  language: python
+  name: python3
+---
+
+(env_var_debug:intro)=
+# Environment Variables: Non-standard Configuration and Debugging Issues
+
+If errors are encountered with models that use environment variables, such as a missing API key or other credential, this may be because the environment variables were set using a non-preferred method or your workflow may be different. The different approaches to setting environment variables is documented [here](https://h2integrate.readthedocs.io/en/latest/getting_started/environment_variables.html). Some of the non-preferred methods to set environment variables may require that environment variables be reset when your environment is activated. If you're setting environment variables with a .env file, then errors may occur because the file is not found.
```
RHammond2:
```suggestion
If errors are encountered with models that use environment variables, such as a missing API key or other credential, this may be because the environment variables were set using a non-preferred method or your workflow may be different. The different approaches to setting environment variables is documented [in the getting started guides](#environment_variables:environment-variables). Some of the non-preferred methods to set environment variables may require that environment variables be reset when your environment is activated. If you're setting environment variables with a .env file, then errors may occur because the file is not found.
```

docs/misc_resources/debugging_environment_variables.md
```diff
@@ -0,0 +1,198 @@
...
+### Determine if environment file is in supported default filepath
+
+If you want to determine whether your environment file is placed in one of the valid locations, the code below will tell you if your environment file is found in any of the default directories:
+
+```{code-cell} ipython3
+from pathlib import Path
+
+from h2integrate import ROOT_DIR
+
+default_directories = [ROOT_DIR, ROOT_DIR.parent, Path.cwd(), Path.home()]
```
RHammond2, line 60: These imports and defaults are already created in the previous cell, so are unnecessary here. Redundant definitions in cod examples is fairly uncommon unless it's putting an entire workflow in one place at the very end.

docs/misc_resources/debugging_environment_variables.md
```diff
@@ -0,0 +1,198 @@
...
+## Debugging a missing environment variable
+For any method of setting environment variables, you can use the code below to determine if your environment variables are being loaded properly or being set.
+```{note}
+If you're setting environment variables with an environment file in a non-standard location or filename, refer to the section above on how to specify the inputs to the `get_environment_variables` function.
+```
+
+Below shows an example of checking if the `NLR_API_KEY` and `NLR_API_EMAIL` environment variables are being **found** and what the values are of the variables that are found:
+
+```python
+import os
```
RHammond2:
```suggestion
```

docs/misc_resources/debugging_environment_variables.md
```diff
@@ -0,0 +1,198 @@
...
+from pathlib import Path
+
+from h2integrate.core.env_tools import get_environment_variables
+
+# Check if environment variables are being set
+get_environment_variables("NLR_API_KEY","NLR_API_EMAIL",set_variables=True)
+
+nlr_api_key = os.environ.get("NLR_API_KEY", None)
+nlr_api_email = os.environ.get("NLR_API_EMAIL", None)
```
RHammond2, line 145: `None` is the default value for `.get()` when there isn't a match, so it's not really needed.

RHammond2, line 145: How come we aren't making use of the values returned (or not) from the `get_environment_variables()`?

h2integrate/core/test/conftest.py
```diff
@@ -35,7 +35,7 @@ def pytest_sessionstart(session):
     # does not mess with a user's environment
     # Set a dummy API key
     os.environ["NLR_API_KEY"] = "a" * 40
-    set_nlr_key_dot_env()
+    _ = get_nlr_developer_api_credential(which="key", set_vars=True)
```
RHammond2: We should actually be treating this step similar to the way `RESOURCE_DIR` was updated to be by using a temporary environment variable for the original value, and then restoring the original in `pytest_sessionfinish`

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+    env_vars = load_env_vars_from_file(env_path)
+    with subtests.test("Empty line and mixed separators (TEST_CREDENTIAL)"):
+        assert env_vars["TEST_CREDENTIAL"] == "my_credential_value"
+    with subtests.test("Empty line and mixed separators (TEST_CREDENTIAL_B)"):
+        assert env_vars["TEST_CREDENTIAL_B"] == "testing@yahoo.fake"
+    with subtests.test("Empty line and mixed separators (TEST_CREDENTIAL_D)"):
+        assert env_vars["TEST_CREDENTIAL_D"] == "testingValue"
+    with subtests.test("Empty line and mixed separators (skipped sentence line)"):
+        extra_vars = set(env_vars) - {"TEST_CREDENTIAL", "TEST_CREDENTIAL_B", "TEST_CREDENTIAL_D"}
+        assert len(extra_vars) == 0
```
RHammond2:
```suggestion
    env_path = temp_dir / ".env"
    with env_path.open("w+") as file:
        file.write("TEST_CREDENTIAL=my_credential_value\n")  # = with no spaces
        file.write("TEST_CREDENTIAL_B=testing@yahoo.fake \n")  # = with spaces
        file.write("TEST_CREDENTIAL_C_IS_A_SENTENCE\n")  # should be skipped
        file.write("TEST_CREDENTIAL_D : testingValue\n")  # : with spaces

    env_vars = load_env_vars_from_file(file_path=env_path)
    assert len(env_vars) == 3
    assert env_vars["TEST_CREDENTIAL"] == "my_credential_value"
    assert env_vars["TEST_CREDENTIAL_B"] == "testing@yahoo.fake"
    assert env_vars["TEST_CREDENTIAL_D"] == "testingValue"
```

The test as written covers all the bases well, but is a bit too verbose and hard to follow for what is a pretty straightforward test.

RHammond2: It could even be worth adding a consecutive delimiter example like the below modification to highlight what happens when users don't pay attention.
```python
file.write("TEST_CREDENTIAL_E :: another_credential\n")
assert len(env_vars) == 4
assert env_vars["TEST_CREDENTIAL_E"] == ": another_credential"

File: h2integrate/core/test/test_env_tools.py [hunk 29]
```diff
@@ -0,0 +1,231 @@
...
+        assert env_vars["TEST_CREDENTIAL"] == "new_value"
+
+    with subtests.test("NLR_API_KEY returned"):
+        assert env_vars["NLR_API_KEY"] == "a" * 40
+
+    with subtests.test("TEST_CREDENTIAL environment variable (ending)"):
+        assert os.environ.get("TEST_CREDENTIAL") == "new_value"
+
+    with subtests.test("NLR_API_KEY environment variable (ending)"):
+        assert os.environ.get("NLR_API_KEY") == "a" * 40
```
RHammond2, line 129: In general, there could be fewer subtests to make it easier to read the tests. For example, I would simplify this test to the following subtests: "Check starting environment variables," "Check retrieved environment variables," and "Check original environment variables are unmodified." Most of the tests follow a similar workflow, and this simplification adds a lot to legibility without losing any debugging information.

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+        assert os.environ.get("TEST_CREDENTIAL_B") is None
+
+    current_dir = Path.cwd()
+
+    os.chdir(temp_dir)
+
+    env_path = temp_dir / "myfile.env"
+
+    env_file_txt = f"TEST_CREDENTIAL_A={temp_dir}\nTEST_CREDENTIAL_B=howdyFolks\n"
+    env_file_txt += "TEST_CREDENTIAL_C=i<3H2I\n"
```
RHammond2, line 187: lol

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+
+    with env_path.open("w+") as file:
+        file.write(env_file_txt)
+
+    env_vars = get_environment_variables(
+        "TEST_CREDENTIAL_A", "TEST_CREDENTIAL_B", file_name="myfile.env", set_variables=False
```
RHammond2:
```suggestion
        "TEST_CREDENTIAL_A", "TEST_CREDENTIAL_B", file_name=env_path, set_variables=False
```
I would remove the `os.chdir` calls in addition to this as well since we really shouldn't be changing directories without fail safes in place to ensure the directory is changed back in case of a failure.

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+    # were not set as environment variables
+    with subtests.test("TEST_CREDENTIAL_A not set"):
+        assert os.environ.get("TEST_CREDENTIAL_A") is None
+
+    with subtests.test("TEST_CREDENTIAL_B not set"):
+        assert os.environ.get("TEST_CREDENTIAL_B") is None
+
+
+@pytest.mark.unit
+def test_get_environment_variables_from_default_folder(subtests, temp_dir):
```
RHammond2, line 170: Same comment about grouping subtests applies here, though there are more categories.

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+        assert env_vars["TEST_CREDENTIAL_C"] == "i<3H2I"
+
+    # were not set as environment variables
+    with subtests.test("TEST_CREDENTIAL_B set as environment variable"):
+        assert os.environ.get("TEST_CREDENTIAL_B") == "byeFolks"
+
+    with subtests.test("TEST_CREDENTIAL_C set as environment variable"):
+        assert os.environ.get("TEST_CREDENTIAL_C") == "i<3H2I"
+
+    os.chdir(current_dir)
```
RHammond2:
```suggestion
    env_path = temp_dir / ".env"

    env_file_txt = f"TEST_CREDENTIAL_A={temp_dir}\nTEST_CREDENTIAL_B=byeFolks\n"
    env_file_txt += "TEST_CREDENTIAL_C=i<3H2I\n"

    with env_path.open("w+") as file:
        file.write(env_file_txt)

    with chdir(temp_dir):
        env_vars = get_environment_variables(
            "TEST_CREDENTIAL_B", "TEST_CREDENTIAL_C", set_variables=True
        )

    with subtests.test("TEST_CREDENTIAL_B value"):
        assert env_vars["TEST_CREDENTIAL_B"] == "byeFolks"

    with subtests.test("TEST_CREDENTIAL_C value"):
        assert env_vars["TEST_CREDENTIAL_C"] == "i<3H2I"

    # were not set as environment variables
    with subtests.test("TEST_CREDENTIAL_B set as environment variable"):
        assert os.environ.get("TEST_CREDENTIAL_B") == "byeFolks"

    with subtests.test("TEST_CREDENTIAL_C set as environment variable"):
        assert os.environ.get("TEST_CREDENTIAL_C") == "i<3H2I"
```

The imports will now also need `from contextlib import chdir`, but this will safely change the directory to achieve the desired behavior in this test.

h2integrate/core/test/test_env_tools.py
```diff
@@ -0,0 +1,231 @@
...
+        assert os.environ.get("TEST_CREDENTIAL_A") is None
+
+    with subtests.test("TEST_CREDENTIAL_B not set"):
+        assert os.environ.get("TEST_CREDENTIAL_B") is None
+
+    os.chdir(current_dir)
+
+
+@pytest.mark.unit
+def test_get_environment_variables_from_default_cwd(subtests, temp_dir):
```
RHammond2, line 214: Same comment about grouping subtests.

h2integrate/resource/test/conftest.py
```diff
@@ -70,7 +70,7 @@ def pytest_sessionstart(session):
 
     # Set a dummy API key
     os.environ["NLR_API_KEY"] = "a" * 40
-    set_nlr_key_dot_env()
+    _ = get_nlr_developer_api_credential(which="key", set_vars=True)
```
RHammond2: Same comment from the core/ conftest.

test/conftest.py
```diff
@@ -21,7 +21,7 @@ def pytest_sessionstart(session):
 
     # Set a dummy API key
     os.environ["NLR_API_KEY"] = "a" * 40
-    set_nlr_key_dot_env()
+    _ = get_nlr_developer_api_credential(which="key", set_vars=True)
```
RHammond2: Same comment from the core/ conftest.

docs/misc_resources/debugging_environment_variables.md
elenya-grant: I see what you're saying. The intention is that each of these blocks could be copy-pasted to a new python file for a user to run for debugging - that is why there is redundant code. I expect this page to be most useful to folks who are potentially a bit newer to python - so I'm trying to keep it very clear the exact chunk of code they can run to debug stuff. Let me know if you still think the redundant code should be avoided.

I could also make this a python formatted code cell rather than an executable markdown code cell (so it doesn't actually run when the docs build) but I wanted to have it run so that if any future changes would break this code, then we remember to change it.

docs/misc_resources/debugging_environment_variables.md
elenya-grant: This is intended to be a really comprehensive debugging workflow. In the code block above, a user can determine if the environment variables are being **found** properly. If the variables are being found, but for some reason the user is still getting errors with (for example) API calls, they can check what the values that the environment variables values are being **set** to. 

This can be helpful for debugging if someone encounters a weird case where the environment variables are being found properly but not being set OR the value that the environment variables are being _set to_ is different than the value the environment variables are _found_ as.

h2integrate/core/test/conftest.py
elenya-grant: good point. I will work on that.

docs/misc_resources/debugging_environment_variables.md
elenya-grant: I think either way works. I suppose that I prefer using the `os.environ.get` method so that a) users learn about this function if they weren't aware of it before and b) a user can figure out if theres some weird bug in `get_environment_variables` in setting the environment variables (maybe its being set to some value in `get_environment_variables` but its getting modified somewhere else so that when `os.environ.get` is called the value is different.

h2integrate/core/test/test_env_tools.py
elenya-grant: `file_name` is supposed to be a _filename_, `file_path` is the input for _file paths_. 

The reason `file_name` is "myfile.env" rather than env_path and why we're changing directories is to test the case where the environment file is loaded from one of the default folders (one of those is the CWD). I didn't want to make this folder in the ROOT_DIRECTORY or the person's home directory. I updated this to use the `from contextlib import chdir` approach you suggested for a different test! So theres no more `os.chdir` in this test.

The test `test_get_environment_variables_from_filepath` tests when `file_path` input is set to `env_path`.

h2integrate/core/test/test_env_tools.py
elenya-grant: I'm making a similar change to `test_get_environment_variables_from_default_folder` to use this form of `chdir`

h2integrate/core/test/test_env_tools.py
elenya-grant: yeah - I agree. I've just gotten feedback from other folks in the past that like to see a subtest per assert statement. Which, especially here, does make the tests harder to follow but would make it easier to debug if a subtest fails. I tried to add in comments to the tests to show the workflow and hopefully make understanding the tests a bit easier

h2integrate/core/test/test_env_tools.py
elenya-grant: done!

h2integrate/core/test/test_env_tools.py
elenya-grant: same response about I hear ya but I kept the overly verbose subtests and tried to add in comments to make it more clear.

h2integrate/core/test/test_env_tools.py
elenya-grant: I made this change

h2integrate/resource/test/conftest.py
elenya-grant: done

test/conftest.py
elenya-grant: done

h2integrate/core/test/conftest.py
elenya-grant: done!

docs/misc_resources/debugging_environment_variables.md
elenya-grant: yeah - I suppose my justification is again that I'm trying to make this extremely user-friendly (and thinking about users who may be newer to python) so I'm trying to give them more clues about how these functions work even if it is a bit unnecessary.

h2integrate/core/env_tools.py
elenya-grant: I think the only odd use-case I can think of is if the user has a file in the easy-to-miss place (like the home directory) that has old or other environment variables and they aren't intending to use those. I wonder if any other folks @johnjasa @kbrunik have any opinions or thoughts here? I don't have strong feelings - just want to see if other folks are leaning any kind of way.

elenya-grant: Okay! I did not change the logic. I am still having it prioritize environment variables that are set. I did this because I'm worried about cases where tests for the NLR resource models pass locally and fail on the CI and it's hard to debug. 

If someone changes the lat, lon, or resource year somewhere that would require a new resource file download (that isn't in the H2I resource files folder). How it is now, if they run the test (without having the new resource file downloaded beforehand) the test will fail because the NLR API key is invalid. If I change the logic here and a user has a .env file in the H2I root directory, then the test won't fail because when the `get_nlr_developer_api_credential` is called in the `conftest.py` file, their (valid) API key from their environment file will overwrite the API key set in `conftest.py`. 

To hopefully make all parties reasonably happy - I added in a user warning that gets thrown if an environment variable is defined twice and has mismatched values. This is an edge-case situation, it will only happen if all the below things are true:
- only some of the environment variables are already set (not loaded from file)
- one of the environment variables that was already set is also defined in a file and those two environment variables have different values between the two ways its been set.

I think we can change this logic in the future too. This kind of thing is honestly best tested when people are using H2I with their specific environment variable setup.

RHammond2: That's good enough for me, thanks for the update!

docs/misc_resources/debugging_environment_variables.md
RHammond2: If the cell doesn't need to run, then I'd say make it a python code block rather than executable cell like most of this documentation page, but this explanation works for me.

docs/misc_resources/debugging_environment_variables.md
RHammond2: My opinion is that's overkill given it's just how dictionaries operate in Python, but I'll let you decide since it's not wrong by any means.

RHammond2: This works for me.

docs/misc_resources/debugging_environment_variables.md
elenya-grant: done! pushed it up!

h2integrate/core/test/test_env_tools.py
RHammond2: It's possible the changes in my below comment weren't pushed, but I do think this should just be a basic test at the beginning that simply checks the length of the returned `env_vars` as this is a bit of a convoluted way to do that.

h2integrate/core/test/test_env_tools.py
RHammond2: We've talked about this offline, and leave it to be addressed as a larger change for later. The summary of the conversation being that this approach often adds unnecessary verbosity without adding benefits for readability, maintenance, or debugging, and is more or less an abuse of what makes subtests good.

RHammond2 (approved):
Great work, @elenya-grant! I think once you've addressed everything we've talked about, this is good from my end since we cleared everything up in a discussion offline.

h2integrate/core/test/test_env_tools.py
elenya-grant: I think these changes should be pushed up. In `test_load_env_vars_from_file` is has `assert len(env_vars)=4` (because I added another environment variable to the test) instead of the weird version I had before.

johnjasa (approved):
This is fantastic, thank you for this @elenya-grant and for iterating with @RHammond2!

PR #774, opened 2026-06-02
Allow for connections including slice specifications

jaredthomas68:
# Allow connections using src_indices/slices

As noted in #763, there are times when it can be helpful to connect only part of a source variable to a target variable in the OpenMDAO connections. OpenMDAO allows for this, as described in the [docs](https://openmdao.org/newdocs/versions/latest/features/core_features/working_with_groups/connect.html). This PR extends the connection specifications in H2Integrate to allow users to connect from part of an array variable to a target variable.

## Section 1: Type of Contribution

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [ ] Fill out TODO list steps
- [ ] Describe requested feedback from reviewers on draft PR
- [ ] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [ ] Gather feedback on draft PR
- [ ] Implement suggested changes
- [ ] Add doc page
- [ ] Add or update example using this feature

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:
- I am mostly looking for feedback on the user interface for slice connections

**Other feedback**:

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Added tests for new functionality or bug fixes
- [x] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [x] Examples have been updated (if applicable)
- [x] `CHANGELOG.md`
  - [x] At least one complete sentence has been provided to describe the changes made in this PR
  - [x] After the above, a hyperlink has been provided to the PR using the following format:
    "A complete thought. [PR XYZ]((https://github.com/NatLabRockies/H2Integrate/pull/XYZ)", where
    `XYZ` should be replaced with the actual number.

## Section 4: Related Issues

- Fully resolves #763 

### Section 5.1: New Files
- `h2integrate/core/test/test_split_indices.py`: tests for `_split_indices_from_connected_parameter_definition`
[...]

PR #799, opened 2026-07-06
Multi-year sell-price in ProFAST NPV

elenya-grant:
# Multi-year sell-price in ProFAST NPV

Updated `ProFastNPV` so that the commodity sell price input is the same length as the plant life. This enables users to be able to set a sell price for the NPV calculation that is different per year. 

Shoutout to @vijay092 for bringing this up!

**Please look at Section 2 for specific questions for reviewers**

**(Out-dated) Question/Note**: until PR #774 is merged, this change would prevent someone from being able to connect a single value to the sell price input (such as `[finance_subgroup_electricity`, `finance_subgroup_electricity_NPV`, `LCOE`, .`commodity_sell_price_electricity`]). If we want to support this type of connection before #774 is merged in, then I could change the logic (it will be not as clean) so that the sell-price can be input with either a shape of 1 or a shape equal to the plant life. Please let me know whether the handling to enable this type of connection would be valuable in this PR, or if that doesn't need to be included here because it'll be easily enabled with PR #774.

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Update documentation for profast npv model (`docs/finance_models/ProFastNPV.md`)

**Structural feedback**:

**Implementation feedback**:
When we call ProFAST to calculate the NPV (`pf.cash_flow(price=sell_price)`), we can either input the sell price as a dictionary (keys as the operational year) or a list. If we input it as a dictionary, ProFAST does not apply the `params['commodity']['escalation']` to the price. If we input it as a list or scalar, the ProFAST will apply the `params['commodity']['escalation']` to the price. 

What makes sense for most users? Have them apply any inflation needed or not?

**Other feedback**:

- [x] Documentation
- [ ] `CHANGELOG.md`

### Section 5.1: New Files
N/A

- `h2integrate/finances/profast_npv.py`
  - `ProFastNPV.setup()`: made the length of the commodity sell price input equal to plant life. Also added more error messages when checking the commodity sell price input.
  - `ProFastNPV.compute()`: added padding to the price profile input so that its the proper length for the profast `cash_flow` calculation
- `h2integrate/finances/test/test_profast_npv.py`: added two tests:
   - `test_profast_npv_multi_year_sell_price`: tests that NPV is output as expected if a list of sell prices are input
   - `test_profast_npv_multi_year_error`: tests that an error is thrown if sell price is the wrong length
- `h2integrate/finances/profast_base.py`
   - `BasicProFASTParameterConfig.from_dict()`: updated so that if a user specifies an escalation rate for a specific parameter, that it isn't over-written to be the inflation rate. This allows users to specify a different escalation rate for the commodity vs other parameters like incidental revenue, road tax, etc.


h2integrate/finances/profast_npv.py
```diff
@@ -65,9 +64,22 @@ def setup(self):
 
         super().setup()
 
+        if isinstance(self.commodity_sell_price, float | int):
+            if self.commodity_sell_price is None:
+                raise ValueError("commodity_sell_price is missing as an input")
```
vijay092: None is neither a float nor an int so if the `self.commodity_sell_price` is actually None, it won't enter the previous condition.

vijay092: Also worth adding as a test

h2integrate/finances/test/test_profast_npv.py
```diff
@@ -332,3 +332,87 @@ def test_profast_npv_nonstandard_price_units(
...
+    prob = om.Problem()
+    profast_inputs_no2["commodity_sell_price"] = [0.07] * 30
+    plant_config = {
+        "plant": {
+            "plant_life": 30,
+        },
+        "finance_parameters": {"model_inputs": profast_inputs_no2},
+    }
+    pf = ProFastNPV(
+        driver_config={},
```
vijay092, line 352: Both test fixtures use inflation_rate: 0.0. Can you add a case with nonzero inflation to confirm the per-year array isn't getting escalated a second time by ProFAST's own calculation? If users are meant to supply already-nominal per-year prices, we may need to force infation_rate = 0 when an array is passed.

vijay092: Also, could we clarify in the docs whether `commodity_sell_price` should be real or nominal dollars? Per our discussion with @jaredthomas68, we need to set nominal prices with inflation_rate = 0 (since ProFAST would otherwise apply its own escalation on top). Worth spelling that out explicitly.

h2integrate/finances/profast_npv.py
vijay092, line 44: Could we add here that `commodity_sell_price` could be an array?

h2integrate/finances/profast_npv.py
```diff
@@ -65,9 +64,22 @@ def setup(self):
 
         super().setup()
 
+        if isinstance(self.commodity_sell_price, float | int):
```
vijay092: Nit: Good to check for an edge case when use user passes in [x] (a single element list).

vijay092 (commented):
Thank you so much, Elenya!! This looks great to me! I had some minor comments about inflation rate.

h2integrate/finances/profast_npv.py
elenya-grant: good catch! thank you!

h2integrate/finances/profast_npv.py
elenya-grant: Clarifying question: do you think it should be OK if a user passes in a single-element list? So - if it's a list of length 1 then don't throw an error?

h2integrate/finances/profast_npv.py
elenya-grant: done!

h2integrate/finances/profast_npv.py
elenya-grant: I fixed the logic and added in a test for it `test_profast_npv_missing_sell_price`

h2integrate/finances/test/test_profast_npv.py
elenya-grant: >  Can you add a case with nonzero inflation to confirm the per-year array isn't getting escalated a second time by ProFAST's own calculation

I played around with ProFAST using a 2% inflation rate and have some info in case its useful:
- `pf.cash_flow(price=0.07)` and `pf.cash_flow(price=[0.7]*30)` result in the same NPV 
- The NPV with an inflation rate of 2% is about 2.2x the NPV with 0 inflation (both using a sell price of 0.7)

I think ProFAST **does escalate the sell price**, it basically does this:
```python
analysis_length = plant_life + (installation_period_months/12) 
analysis_years = np.arange(0, analysis_length + 1)
sales_price = commodity_sell_price*(1.0+inflation)**(analysis_years-1)
revenue_from_sales = sales_price*annual_production
```

Aka - the commodity sell price should provided
-  in non-escalated dollars (I think this means **real** - if I'm interpreting Jared's helpful doc page properly)
- in the same cost year as set in `plant_config["finance_parameters"]["cost_adjustment_parameters"]["target_dollar_year"]` (the commodity_sell_price does not go through AdjustedCapexOpexComp)

I would be happy to update the docs! @jaredthomas68  - can you confirm that the commodity_sell_price should be input in **real** dollars?

elenya-grant: Can yall (@vijay092 and @jaredthomas68) tell me whether this is True or false:
- if using an inflation rate of 0 in ProFAST, then the commodity sell price should be provided in **nominal** dollars?
- if using a nonzero inflation rate in ProFAST, then the commodity sell price should be provided in **real** dollars since ProFAST will escalate the price?

Can y'all look at the test `h2integrate/finances/test/test_profast_npv.py::test_profast_npv_with_inflation` and let me know what y'all think?

h2integrate/finances/profast_npv.py
vijay092: I think it is ok to pass a list of length one. We can use the first element!

elenya-grant: understood! I think I'm torn on this. I see how it can be more user-friendly but I think that it maybe complicates the "checking" logic in a way that may make the "checking" code slightly harder to understand. If other folks have opinions on this - would love some additional thoughts (@johnjasa, @jaredthomas68)

jaredthomas68: I think it is fine either way, so maybe go the easy way (just support `float` and `int` for now).

h2integrate/finances/test/test_profast_npv.py
elenya-grant: I added stuff to the documentation here `docs/finance_models/ProFastNPV.md` if you want to let me know your thoughts on it!

elenya-grant: okay - talked to Jared and he explained some stuff to me. Anyway - main remaining question for @vijay092  and @jaredthomas68  is:

When we call ProFAST to calculate the NPV (`pf.cash_flow(price=sell_price)`), we can either input the sell price as a dictionary (keys as the operational year) or a list. If we input it as a dictionary, ProFAST does not apply the `params['commodity']['escalation']` to the price. If we input it as a list or scalar, the ProFAST will apply the `params['commodity']['escalation']` to the price. 

What makes sense for most users? Have them apply any inflation needed or not?

vijay092: I think I prefer the latter option. If ProFAST applies the inflation rate, then I won't have to do any pre-processing

kbrunik: I agree with @vijay092, I think it would make sense to have it handled the same way if you have multiple prices or just a single price. It's also easier if ProFAST handles the escalation since it can do escalation for OpEx and other aspects of the finances.

docs/finance_models/ProFastNPV.md
```diff
@@ -24,6 +24,14 @@ finance_parameters:
     cost_year_adjustment_inflation: 0.025 # used to adjust costs for technologies to target_dollar_year
 ```
 
+```{note}
+`commodity_sell_price` can be input as a scalar value (as shown above) or as a list that has the same length as the `plant_life` specified in the plant configuration file. Please refer to the documentation on [different types of financial analysis](https://h2integrate.readthedocs.io/en/latest/finance_models/financial_analyses.html) to determine how to input the `commodity_sell_price` across different years.
+```
+
+```{important}
```
vijay092, line 31: Thanks for adding docs!

h2integrate/finances/test/test_profast_npv.py
```diff
@@ -332,3 +333,194 @@ def test_profast_npv_nonstandard_price_units(
...
+        )
+
+    nominal_price = np.concatenate(
+        [np.zeros(4), prob.get_val("pf.sell_price_electricity_no2", units="USD/(kW*h)")]
+    )
+
+    # Remove inflation from nominal price
+    real_price = nominal_price / ((1.0 + price_escalation) ** (years - 1))
+
+    prob.set_val("pf.sell_price_electricity_no2", real_price[4:], units="USD/(kW*h)")
```
vijay092, line 519: Awesome, thanks for this test!

vijay092 (approved):
Looks good to me! Thanks so much, Elenya!

h2integrate/finances/profast_npv.py
```diff
@@ -112,6 +127,8 @@ def compute(self, inputs, outputs):
         else:
             pf = self.populate_profast(inputs)
 
-        outputs[f"NPV_{self.output_txt}"] = pf.cash_flow(
-            price=inputs[f"sell_price_{self.output_txt}"][0]
+        non_op_Nyears = int(np.ceil(self.params.installation_time / 12) + 1)
+        sell_profile = np.concatenate(
```
kbrunik, line 131: Nice! Would have been easy to miss accounting for the non-operable years (installation time).

kbrunik (approved):
This looks great to me and adds more flexibility in defining future prices which is awesome! Thanks @elenya-grant

h2integrate/finances/test/test_profast_npv.py
jaredthomas68: I was a little slow weighing in here. If we are going to apply the inflation to a list of prices, then having both the option of having the inflation applied to the sale price or not is important because backing inflation out of historical prices can be tricky. I guess as it is will work, but it is dangerous to have it be different between the financial models. It should be ok if the documentation and doc strings are very clear in both.

PR #805, opened 2026-07-20
Thermodynamic NGCT model

cfrontin:
# Thermodynamic models for NGCT (w/ support for future `combustion_machines`)

The performance of combustion machines, including both combustion turbines, combined cycle turbines, and reciprocating internal combustion engines, has significant dependence on ambient conditions (real time _and_ design-specified).

These models use standard thermodynamic cycles to increase the fidelity of ICE combustion systems in H2I. These are critical for understanding complementarity of renewables and fuel combustion systems in hybrid energy generation, where ambient conditions determine the performance of nominally firm power sources as well as renewables, often in mutually complimentary ways.

Demonstration of code addendum showing diurnal and seasonal performance cycles as a function of ambient pressure and temperature.

## Section 1: Type of Contribution

- [x] Feature Enhancement
  - [ ] Framework
  - [x] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [x] Fill out TODO list steps
- [ ] Describe requested feedback from reviewers on draft PR
- [ ] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [ ] remove debug runner script
- [ ] add upgraded cost model w/ fuel-consumption driven costs
- [ ] develop example case(s)

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [ ] Added tests for new functionality or bug fixes
- [ ] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [ ] Documentation
  - [ ] Docstrings are up-to-date
  - [ ] Related `docs/` files are up-to-date, or added when necessary
  - [ ] Documentation has been rebuilt successfully
  - [ ] Examples have been updated (if applicable)
- [ ] `CHANGELOG.md`
  - [ ] At least one complete sentence has been provided to describe the changes made in this PR
  - [ ] After the above, a hyperlink has been provided to the PR using the following format:
    "A complete thought. [PR XYZ]((https://github.com/NatLabRockies/H2Integrate/pull/XYZ)", where
    `XYZ` should be replaced with the actual number.

[...]

PR #773, opened 2026-05-28
Converter PLM dispatch

jaredthomas68:
# Converter PLM control

This PR adds peak load management dispatch for converters. It handles two load profiles, one representing an upstream (grid) load and the other representing either a local load that must be met or a price profile. The converter will be dispatched when either of the demand profiles exceeds their respective peak limits. If a price profile is given, then dispatching will only occur when the price exceeds the provided cutoff.

This PR also provides minor updates to the fuel cell model to determine VarOpEx.

## Section 1: Type of Contribution

- [x] Feature Enhancement
  - [x] Framework
  - [x] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [ ] Bug Fix
- [x] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [x] Fill out TODO list steps
- [x] Describe requested feedback from reviewers on draft PR
- [x] Complete Section 7: New Model Checklist (if applicable)

### TODO:
- [x] Add tests
- [x] Finish example
- [x] Add documentation

### Type of Reviewer Feedback Requested (on Draft PR)

Ready for any and all feedback. Let's get this PR polished and merged.

**Structural feedback**:
~~Looking for high-level feedback at this point from any who wish to give it.~~

**Implementation feedback**:

**Other feedback**:

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Added tests for new functionality or bug fixes
- [x] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
  - [x] Examples have been updated (if applicable)
- [x] `CHANGELOG.md`
  - [x] At least one complete sentence has been provided to describe the changes made in this PR
  - [x] After the above, a hyperlink has been provided to the PR using the following format:
    "A complete thought. [PR XYZ]((https://github.com/NatLabRockies/H2Integrate/pull/XYZ)", where
    `XYZ` should be replaced with the actual number.

### Section 5.1: New Files
- plm_openloop_converter_controller.py
  - `PeakLoadManagementHeuristicOpenLoopConverterControllerConfig`: New config plm converter control
[...]

PR #808, opened 2026-07-21
Demand Response Optimization for Battery Energy Storage (Stage 2)

abhineet-gupta:
# Demand Response Optimization for Battery Energy Storage (Stage 2)

This PR is a continuation of [PR 679](https://github.com/NatLabRockies/H2Integrate/pull/679).
It adds a Pyomo optimization based controller for the BESS system.
The controller optimizes the battery dispatch to minimize the Co-Op's expenditure while allowing demand response  capabilities based on G&T requirements during peak window.

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

**Structural feedback**:
Do you agree with the terminology used in describing this approach in documentation.

**Implementation feedback**:
Is [this](https://github.com/abhineet-gupta/AG_H2Integrate/blob/bd1701a8a94e7c7ef17432e631034e1083f183ef/.github/workflows/ci.yml#L62)  the correct way to add highs solver for pyomo to github workflow. It has already been added to `environment.yml`.
**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

- `h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py`
    - Modified implementation under `PeakLoadManagementOptimizedStorageController`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


docs/control/technology_level_control/figures/plm_optimized_dispatch.png
johnjasa, line 1: Could you please add definitions or explanations for the shaded gold regions in the figure or legends?

environment.yml
```diff
@@ -3,6 +3,7 @@ channels: [conda-forge, defaults]
 dependencies:
   - python=3.11
   - glpk
+  - highspy
```
johnjasa, line 6: Thanks for adding the `highspy` req here! Could you please add it to the conda install command within the `install.md` doc page too please?

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -211,14 +211,14 @@ def test_optimizer_dispatch_only_on_eligible_timesteps(subtests, base_config):
         storage_capacity=base_config.max_capacity,
     )
 
-    PeakLoadManagementOptimizedStorageController.glpk_solve_call(model)
+    PeakLoadManagementOptimizedStorageController.pyomosolver_solve_call(model)
```
johnjasa, line 215: I like this generalization away from `glpk` to the `pyomosolver`! Thanks

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -748,6 +825,20 @@ def _get_storage_dispatch_commands(self) -> list:
...
+
+    @staticmethod
+    def _GnT_pricingfunction(lmp: float) -> float:
+        """Compute the cost a G$T charges to a CoOp based on the current grid price.
+
+        Args:
+            lmp (float): Current grid price (locational marginal price), e.g., in $/MWh.
+
+        Returns:
+            float: Cost charged by the G$T in the same units as `lmp`.
```
johnjasa: Should this be `G&T` instead of `G$T`? Maybe I'm just not familiar with the initialism

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -748,6 +825,20 @@ def _get_storage_dispatch_commands(self) -> list:
...
+        """
+        return 1.05 * lmp + 20
```
johnjasa: Should these multiplicative and offset coefficients be input or settable in some way? Or is it okay that they're hardcoded here? Will they always be these values?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -717,26 +794,26 @@ def solve_dispatch_model(self, start_time: int = 0, n_days: int = 0):
         )
 
     @staticmethod
-    def glpk_solve_call(
+    def pyomosolver_solve_call(
```
johnjasa, line 851: My own curiosity: why move away from GLPK here to use highs? Is it more robust, faster, better for this problem, or something else? Should we avoid using GLPK in other contexts within H2I?

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,16 +111,17 @@ tech_to_dispatch_connections: [
 ]
 ```
 
-# Optimized Demand Response Controller
+## Optimized Demand Response Controller
```
johnjasa, line 115: Suggestion: the actual class file has `PeakLoadManagement` in the name; I'd suggest keeping the same naming convention here in the docs, and maybe even referencing or linking to the class itself so users can dig deeper.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -207,18 +209,6 @@ def setup(self):
...
-
-        Args:
-            inputs (dict): OpenMDAO inputs dict. Recognized keys are
-                ``'max_charge_rate'`` and ``'storage_capacity'``.
-        """
-        if "max_charge_rate" in inputs:
-            object.__setattr__(self.config, "max_charge_rate", float(inputs["max_charge_rate"][0]))
-        if "storage_capacity" in inputs:
-            object.__setattr__(self.config, "max_capacity", float(inputs["storage_capacity"][0]))
-
```
johnjasa, line 233: Is this just no longer necessary? Can we still optimize the `max_charge_rate` and `storage_capacity` values at the OM level and they'll get updated? It looks like yes as they're used on line 250, but I'm asking to double-check!

johnjasa (dismissed):
This is great, thanks for the follow-on PR, @abhineet-gupta! I love how you updated the code, examples, tests, and docs. Really slick stuff with some good solver generalizations as well.

I've left some questions and suggestions, nothing huge but I do think some of the clarifications will be useful for users approaching this code.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,16 +111,17 @@ tech_to_dispatch_connections: [
...
 
 The controller works at any simulation timestep resolution (`dt`). All time-based parameters ( `event_duration`, `min_peak_separation`) are specified in physical time units (hours, minutes, etc.) and are internally converted to timesteps using `dt`.
 
-## Definitions
+### Definitions
 
 **Given:**
-- $\lambda_t$ := `supervisory_signal`: price, demand, or price $\times$ demand time series at timestep $t$
+- $\lambda_t$ := `lmp_signal`: electricity price time series at timestep $t$
```
vijay092: I'd call this a supervisory signal, since it could be either LMP or demand- essentially any external signal that the battery needs to respond to.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -140,33 +141,37 @@ The controller works at any simulation timestep resolution (`dt`). All time-base
...
 
-### Objective
+#### Objective
 
-Maximize total incentive revenue over the window:
+Minimize Co-op's cost and maximize total incentive revenue over the optimization window:
 
 $$
-\max_{u_t, v_t,p_{d,t}, p_{c,t}} \quad \gamma \cdot \Delta t \sum_{t \in \mathcal{T}} p_{d,t}
+\min_{u_{gt,t},u_{coop,t},v_t,p^d_{gt,t},p^d_{coop,t},pc_t,\text{SOC}_t,p_{gt2coop,t}} \quad
```
vijay092, line 175: Could you confirm if the units are uniform in the first and second term? Incentive is $ / kwh and LMP is \$/MWh?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
vijay092: I agree.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -332,13 +332,14 @@ def test_optimizer_dispatch_respects_charge_discharge_exclusivity(subtests, base
...
 
     for t in range(24):
         charge = pyomo.value(model.charge[t])  # type: ignore[index]
-        discharge = pyomo.value(model.discharge[t])  # type: ignore[index]
-        with subtests.test(f"No simultaneous charge and discharge at t={t}"):
-            assert not (charge > 0.5 and discharge > 0.5)
+        discharge_gt = pyomo.value(model.discharge_gt[t])  # type: ignore[index]
+        discharge_coop = pyomo.value(model.discharge_coop[t])  # type: ignore[index]
+        with subtests.test(f"No simultaneous charge, discharge_gt or discharge_coop at t={t}"):
+            assert not (charge > 0.5 and discharge_gt > 0.5 and discharge_coop > 0.5)
```
vijay092: This allows two conditions to be > 0.5 which should not happen, right?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -573,37 +605,57 @@ def _build_dr_model(
...
         m.soc = pyomo.Var(
             m.T,
             domain=pyomo.NonNegativeReals,
             bounds=(soc_min, soc_max),
             doc="State of charge SoC_t",
         )
 
-        # Incentive revenue is earned for every kWh discharged.
+        # Power transmitted to CoOp
+        m.p_tocoop = pyomo.Var(
```
vijay092, line 682: `p_tocoop` is unbounded above  which might cause issues later on. Good to add an upper bound.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -18,52 +20,57 @@
...
-if event_dur_cfg is not None:
-    half_td = pd.Timedelta(value=event_dur_cfg["val"], unit=event_dur_cfg["units"]) / 2
-
-threshold_pct = np.percentile(lmp, percentile)
-discharge_mask = battery_power > 0
-
-
+# Intermediate MILP decision variables
+u_dicharge_gt = controller.discharge_gt_bin_history
+u_dicharge_coop = controller.discharge_coop_bin_history
```
vijay092: typo

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -18,52 +20,57 @@
...
+v_charge = controller.charge_bin_history
+p_dicharge_gt = controller.p_discharge_gt_history
+p_dicharge_coop = controller.p_discharge_coop_history
```
vijay092: typo

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -30,9 +30,9 @@ class PeakLoadManagementOptimizedControllerConfig(PyomoStorageControllerBaseConf
 
     Attributes:
         max_charge_rate (float): Maximum charge and discharge rate (kW).
-        supervisory_signal (list[float]): Price, demand, or price*demand
-            forecast time series. The rolling horizon solver uses one window of
-            length ``n_control_window_hours`` per solve.
+        lmp_signal (list[float]): Locational Marginal Pricing (LMP)
+            forecast time series.
+        demand_signal (list[float]): Consumer demand forcast time series
```
vijay092: typo

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -573,37 +605,57 @@ def _build_dr_model(
...
+        # Power transmitted to CoOp
+        m.p_tocoop = pyomo.Var(
+            m.T,
+            domain=pyomo.NonNegativeReals,
+            bounds=(0, None),
+            doc="Power supplied by G&T to Co-Op (kW)",
+        )
+
+        # Objective is maximizing incentive revenue is earned for every kWh discharged and
+        # minimzing the cost of energy for the CoOp.
```
vijay092: typo

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -304,13 +306,34 @@ def pyomo_dispatch_solver(
                 # Solve the optimzation problem
```
vijay092: typo

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -515,36 +525,100 @@ def test_plm_optimized_controller_om_problem_soc_bounds(subtests, om_plant_confi
...
-                    discharge[t] <= 1e-4
-                ), f"Discharge {discharge[t]:.4f} kW outside peak window at t={t}"
-
     with subtests.test("Discharge never negative"):
         assert np.all(discharge >= -1e-4)
 
     with subtests.test("Discharge never above max_charge_rate"):
         assert np.all(discharge <= 1.0 + 1e-4)
 
-    with subtests.test("SOC at t=0 equal to init_soc_fraction"):
```
vijay092, line 610: Why is this removed?

vijay092 (commented):
Thank you for working on this! I had some minor comments.

docs/control/technology_level_control/figures/plm_optimized_dispatch.png
abhineet-gupta: Addressed in 63786a76. Added explanation below the figure.

environment.yml
abhineet-gupta: Addressed in 63786a76.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Highs is supposed to be faster. We have more optimization variables and constraints in this updated framework but get reasonable speed with highs.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Addressed in 63786a76. Added the class name in the description following the convention used for the other controllers described in `docs/control/technology_level_control/pyomo_controllers.md` page.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: This controller needs both LMP and demand signals as inputs at the same time, hence the distinction.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: The optimization setup expects $/kWh for both. I will double check the input data in example to ensure that they follow the convention.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
abhineet-gupta: That's correct. Fixed in 9cb57389.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: That's a good point. There would be a practical upper bound (based on interconnections etc). However, there is no theoretical upper bound  (we can potentially model as large a G&T and CoOp under this framework as we like). Not sure how to pick a reasonable upper bound here.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: These lines were redundant as [these](https://github.com/abhineet-gupta/AG_H2Integrate/blob/021e803cf525cdda3df9d0df8b8cc8e1d8f554d5/h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py#L154) lines add the inputs to config properly.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
abhineet-gupta: SOC at t = 0 is calculated after the first iteration is complete. So this test is only valid  if battery doesn't discharge at the first step. Previously, that wasn't possible because the battery could only discharge during peak window but now it can also discharge at the first step.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Fixed in 021e803c

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Fixed in 021e803c

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
abhineet-gupta: Fixed in 021e803c

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
abhineet-gupta: Fixed in 021e803c

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Good point. Added the coefficients as inputs now. Addressed in commit 938b3956.

abhineet-gupta:
@johnjasa,
Thanks for the review, I have addressed your comments with and tagged them with specific commit hashes.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
jaredthomas68: I think the upper bound could be exposed as a user input to the control in the tech_config. I think it would appropriate to have a default value of None but allow the user to set it as well. It may also work to pull it from the performance model rating information, perhaps with a buffer.

docs/control/technology_level_control/pyomo_controllers.md
jaredthomas68: Is there just one demand signal? If the LMP could also be an upstream demand signal (as in the heuristic control work), then it may make sense to adjust the name as @vijay092 suggested. I see a supervisory signal file, but I only see one demand and lmp in the docs formulation.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -176,21 +181,25 @@ The factor $\Delta t$ converts power (kW) to energy (kWh), so the objective is c
...
 $$
-u_t = 0 \qquad \forall\, t \notin \mathcal{D}
+u_{gt,t} = 0 \qquad \forall\, t \notin \mathcal{D}
 $$
 
 - Maximum $N_{\max}$ discharge events per month. Because `event_duration` fixes each event to exactly $\tau$ timesteps, the event cap translates directly into a timestep cap:
 
 $$
-\sum_{t \in \mathcal{M}_m \cap \mathcal{T}} u_t \leq B_m \cdot \tau \qquad \forall\, m
+\sum_{t \in \mathcal{M}_m \cap \mathcal{T}} u_{gt,t} \leq B_m \cdot \tau \qquad \forall\, m
```
jaredthomas68, line 195: It appears that only the number of hours (or timesteps?) are constrained per month, rather than the number of events, which I think is intended to be a full battery cycle per event. From the plot, it looks like multiple discharge events happen per day, while we need to be able limit the number of events per day as well as the number of events per month. I may be misunderstanding, but I think it would be worth revisiting this formulation against the requirements. My understanding of what is needed:
- constrain the number of events in a month
- constrain the number of events in a day
- specify whether an event should be a full battery cycle or use some other definition of what an "event" is.

To be clear, I like what you have here and I think we should keep it. I also think it is not sufficient for the current project needs.

library/demand_profiles/demand_profile_consumers.yaml
jaredthomas68, line 1: It may be worth moving these demand and lmp profiles to the `library/demand_profiles` as I did in #773, but probably less important here since they are only used in one example so far.

jaredthomas68, line 1: Can you add descriptions at the top of each lmp/demand/other signal with information about the data, where it came from, units, etc?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -269,6 +261,18 @@ def pyomo_dispatch_solver(
             storage_out = np.zeros(self.n_timesteps)
             soc_out = np.zeros(self.n_timesteps)
 
+            # Per-timestep history of the MILP decision variables, accumulated
```
jaredthomas68, line 279: I think this is a fine way of storing if you just need them for use from the object immediately after a run. If you want them saved in the openmdao recorder database for later analysis from a saved file then we should consider making them outputs of the component.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -301,16 +305,37 @@ def pyomo_dispatch_solver(
...
                 for t in range(window_len):
-                    discharging = pyomo.value(self.dr_model.discharge[t]) > 0.5
-                    prev_discharging = t > 0 and pyomo.value(self.dr_model.discharge[t - 1]) > 0.5
+                    abs_t = window_start + t
+
+                    d1_val = pyomo.value(self.dr_model.discharge_gt[t])
+                    d2_val = pyomo.value(self.dr_model.discharge_coop[t])
+                    c_val = pyomo.value(self.dr_model.charge[t])
+
+                    self.discharge_gt_bin_history[abs_t] = d1_val
```
jaredthomas68, line 351: This is nit-picky and optional, but I think this controller is applicable more broadly than just for a coop and G&T, so I think it would be worth changing the names to use patterns like `_local_` and `_supervisor_` or `_upstream_`that are more general.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -573,37 +607,57 @@ def _build_dr_model(
...
         m.charge = pyomo.Var(
             m.T, domain=pyomo.Binary, doc="1 if a charge event is active at timestep t"
         )
         # Actual kW dispatched each timestep.
-        m.p_discharge = pyomo.Var(
+        m.p_discharge_gt = pyomo.Var(
+            m.T,
+            domain=pyomo.NonNegativeReals,
+            bounds=(0, P_max),
+            doc="Discharge power (kW) to G&T at timestep t",
```
jaredthomas68: I think this is never actual discharge "to" the G&T/supervisor/upstream. I this is just Discharge to the local/coop as commanded by the G&T/supervisor/upstream.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -573,37 +607,57 @@ def _build_dr_model(
...
+        )
+        m.p_discharge_coop = pyomo.Var(
             m.T,
             domain=pyomo.NonNegativeReals,
             bounds=(0, P_max),
-            doc="Discharge power (kW) at timestep t",
+            doc="Discharge power to Co-Op (kW) at timestep t",
```
jaredthomas68: as determined by the local/coop

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -656,15 +714,16 @@ def soc_evolution_rule(mdl, t):
...
                 + eta_c * mdl.p_charge[t] * dt_hours / E_max
-                - mdl.p_discharge[t] * dt_hours / (eta_d * E_max)
+                - mdl.p_discharge_gt[t] * dt_hours / (eta_d * E_max)
+                - mdl.p_discharge_coop[t] * dt_hours / (eta_d * E_max)
             )
 
         m.soc_evolution = pyomo.Constraint(m.T, rule=soc_evolution_rule)
 
-        # Can't charge and discharge at the same time.
+        # Can't charge to both G&T and CoOp or discharge at the same time.
```
jaredthomas68: The discharge only ever impacts the coop directly, the impact on the G&T is secondary.

``` suggestion
        # Can't simultaneously discharge according to both G&T and COOP, or charge and discharge at the same time.
```

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -675,6 +734,26 @@ def soc_evolution_rule(mdl, t):
             ),
         )
 
+        # Can't discharge to CoOp in the dispatch window.
```
jaredthomas68:
``` suggestion
        # Can't follow discharge commands from the COOP in the dispatch window.
```

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -717,26 +796,26 @@ def solve_dispatch_model(self, start_time: int = 0, n_days: int = 0):
         )
 
     @staticmethod
-    def glpk_solve_call(
+    def pyomosolver_solve_call(
```
jaredthomas68, line 851: I like the move away from having a method call in the control to call a specific solver by name, but I think we should use that general method (`pyomosolver_solve_call`) to run the solver of choice (GLPK, HIGHs, etc). could we either add some logic or methods for each solver that can be called by `pyomosolver_solve_call` so the user can choose the solver to use?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -748,6 +827,22 @@ def _get_storage_dispatch_commands(self) -> list:
...
+        """Compute the cost a G&T charges to a CoOp based on the current grid price.
+
+        Args:
+            lmp (float): Current grid price (locational marginal price), e.g., in $/kWh.
+
+        Returns:
+            float: Cost charged by the G$T in the same units as `lmp`.
+        """
+        return (
+            self.config.GnT_pricingfunction_coeffs[0] * lmp
```
jaredthomas68: I think the COOP price needs to be independent of LMP in some cases. Maybe that just means setting the first coeff to 0? I'm also wondering how we may be able to build in other electricity pricing approaches than what is represented here. We should probably discuss offline about other project needs and make sure we make the pricing and objective functions here work for what is needed.

jaredthomas68 (dismissed):
Nice work, this is looking really good! Most of my comments are related to generality and needs of other projects hoping to use this work. If you have any questions or concerns I'm happy to clarify and discuss potential solutions.

abhineet-gupta:
> Nice work, this is looking really good! Most of my comments are related to generality and needs of other projects hoping to use this work. If you have any questions or concerns I'm happy to clarify and discuss potential solutions.

Thank you for the detailed comments @jaredthomas68. Addressing them soon.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 0681dcfd.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 0681dcfd.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 0681dcfd.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 0681dcfd.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 114b921b.
Added `pyomo_solver` and `pyomo_solver_options` as user inputs.

Right now, the configuration is being read and used only by `plm_optimized_storage_controller.py` but it might be a good idea to move this logic to `pyomo_storage_controller_baseclass.py` next (in a separate PR).

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: There are two signals, one for lmp and one for demand. The objective function of the optimization assumes the LMP signal has units of $/kWh and demand signal has units of kWh. I don't think this framework allows for the lmp signal to represent a demand.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: I have made all the units consistent. Everything is in $/kWh, kWh and kW.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in 0642b27b.

library/demand_profiles/demand_profile_consumers.yaml
abhineet-gupta: Addressed in 59513f2a

abhineet-gupta: Addressed in 59513f2a

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: My understanding is that that would require modifying the base class `pyomo_storage_controller_baseclass`. If that is correct then would that not break the functionality of other controller that depend on this baseclasee but do not have signals like `p_discharge_gt_history`.

docs/control/technology_level_control/figures/plm_optimized_dispatch.png
vijay092, line 1: So I was wondering why the battery is discharging in response to the troughs in the plot. It is cycling twice in the off peak region.

vijay092: Also, the LMP is $ / MWh here which is fine but we just need to make sure the input is $/kWh

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -749,6 +852,22 @@ def _get_storage_dispatch_commands(self) -> list:
...
+    def _GnT_pricingfunction(self, lmp: float) -> float:
+        """Compute the cost a G&T charges to a CoOp based on the current grid price.
+
+        Args:
+            lmp (float): Current grid price (locational marginal price), e.g., in $/kWh.
+
+        Returns:
+            float: Cost charged by the G$T in the same units as `lmp`.
+        """
+        return (
```
vijay092: Make sure the length is validated somewhere

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -18,52 +20,57 @@
 EXAMPLE_DIR = Path(__file__).parent
 
 
+# Run H2Ingegrate
```
vijay092: H2Integrate

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
```diff
@@ -18,52 +20,57 @@
...
-        "supervisory_signal"
-    ]
-)[:N]
-
+# Read inputs from config
+control_params = model.technology_config["technologies"]["battery"]["model_inputs"][
+    "control_parameters"
+]
 n_timesteps = int(model.plant_config["plant"]["simulation"]["n_timesteps"])
+lmp = np.array(control_params["lmp_signal"])[:n_timesteps]
```
vijay092, line 33: Is this in $ / MWh

docs/control/technology_level_control/figures/plm_optimized_dispatch.png
vijay092: Also why is it never going to the SOC max limit in black dotted lines?

vijay092 (approved):
Thanks for addressing all our comments! Very nice work, and I'm excited to use this for the demo.

docs/control/technology_level_control/figures/plm_optimized_dispatch.png
abhineet-gupta: I have updated the plot with the latest plm implementation and that has addressed these issues. The plot should look good now. Let me know what you think now.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Based on Jared's comment, I have made this a variable length array that can essentially model any order polynomial function.

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
abhineet-gupta: Fixed

examples/34_plm_optimized_dispatch/run_plm_optimized_dispatch.py
abhineet-gupta: I have switched it to $/kWh and labeled both the yaml file and plot accordingly.

johnjasa:
I probably won't get a chance to re-review this as I'm out early next week; feel free to merge it in once we get another approval on it. I've dismissed Jared and my's reviews so we don't hold up the process.

abhineet-gupta:
> I probably won't get a chance to re-review this as I'm out early next week; feel free to merge it in once we get another approval on it. I've dismissed Jared and my's reviews so we don't hold up the process.

Thank you, @johnjasa.

@vijay092, @genevievestarke, @elenya-grant 
I need some help with a bug in the repo's configuration that might have been inadvertently introduced in PR #805.
If any one of you can take a quick look at the following and approve my suggested bugfix, that'd be great.

The readthedocs build fails for this PR and the logs say:
```
failed to import h2integrate.converters.combustion_machines.
  Possible hints:
  * AttributeError: module 'h2integrate.converters' has no attribute 'combustion_machines'
  * ImportError:
  * ModuleNotFoundError: No module named 'pyfluids'
```
which was introduced in PR #805.

It seems like documentation build needs to import the `pyfluids` module as it is needed [here](https://github.com/NatLabRockies/H2Integrate/blob/af41f11b6273c1bface2ab9e333e3b04ba1f6aa3/h2integrate/converters/combustion_machines/NGCT_thermo_model.py#L2) but it is not included in the `develop` dependency group. 

I you have any thoughts on the following, I'd appreciate it.
- I am not sure why the readthedocs build was not triggered for PR #805.
- I propose to including "h2integrate[thermo]" under the develop dependency group in `pyproject.toml`. Would that be OK?

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,16 +111,17 @@ tech_to_dispatch_connections: [
...
+The optimized demand response controller is specified by setting the storage control to `PeakLoadManagementOptimizedStorageController`. This controller optimizes the dispatch of a Battery Energy Storage System (BESS). It is demonstrated for a scenario in which a Generation and Transmission cooperative (G&T) is connected to a Distribution Co-operative (Co-op). The battery is owned and operated by the Co-op, primarily to reduce its electricity cost; in addition, the G&T can request battery dispatch a limited number of times during peak LMP periods, in exchange for incentive payments. Using a pre-defined Locational Marginal Price (LMP) profile and consumer power demand profile as inputs, the controller maximizes the incentive payments earned from G&T-requested dispatches while minimizing the Co-op's electricity cost, subject to constraints on the maximum number of dispatch events per month and the battery's state of charge. The result demonstrates peak load management and demand response from a single coordinated controller.
```
genevievestarke: This is nitpicky, but we are spelling cooperative two different ways in this paragraph, back to back in the G&T and Co-op definition sentence. I think we should pick one and stick with that spelling!

docs/control/technology_level_control/pyomo_controllers.md
genevievestarke, line 115: Does this also enable you to do demand response without a G&T signal? I think it would be good to note that that is also possible here.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,16 +111,17 @@ tech_to_dispatch_connections: [
...
 
 The controller works at any simulation timestep resolution (`dt`). All time-based parameters ( `event_duration`, `min_peak_separation`) are specified in physical time units (hours, minutes, etc.) and are internally converted to timesteps using `dt`.
 
-## Definitions
+### Definitions
```
genevievestarke, line 121: I think the nomenclature here makes sense, but I'm not sure we need `null` to be emphasized. The rest of the emphasis seems to be for variables/inputs?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -71,11 +72,17 @@ class PeakLoadManagementOptimizedControllerConfig(PyomoStorageControllerBaseConf
...
+        pyomo_solver_options (dict): A dictionary containing options for pyomo solver.
+        gt2coop_limit (float): Upper limit of power transferred by G&T to CoOp (kW)
     """
 
     max_charge_rate: float = field()
-    supervisory_signal: list = field()
+    lmp_signal: list = field()
+    demand_signal: list = field()
     peak_window: dict = field()
+    GnT_pricingfunction_coeffs: list = field()
```
genevievestarke, line 89: To not have a g&t signal, do you just set these to zero?

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
genevievestarke: I agree that at some point we should remove the explicit g&t and coop references and make it more general.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -29,11 +29,12 @@ class PeakLoadManagementOptimizedControllerConfig(PyomoStorageControllerBaseConf
...
         max_charge_rate (float): Maximum charge and discharge rate (kW).
-        supervisory_signal (list[float]): Price, demand, or price*demand
-            forecast time series. The rolling horizon solver uses one window of
-            length ``n_control_window_hours`` per solve.
+        lmp_signal (list[float]): Locational Marginal Price (LMP)
+            forecast time series ($/kWh).
+        demand_signal (list[float]): Consumer demand forecast time series (kW).
         peak_window (dict): Hours eligible for dispatch. Keys ``'start'``
             and ``'end'`` must be strings in ``HH:MM:SS`` format.
+        GnT_pricingfunction_coeffs (list[float]): Coefficients for G&T pricing function
```
genevievestarke: I think this could use a little more explanation how the user is meant to input these, and how they are used in the code. From the function, it looks like they are building a polynomial, where the coefficients go from left to right (highest power to lowest power)?

examples/34_plm_optimized_dispatch/tech_config.yaml
```diff
@@ -23,11 +23,13 @@ technologies:
         demand_profile: 0.0
       control_parameters:
         system_commodity_interface_limit: 1.0e9
-        supervisory_signal: !include demand_profiles/supervisory_signal.yaml  # found in library/demand_profiles
+        lmp_signal: !include demand_profiles/lmp_profile.yaml
+        demand_signal: !include demand_profiles/demand_profile_consumers.yaml
+        GnT_pricingfunction_coeffs: [1.05, 0.02]
```
genevievestarke, line 28: How sensitive is the optimization to the input variables `GnT_pricingfunction_coeffs` and 'performance_incentive_per_event`? Is there a range that they should stay in? Is there a relationship between these values?

genevievestarke (commented):
It looks really good @abhineet-gupta! I had some comments about the inputs and documentation, mostly! Thank you for implementing this!

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Good point. I'll fix it.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: No, this setup requires the lmp and the cost G&T charges the CoOp.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Agreed. I'll fix it.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: All these are required in this framework.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Agreed, I'll add an explanation.

examples/34_plm_optimized_dispatch/tech_config.yaml
abhineet-gupta: The optimization is not sensitive to small changes in incentive but is sensitive to `GnT_pricingfunction_coeffs`. There is no explicit relationship between these parameters. 
Only constraint is that coefficient should be such that they represent a `strictly increasing function` for the finances to make sense but I haven't implemented that constraint.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Addressed in f8f9e8ad.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Addressed in f8f9e8ad.

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Addressed in f8f9e8ad

abhineet-gupta:
Thanks @genevievestarke. I have addressed your comments. Let me know what you think.
I will create a separate PR to refactor the G&T and Coop.

abhineet-gupta:
@johnjasa Thanks for fixing the issue in the develop branch. That fixed the readthedocs build.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,86 +111,98 @@ tech_to_dispatch_connections: [
...
 - $P_{\max}$ := `max_charge_rate` (kW): maximum charge and discharge rate
 - $E_{\max} :=$ `max_capacity` $\times$ (`max_soc_fraction` $-$ `min_soc_fraction`): usable energy capacity (kWh)
 - $\eta_c$ := `charge_efficiency`, $\quad \eta_d$ := `discharge_efficiency`
 - $\text{SoC}_{\max}$ := `max_soc_fraction`, $\quad \text{SoC}_{\min}$ := `min_soc_fraction`
+- $\text{gt2coop_limit}$ := upper limit on power transmitted from G&T to Coop.
 - `n_control_window_hours` := rolling horizon length in hours; converted to $T =$ `n_control_window_hours` / $\Delta t$ timesteps
 - $\mathcal{T} := \{0, 1, \ldots, T-1\}$: timesteps in the current rolling window
 - $\mathcal{M}_m$ := set of timesteps in month $m$, for $m = 1, \ldots, 12$
 - $N_{\max}$ := `n_max_events`: maximum number of discharge events per calendar month
 - $\tau$ := `steps_per_event` : number of timesteps per event (1 when `event_duration` is `null`)
```
genevievestarke: There's a few more `null` definitions running around the docs. It's not super important, but I did want to call them out to be consistent.

docs/control/technology_level_control/pyomo_controllers.md
```diff
@@ -111,86 +111,98 @@ tech_to_dispatch_connections: [
...
 - $B_m$ := remaining event budget for month $m$ = $N_{\max}$ minus events already dispatched in prior windows
 
-## Dispatch Window Construction
+### Dispatch Window Construction
 
 Before the MILP is solved, the dispatch window $\mathcal{D}$ is built in two steps:
 
 **Step 1 : Peak selection:** Within $\mathcal{W}$, timesteps at or above the `signal_threshold_percentile` of $\lambda_t$ are marked eligible: $\mathcal{E} = \{t \in \mathcal{W} : \lambda_t \geq \lambda_*\}$. If `min_peak_separation` is set, only the first peak is chosen.
 
 **Step 2 : Event window expansion:** If `event_duration` is specified, each peak in $\mathcal{E}$ is expanded by $\pm$ `event_duration`/2 timesteps to form $\mathcal{D}$. If `event_duration` is `null`, $\mathcal{D} = \mathcal{E}$.
```
genevievestarke: Final `null`

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
```diff
@@ -29,11 +29,16 @@ class PeakLoadManagementOptimizedControllerConfig(PyomoStorageControllerBaseConf
...
         max_charge_rate (float): Maximum charge and discharge rate (kW).
-        supervisory_signal (list[float]): Price, demand, or price*demand
-            forecast time series. The rolling horizon solver uses one window of
-            length ``n_control_window_hours`` per solve.
+        lmp_signal (list[float]): Locational Marginal Price (LMP)
+            forecast time series ($/kWh).
+        demand_signal (list[float]): Consumer demand forecast time series (kW).
         peak_window (dict): Hours eligible for dispatch. Keys ``'start'``
             and ``'end'`` must be strings in ``HH:MM:SS`` format.
+        GnT_pricingfunction_coeffs (list[float]): Coefficients for G&T pricing function.
```
genevievestarke, line 37: Than you for adding a description of how this is used, I think it's much more clear! Can we add here a note here that they have to describe a strictly increasing function?

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Agreed. Fixed in e9295e10.

docs/control/technology_level_control/pyomo_controllers.md
abhineet-gupta: Fixed in e9295e10

h2integrate/control/control_strategies/storage/plm_optimized_storage_controller.py
abhineet-gupta: Added in e9295e10

genevievestarke (approved):
Nice, looks good @abhineet-gupta !

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -589,36 +607,100 @@ def test_plm_optimized_controller_om_problem_soc_bounds(subtests, om_plant_confi
...
+            0.0,
+        ]
+    )
+
+    for t in range(n):
+        with subtests.test(f"p_discharge_coop_history[{t}] matches expected"):
+            assert abs(p_discharge_coop_history[t] - expected_history[t]) < 1e-4, (
+                f"p_discharge_coop_history mismatch at t={t}: "
+                f"got {p_discharge_coop_history[t]:.4f}, expected {expected_history[t]:.4f}"
+            )
```
vijay092:
```suggestion
    active          = p_discharge_coop_history > 0.5
    expected_active = expected_history > 0.5
    
    with subtests.test("total coop discharge energy matches"):
        assert abs(p_discharge_coop_history.sum() - expected_history.sum()) < 1e-4
    with subtests.test("number of coop discharge steps matches"):
        assert active.sum() == expected_active.sum()
```

vijay092 (commented):
Please try the suggested changes to fix failed tests.

h2integrate/control/control_strategies/storage/test/test_plm_optimized_storage_controller.py
```diff
@@ -589,36 +607,101 @@ def test_plm_optimized_controller_om_problem_soc_bounds(subtests, om_plant_confi
...
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+            0.0,
+        ]
+    )
+
+    active          = p_discharge_coop_history > 0.5
```
vijay092:
```suggestion
    active = p_discharge_coop_history > 0.5
```

Issue #809, opened 2026-07-21
More checking on `technology_interconnections`


elenya-grant:
# More checking on `technology_interconnections`

As system level control gets further developed - it will likely become even _more_ dependent on `technology_interconnections` for understanding the system architecture. With more dependence on `technology_interconnections`, we should integrate more checking on the connections and throw errors for "invalid" connections or connections that should be done differently. I think a good starting point would be the following:

1. Check for any connections that are formatted as `[source_tech, dest_tech, [commodity_out, commodity_in]]` and throw an error or warning to note that this should be formatted as `[source_tech, dest_tech, commodity, transport_tech]`. The length 3 connection does not provide information about the commodity being passed (when used to make the technology graph). So - this type of connection should be discouraged when length 4 connection would work. This check could be done by first checking if the end of `source_param` and `dest_param` are `_out` and `_in` (respectively). If this is the case, then check if the starts of that string are the same (`if source_param.split("_in",1) == dest_param.split("_out", 1)` then raise error that encourages user to use a length 4 connection and utilize the generic transport component if needed.)
2. Check usage of combiners with storage components. With storage components, its common to have the following type of connection:
```yaml
- [generation_tech, storage_tech, commodity, transport_tech]
- [generation_tech, combiner, commodity, transport_tech]
- [storage_tech, combiner, commodity, transport_tech]
- [combiner, downstream_tech, commodity, transport_tech]
```
In these cases, we should check that:
- technology connected to storage_tech (`generation_tech`) has only two out streams (unless its a splitter). These out streams would be `storage_tech` and `combiner`
- `storage_tech` has only one out stream, which is the same combiner as `generation_tech`

For all other technologies connected with length 4 connections, we should ensure that they each have a maximum of 1 in stream and 1 out stream (except splitters). 

This can be done somewhat easily using the `technology_graph` attribute. Length 4 connections mean that the edge in the technology graph has the "commodity" data being passed. A somewhat simplified approach is below (does not have any pre-filtering for length 4 connections):

```python
in_degs = dict(self.technology_graph.in_degree)
out_degs = dict(self.technology_graph.out_degree)

# check storage technologies
storage_techs = [k for k,v in self.tech_control_classifiers.items() if v=="storage"]
for storage_tech in storage_techs:
   if (n_in := in_degs.get(storage_tech,0)) != 1:
      raise ValueError(f"storage tech {storage_tech} has {n_in} input connections but should only have 1")
[...]

PR #816, opened 2026-07-29
WIP: Helper Methods in SLC BaseClass

elenya-grant:
# Helper Methods in SLC BaseClass

Added some hopefully useful helper methods in the SLC baseclass and intended to resolve Issue #741. I also view this as a pre-requisite to tackling Issue #747. These methods are intended to help future control development be more easily expanded to varying system architectures and make them more friendly to multi-commodity systems (systems that contain converters such as an electrolyzer or Haber Bosch system). To me, this meant being able to understand complex systems in a more understandable way and to be able to easily translate the demanded commodity into the demand of each upstream technology. For example, how to translate the ammonia demand into hydrogen demand, and also translate that hydrogen demand into electricity demand and water demand. 

The methods in the SLC baseclass **do work** for more complex system architectures than we can currently represent with the models in H2I. I think that the methods would be able to handle architectures that include splitters (and if not, I think the changes to them would be more simple). I think that integrating splitters may just change how some of the methods called by controllers that inherit the baseclass provide the necessary inputs. Aka - I think that `_post_setup_multi_commodity()` should work for splitters (or require minor changes). 

This PR also includes some updates to the `DemandFollowingControl` to show-case how to use some of the methods in the SLC baseclass. I am still working on adding all the necessary doc-strings and cleaning up the code.

**Please do not yet push any changes to this branch until the TODOs are complete** 

**Note on splitters and future development** currently (as of 08/11/2026): The `converter_upstreams` output from `_find_converter_techs()` method may not work properly with splitters, which is actually due to some logic in `get_successors_for_tech_with_input_cmod()` which could be easily fixed (I've added comments in on how to update it for splitters). Separately, a ValueError will be raised in `_post_setup_multi_commodity()` if a splitter is used because technologies will exist in multiple groups. This error could be replaced with some special logic to handle groups that contain a splitter. Additionally, during the creation of `simple_graph`, it would be good to keep track of the "priority tech" that the splitters are connected to. Anyway - this is not for this PR but just noting that I've dug into what happens if a splitter is included.

### A note to reviewers:

Hi reviewer! First, I tried to put some "worked examples" in the doc strings of the methods I added, and these can be seen in a nice way if you look at the read the docs build for the SLC base class. All the methods I added are at the end of the file. I also showed examples of the main attributes instantiated in the `_post_setup_multi_commodity()` method in the docs page for the SLC baseclass under the section titled "Heterogeneous Commodity Hybrid System". Give those a look! 

[...]

PR #832, opened 2026-08-11
More Checking on `technology_interconnections`

johnjasa:
# More Checking on `technology_interconnections`

This PR adds structured topology validation for `technology_interconnections` in `H2IntegrateModel`, raising descriptive errors for common configuration mistakes that previously went undetected until OpenMDAO setup or silently produced wrong results.

Summary of changes:
1. Add `_validate_technology_interconnections` method with three topology checks.
2. Extend `tech_control_classifiers` population to always run (was SLC-only).
3. Call the new validation in `setup()` before the existing commodity check.
4. Add integration tests covering all new error cases.

- [x] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [x] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Implement all three topology checks
- [x] Extend `tech_control_classifiers` to non-SLC models
- [x] Add tests for each new error case
- [x] Update CHANGELOG

**Implementation feedback**:
Confirm that the three topology checks (length-3 commodity pair detection, storage topology,
and general max-in/out) match the expected behavior described in the issue.

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues
Solves #809

### Section 5.1: New Files
N/A

  - Add `_validate_technology_interconnections` method.
  - Extend `tech_control_classifiers` population to non-SLC models.
  - Call new validation in `setup()`.
  - Add `test_validate_technology_interconnections` integration test.
  - Update `test_check_tech_interconnections` second subtest to use a non-storage tech.

The three checks implemented are:

1. **Discouraged length-3 `[commodity_out, commodity_in]` connections** - When a length-3 connection passes a list pair like `["hydrogen_out", "hydrogen_in"]` where the prefixes match, a `ValueError` is raised prompting the user to use a length-4 connection with an explicit commodity and transport component.

2. **Storage technology topology** - For each storage technology in `tech_control_classifiers`: exactly 1 length-4 input is required; at most 1 length-4 output is allowed; and the direct upstream technology may have at most 2 output streams (one to storage, one to a combiner).

3. **General max-in/out for non-special technologies** - All other technologies connected via length-4 connections (excluding splitters, combiners, storage techs, and the direct upstream of storage techs) must have at most 1 input stream and 1 output stream. The storage upstream exemption applies only to the output-stream count; input streams are still checked.

The fix to always populate `tech_control_classifiers` (removing the `self.slc` guard) ensures checks 2 and 3 work correctly for non-SLC systems where storage is still present. All other usages of `tech_control_classifiers` were already guarded by `if self.slc:` so there are no unintended side effects.

Validated with targeted test runs on `h2integrate/core/test/test_framework.py` and the full `h2integrate/core/test/` suite. All failures in `test_pose_optimization.py` are pre-existing OpenMDAO API mismatches unrelated to this PR.
- `h2integrate/converters/ammonia/test/test_ammonia_synloop_model.py::test_size_mode_outputs`

All targeted tests passed.

N/A


h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,126 @@ def _as_commodity_list(commodity):
...
+        # In the technology graph, merged edges (multiple L4 connections between the same pair)
+        # appear as a single edge with a list of commodities, so each unique (source, dest)
+        # pair counts as one stream.
+        in_degs_l4: dict[str, int] = {}
+        out_degs_l4: dict[str, int] = {}
+        for source, dest, commodity in self.technology_graph.edges(data="commodity"):
+            if commodity is None:
+                continue  # length-3 connections carry no commodity; skip them
+            out_degs_l4[source] = out_degs_l4.get(source, 0) + 1
+            in_degs_l4[dest] = in_degs_l4.get(dest, 0) + 1
```
elenya-grant, line 2293: I'm wondering if theres a way to do this with `dict(self.technology_graph.in_degree)` - but then removing non 4-length connections. I can't imagine that it'd actually make the code any simpler - so I like your approach.

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,126 @@ def _as_commodity_list(commodity):
...
+        # --- Check 2: storage technology topology ---
+        storage_techs = [k for k, v in self.tech_control_classifiers.items() if v == "storage"]
+        storage_upstream_techs: set[str] = set()
+        for storage_tech in storage_techs:
+            n_in = in_degs_l4.get(storage_tech, 0)
+            if n_in != 1:
+                raise ValueError(
+                    f"Storage technology {storage_tech!r} has {n_in} input connection(s) in "
+                    f"the technology graph but should have exactly 1."
+                )
```
elenya-grant, line 2311: is there a future case where a storage technology would have 2 different commodity inputs? Like - power and hydrogen? If possible - it'd be nice to check that theres only one input connection per commodity input to the storage.

elenya-grant: I think your code would be fine if (for some imaginary storage that takes electricity and hydrogen input) if the electricity and hydrogen come from the same upstream tech. I think that it'd cause an unnecessary error if electricity came from source_a and hydrogen came from source_b.

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,126 @@ def _as_commodity_list(commodity):
...
+                        f"(at most 2 output streams)."
+                    )
+
+        # --- Check 3: max 1 in/out for general technologies connected via length-4 edges ---
+        all_techs_in_l4 = set(in_degs_l4) | set(out_degs_l4)
+        for tech in all_techs_in_l4:
+            if "splitter" in tech or "combiner" in tech:
+                continue
+            if self.tech_control_classifiers.get(tech) == "storage":
+                continue  # already validated in check 2
```
elenya-grant: I think we should change the `_control_classifier` for the splitter to be `splitter` and the `_control_classifier` for the combiner to be named `combiner`. This way - we can check the control classifier rather than having logic based on technology names (i.e., a splitter tech has to have "splitter" in the name and a combiner has to have "combiner" in the name)

elenya-grant (commented):
really high-level initial review - thanks for addressing this issue! I think it'd be nice to see some tests with some "fake" systems that are more complex than is possible with models in H2I right now to test the limitations of this. For example, looking at systems with converter and storage technologies that have multiple input or output streams. 

Some examples that come to mind are:
- storage tech that has two input commodities (electricity and hydrogen)
- converter that has two input commodities and two output commodities

I also think that the examples in H2I will do a good job of testing this functionality. Thanks John! Will re-review again later (I want to give it a deeper look)

h2integrate/core/h2integrate_model.py
johnjasa: Thanks for this comment and context! I haven't changed anything here, but good to be aware.

h2integrate/core/h2integrate_model.py
johnjasa: This is a good point for future cases; I modified the behavior of this check. Now, a storage tech receiving electricity from `source_a` and hydrogen from `source_b` now passes cleanly - two edges but each commodity has exactly one source. The only invalid case is when the same commodity arrives from two different technologies, e.g. hydrogen from both `electrolyzer` and `h2_storage_bypass`, when we should just use a combiner.

h2integrate/core/h2integrate_model.py
johnjasa: Good call, I changed these two cases:
for `combiner`: "connector" ->"combiner"
for `splitter`: "connector" -> "splitter"

examples/36_nuclear_reactor_htse/plant_config.yaml
```diff
@@ -11,7 +11,7 @@ plant:
 # with the reverse definition.
 # this will naturally grow as we mature the interconnected tech
 technology_interconnections:
-  - [nuclear, htse, [heat_out, heat_in]]
+  - [nuclear, htse, heat, pipe]
```
elenya-grant: can you use the generic transport model here instead? I think that we should have more examples showing that model because I think it'll be very useful as we develop new models with new commodity streams. I don't think we should add heat as a supposed transport item to pipe.

elenya-grant: I've meant to make an issue about this - but I was talking to @genevievestarke a while back and we think that we should more examples using the generic transport model - and I think here is a good opportunity to do that.

h2integrate/transporters/pipe.py
```diff
@@ -41,6 +42,8 @@ def setup(self):
             units = "galUS/h"
         elif transport_item == "co2":
             units = "kg/h"
+        elif transport_item == "heat":
+            units = "kW"
```
elenya-grant: see my comment earlier for example 36. I don't think we should add heat to the pipe right now.

elenya-grant (commented):
just left some small comments

h2integrate/core/h2integrate_model.py
elenya-grant: I agree - that sounds great! Thank you! I think thats perfect.

elenya-grant: In PR #816, I added a fake system that handles another type of complex system I imagine we could have in H2I. I don't think that you have to use that specific fake system - but could be good to use something like that to test these checks for the connections into/out of a converter tech (h2integrate/control/control_strategies/system_level/test/test_slc_baseclass.py::test_find_converter_techs_fake_system)

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,159 @@ def _as_commodity_list(commodity):
...
+        for connection in technology_interconnections:
+            if len(connection) != 3:
+                continue
+            connected_parameter = connection[2]
+            if not isinstance(connected_parameter, list | tuple) or len(connected_parameter) != 2:
+                continue
+            source_param, dest_param = connected_parameter
+            if not isinstance(source_param, str) or not isinstance(dest_param, str):
+                continue
+            if source_param.endswith("_out") and dest_param.endswith("_in"):
```
elenya-grant: Not trying to nit-pick but I just remembered PR #774 and wonder if we should do:

```python
if source_param.split("[")[0].endswith("_out") and dest_param.split("[")[0].endswith("_in"):
```

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,159 @@ def _as_commodity_list(commodity):
...
+                if commodity_from_source == commodity_from_dest:
+                    source_tech, dest_tech = connection[0], connection[1]
+                    raise ValueError(
+                        f"Connection [{source_tech!r}, {dest_tech!r}, "
+                        f"[{source_param!r}, {dest_param!r}]] passes commodity "
+                        f"{commodity_from_source!r} between technologies using a "
+                        f"length-3 format. Use a length-4 connection instead: "
+                        f"[{source_tech!r}, {dest_tech!r}, {commodity_from_source!r}, "
+                        f"'<transport_tech>']. Use the generic transport component if "
+                        f"no specific transport is needed."
```
elenya-grant: Could we rephrase this part of the error message: "Use the generic transport component if no specific transport is needed"?

I think that the generic transport component is useful if a) no specific transport is needed and b) it's a commodity that isn't supported by cable or pipe models (perhaps because its a new commodity). I think the most simple re-phrasing would be "You can use the `GenericTransporterPerformanceModel` to transport {commodity_from_source!r}"? But I'm open to other ideas

h2integrate/core/test/test_framework.py
```diff
@@ -150,11 +153,306 @@ def test_check_tech_interconnections(subtests, temp_copy_of_example):
...
+        ["source_h2", "storage", "hydrogen", "pipe"],
+        ["storage", "h2_combiner", "hydrogen", "pipe"],
+    ]
+    classifiers = {
+        "storage": "storage",
+        "h2_combiner": "combiner",
+    }
+    fake = _make_fake_model(interconnections, classifiers)
+
+    with subtests.test("two-commodity storage passes validation"):
```
elenya-grant, line 307: I know that this use-case is still not something we imagine happening a ton yet - but in my head, this should be a valid system of interconnections but an error is thrown:

```python
interconnections = [
        ["source_elec", "storage", "electricity", "cable"],
        ["source_h2", "storage", "hydrogen", "pipe"],
        ["storage", "h2_combiner", "hydrogen", "pipe"],
        ["source_h2", "h2_combiner", "hydrogen", "pipe"],
        ["source_elec", "elec_combiner", "electricity", "cable"],
        ["storage", "elec_combiner", "electricity", "cable"],
]
classifiers = {
        "storage": "storage",
        "h2_combiner": "combiner",
        "elec_combiner": "combiner",
}
```

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,159 @@ def _as_commodity_list(commodity):
...
+            for upstream_tech in upstream_techs_l4:
+                storage_upstream_techs.add(upstream_tech)
+                n_out_upstream = out_degs_l4.get(upstream_tech, 0)
+                if n_out_upstream > 2:
+                    raise ValueError(
+                        f"Technology {upstream_tech!r} feeds storage technology "
+                        f"{storage_tech!r} but has {n_out_upstream} output connection(s). "
+                        f"It should connect only to {storage_tech!r} and a combiner "
+                        f"(at most 2 output streams)."
+                    )
```
elenya-grant: I think that here we should check that n_out_upstream based on the commodity

```suggestion
            for upstream_tech in upstream_techs_l4:
                storage_upstream_techs.add(upstream_tech)
                for commodity in self.technology_graph.edges[upstream_tech, storage_tech].get("commodity"):
                
                n_out_upstream = out_commodity_dests.get(upstream_tech, {}).get(commodity,0)
                if n_out_upstream > 2:
                    raise ValueError(
                        f"Technology {upstream_tech!r} feeds storage technology "
                        f"{storage_tech!r} but has {n_out_upstream} output connection(s). "
                        f"It should connect only to {storage_tech!r} and a combiner "
                        f"(at most 2 output streams)."
                    )
```

h2integrate/core/h2integrate_model.py
```diff
@@ -2219,6 +2220,159 @@ def _as_commodity_list(commodity):
...
+                        f"Storage technology {storage_tech!r} receives commodity "
+                        f"{commodity!r} from {n_sources} sources in the technology graph "
+                        f"but should receive it from at most 1."
+                    )
+            n_out = out_degs_l4.get(storage_tech, 0)
+            if n_out > 1:
+                raise ValueError(
+                    f"Storage technology {storage_tech!r} has {n_out} output connection(s) in "
+                    f"the technology graph but should have at most 1."
+                )
```
elenya-grant:
```suggestion
            for commodity, n_out in out_commodity_dests.get(storage_tech, {}).items():
                if n_out > 1:
                    raise ValueError(
                        f"Storage technology {storage_tech!r} has {n_out} output connection(s) in "
                        f"of commodity {commodity} but should have at most 1."
                    )
```

elenya-grant (changes requested):
Just left some comments on some edge cases that I think are valid but would cause errors when they shouldn't. I can share the tests I made to check these if you want. I will finish this review tommorrow! Looks good so far! Only some unlikely (at the moment) edge-cases that I found that this would throw unnecessary errors for

h2integrate/core/test/test_framework.py
johnjasa: Thanks for the example connections. The main difference here is that a singular storage technology accepts and stores both hydrogen and electricity, is that right? I don't immediately know of a physical system that behaves like that -- do you? I would argue we don't need to support this edge case at this moment, but let me know if you have a need or example case for it.

elenya-grant: you're right - I can't imagine a physical system that behaves like this. Change isn't needed!

h2integrate/core/test/test_framework.py
```diff
@@ -150,11 +153,324 @@ def test_check_tech_interconnections(subtests, temp_copy_of_example):
...
+        ["converter", "dest_b", "ammonia", "pipe"],  # ammonia goes to two places
+    ]
+    bad = _make_fake_model(bad_interconnections, {"converter": "dispatchable"})
+
+    with subtests.test("same output commodity to two destinations raises error"):
+        with pytest.raises(ValueError) as excinfo:
+            H2IntegrateModel._validate_technology_interconnections(bad)
+        err = str(excinfo.value)
+        assert "converter" in err
+        assert "destinations" in err
```
elenya-grant: could we have more detail in checking these error messages? I like there to be enough info here that someone could cntrl+f and find the error message in H2I.

h2integrate/core/test/test_framework.py
```diff
@@ -150,11 +153,324 @@ def test_check_tech_interconnections(subtests, temp_copy_of_example):
...
+        assert "length-4 connection" in err
+
+
+@pytest.mark.unit
+def test_validate_interconnections_length3_commodity_pair_with_slices_raises(subtests):
+    """A length-3 [X_out[...], X_in[...]] parameter pair must raise the
+    same guidance to use a length-4 commodity connection."""
+    interconnections = [
+        ["wind", "electrolyzer", ["electricity_out[0:4]", "electricity_in[0:4]"]],
+    ]
```
elenya-grant, line 453: thanks for adding this test in!

elenya-grant (commented):
Thanks for making these changes John! I haven't dug deeply into the logic introduced to handle splices (I can review it in more detail if you'd like). My only remaining suggestion is that I'd like to see more info on the strings for the error message assert statements. As-is, I think it'd be hard to debug a test failure and to be really confident the expected error message is being raised.

I imagine that as H2I develops, we will run into some use-cases that test the bounds of this logic but I think this is great for now!

h2integrate/core/test/test_framework.py
johnjasa: Thanks for the suggestion! I added more words to the asserts in the tests

elenya-grant (approved):
Looks good to me! thanks John!

PR #833, opened 2026-08-11
Reliability Prototype

RHammond2:
# Integration of Basic Reliability Modeling

This is an early stage (non-working) prototype of a Weibull-based reliability model for the `NaturalGasPerformanceModel`. 

For now, the model is designed to slot into any performance model using an hourly timestep over a single year. The key output is an array of availability (0 for downtime and 1 for operational) that is multiplied by the demand array.

## Working example

```python
from h2integrate.reliability import models

reliability_config = {
    "failure_model": "WeibullReliability",
    "maintenance_model": "FixedIntervalReliability",
    "failure_parameters": {
        "scale": 0.5,
        "shape": 1,
        "burn_in": 3.33,
        "n_components": 2,  # note: can also use a list of scale/shape values, this is just convenient when identical
        "downtime": {
            "model": "FixedDowntime",
            "hours": 5,
            "n_components": 2,
        },
    },
    "maintenance_parameters":{
        "frequency": [0.25, 1, 4],
        "downtime": {
            "model": "LogNormalDowntime",
            "mean": 2,
            "sigma": 1,
            "n_components": 3,
        },
    },
}

# we can also do the same thing for maintenance, the whole operation just hasn't been combined yet
failure = models.create_failure_model(reliability_config)

# energy-based availability without ramping of downtime events...yet
print(f"System-level availability (min of all components): {failure.system_availability.sum() / failure.system_availability.size:.4%}")
```

- [ ] Feature Enhancement
  - [ ] Framework
  - [x] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO: (see other feedback/considerations for what I'm considering or other aspects I could be missing)
- [x] Get the model working
- [x] Burn in period (skipping forward x number of years) - suggestion from @cfrontin 
- [x] Componentizable approach (matrix of failures vs single model)
- [x] Maintenance (fixed interval) and Failures (random)
  - [x] Integrate failure and maintenance into overarching system availability
- [x] Enable randomized downtime
- [ ] Tests
- [ ] Better docs
- [ ] Rethink approach to base models
- [x] Remove setup steps from individual models for easy addition of new models
- [ ] Ramping up/down production from 0-1/1-0
- [x] create array gt/ge validators outside of class
- [x] `n_timesteps` as an input to the model in place of using module level `N_TIMESTEPS = 8760`.
- [x] `dt` as an input to the model for duration of the simulation in place of `N_TIMESTEPS = 8760`.
- [ ] allow user to input `use_reliability` for a performance model to toggle its usage or check that a model has been provided if using during model initialization.
- [ ] Potentially convert `hours` to use a `dt` basis naming/input scheme

**Structural feedback**: Anything is welcome

**Implementation feedback**: Should the reliability slot into the performance in a more streamlined way? Any other feedback is welcome.

**Other feedback/considerations for the finalized PR**: It would be great to get feedback on the importance of the following items and any preferred approaches.
- ~better control over random seeding~ - kick the can way down the road
- ~uncertainty quantification~ - getting well ahead of ourselves
- ramping before/after downtime events
- passing `n_timesteps` from the plant configuration
- Calculate availability at initialization or manually run `model.calculate_availability()`?

- [ ] Documentation
- [ ] `CHANGELOG.md`

### Section 5.1: New Files
- `path/to/file.extension`

- `path/to/file.extension`

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


h2integrate/reliability/models.py
```diff
@@ -0,0 +1,94 @@
...
+        # simulation without burdening the memory usage.
+        self.time_to_failures = np.ceil(
+            self.config.rng.weibul(self.config.shape, size=30) * self.config.scale * N_TIMESTEPS
+        ).astype(int)
+        downtime_per_event = np.ceil(np.rng.normal(loc=self.config.downtime, size=30)).astype(int)
+        self.downtime_per_event = np.where(downtime_per_event >= 1, downtime_per_event, 1)
+
+    def calculate_availability(self):
+        """Determine the timing and duration of outages for a single year of simulation time."""
+        accumulated = 0
```
cfrontin: can we "burn-in" here (run some number of simulated years before starting the sampling accumulation)? that way we won't have jan 01 correlated with a brand new system every time

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,206 @@
...
+    TODO:
+        - how to pass n_timesteps through from plant?
+        - stabilize random generator/determine how to manage random seeding across library
+    """
+
+    scale: float = field(validator=validators.instance_of(float))
+    shape: float = field(validator=validators.instance_of(float))
+    downtime: Any = field(converter=generate_downtime_model)
+    downtime_per_event: np.ndarray = field(init=False, validator=validators.instance_of(np.ndarray))
+    availability: np.ndarray = field(
```
elenya-grant: could we make `n_timesteps` a configuration parameter? Then we could have `availability` be initialized in `__attrs_post_init__` as `np.ones(n_timesteps)`

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,206 @@
...
+            interval period to offset events from being based on January 1st in an 8760.
+        downtime (int | float | dict): Either fixed length of each downtime, in hours, or a
+            configuration dictionary for a downtime length model.
+
+    """
+
+    frequency: float = field(validator=(validators.instance_of((float, int)), validators.gt(0)))
+    downtime: Any = field(converter=generate_downtime_model)
+    downtime_per_event: np.ndarray = field(init=False, validator=validators.instance_of(np.ndarray))
+    availability: np.ndarray = field(
```
elenya-grant: same comment here about having `n_timesteps` as a configuration parameter.

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+
+    def __attrs_post_init__(self):
+        self.create_downtime_events()
+        self.calculate_availability()
+
+    def sample_events(self):
+        frequency = np.array([0.1, 20, 30]).reshape(-1, 1)
+        interval = np.ceil(8760 / frequency).astype(int)
+        first_occurrence = rng.integers(0, np.where(interval > 8760, 8760, interval))
```
elenya-grant: should these 8760's be `n_timesteps`?

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,206 @@
...
+        self.availability[self.first_occurrence : end] = 0
+        accumulated = end
+        while (start := accumulated + self.interval) < N_TIMESTEPS:
+            duration, self.downtime_per_event = (
+                self.downtime_per_event[0],
+                self.downtime_per_event[1:],
+            )
+            # TODO: handle start > 8760
+            end = start + duration
+            end = np.where(end > 8760, 8760, end)
```
elenya-grant: same here - should 8760 be `n_timesteps` or `N_TIMESTEPS` (as you have it now?)

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,206 @@
...
+@define(kw_only=True)
+class FixedIntervalReliability(BaseReliability):
+    """Basic fixed interval downtime reliability model.
+
+    Args:
+        frequency (int | float | array-like): The annual frequency of events, e.g., 4 is equivalent
+            to a quarterly downtime event and 0.25 is equivalent to an every 4 years downtime event.
+            For all events the timing of the first event will be sampled within the first year or
+            interval period to offset events from being based on January 1st in an 8760.
+        downtime (int | float | dict): Either fixed length of each downtime, in hours, or a
```
elenya-grant, line 449: could we name this `downtime_hrs` or something? Also - can these models be updated to handle varying timesteps? It seems like a lot of logic is intended for hourly? If so - I think we should should make `dt` an input parameter too.

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
```diff
@@ -75,6 +78,12 @@ def setup(self):
             merge_shared_inputs(self.options["tech_config"]["model_inputs"], "performance"),
             additional_cls_name=self.__class__.__name__,
         )
+        if self.options["tech_config"]["model_inputs"]["reliability"]:
```
elenya-grant: do we need to use `self.options["tech_config"]["model_inputs"].get("reliability", False)` here?

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
```diff
@@ -67,6 +68,8 @@ def initialize(self):
         self.commodity = "electricity"
         self.commodity_rate_units = "MW"
         self.commodity_amount_units = "MW*h"
+        self.reliability_model = None
+        self.use_reliability = False
```
elenya-grant: Could these be parameters in the `NaturalGasPerformanceConfig`? So a user can input `reliability_model` and `use_reliability`? Where the `__attrs_post_init__` checks that `reliability_model` is provided if `use_reliability` is True?

elenya-grant (commented):
Left some big-picture comments! Happy to chat through anything if you'd like! Thanks!

h2integrate/reliability/models.py
RHammond2: Apparently I never posted my comment that this is a great idea, but it's now implemented because it's a good idea!

h2integrate/reliability/models.py
RHammond2: This could be tricky, but I'll add it to the to do section.

h2integrate/reliability/models.py
RHammond2: The generation of availability is now done in `calculate_availability` to make it easier to automate, and adding the `n_timesteps` to the to do section.

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
RHammond2: Not exactly, but that did get me thinking on a better way to use `use_reliability`, so I appreciate the question!

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
RHammond2: That's a good question, it didn't dawn on me that a user could want to provide a definition, but not use it. Though it could be easier to iterate on a problem by simply turning it on/off instead of commenting out a whole section of the inputs. I'll also add this to the to do section.

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
```diff
@@ -160,6 +173,8 @@ def compute(self, inputs, outputs):
             inputs["electricity_command_value"],
         )
         natural_gas_demand = electricity_command_value * heat_rate_mmbtu_per_mwh
+        if self.use_reliability:
+            natural_gas_demand * self.reliability_model.availability
```
johnjasa, line 178: naive question, does this need to be assigned to some variable?

h2integrate/reliability/test/test_models.py
johnjasa, line 1: empty file on purpose? I assume you're gonna build it out

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,483 @@
...
+from h2integrate.reliability.utilities import (
+    update_dimensions,
+    calculate_annual_timesteps,
+    calculate_hourly_timesteps,
+    calculate_simulation_years,
+)
+
+
+# generated from np.random.SeedSequence().entropy
+rng = np.random.default_rng(279299947538423226929715083173412195503)
```
johnjasa, line 19: did you use an rng to get this rng seed? :D

h2integrate/reliability/utilities.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+
+
+SECONDS_IN_HOUR = 60 * 60
+SECONDS_IN_YEAR = 8760 * SECONDS_IN_HOUR
+
+
+def update_dimensions(n_components: int, *args: np.ndarray):
```
johnjasa, line 8: nitpick: could this title be slightly more descriptive? maybe `expand_array_dimensions` (if that's valid) or something that elucidates the type of updating going on?

h2integrate/reliability/utilities.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def calculate_annual_timesteps(value, self_) -> float:
+    """Calculates the number of timesteps in a year from the :py:attr:`self_.dt`, rounded up."""
+    return int(np.ceil(SECONDS_IN_YEAR / self_.dt))
+
+
+def calculate_hourly_timesteps(value, self_) -> float:
+    """Calculates the number of timesteps in an hour from the :py:attr:`self_.dt`, rounded up."""
+    return int(np.ceil(SECONDS_IN_HOUR / self_.dt))
```
johnjasa, line 50: These are good and concise functions that should live here for now. In the future, especially as folks like @jaredthomas68 work more on non-annual simulation and different `dt` values, I imagine these could get moved up above the `reliability` folder level and into a more shared utilities file.

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,483 @@
...
+
+    Attributes:
+        time_to_failures (np.ndarray): Array of the hours to the next failure for each modeled
+            component. Generated by each subclass' :py:meth:`sample_events`.
+        downtime_per_event (np.ndarray): Number of hours of downtime corresponding to each downtime
+            event in :py:attr:`time_to_failures`.
+        availability (np.ndarray): Operational ratio of each modeled component for every time step
+            of the simulation.
+        system_availability (np.ndarray): Minimum operational ratio of all components for every time
+            time step of the simulation.
```
johnjasa, line 152: I wonder if there are more clear names for these availabilities, they're remarkably similar at first glance. Maybe `component_availability` instead of just `availability`? Or `min_system_availability` for `system_availability`? I'm not sure on what the names _should_ be or if it's an issue, but it took me a sec to tease 'em apart

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,483 @@
...
+        if self.failures is not None:
+            failure_availability = self.failures.system_availability
+        if self.maintenance is not None:
+            maintenance_availability = self.maintenance.system_availability
+        self.availability = np.minimum(failure_availability, maintenance_availability)
+
+
+@define(kw_only=True)
+class FixedDowntime(BaseDowntime):
+    """Basic fixed downtime model for generating the length of downtime for a given event.
```
johnjasa, line 328: Could you define fixed here?

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,483 @@
...
+        self.n_components, self.hours = update_dimensions(self.n_components, self.hours)
+
+    def sample_downtime(self):
+        """Return an array of 100 :py:attr:`hours`."""
+        return np.ones((1, 100), dtype=int) * self.hours * self.simulation.n_timesteps_in_hour
+
+
+@define(kw_only=True)
+class UniformDowntime(BaseDowntime):
+    """Basic uniform downtime model for generating the length of downtime for a given event.
```
johnjasa, line 351: Could you define uniform here?

h2integrate/reliability/models.py
johnjasa: i.e. give an example if it's _always_ a certain amount of downtime (my understanding) or can vary

h2integrate/reliability/models.py
```diff
@@ -0,0 +1,483 @@
...
+        validator=(validators.instance_of(np.ndarray), array_ge(1)),
+    )
+    n_components: int = field(default=1, validator=(validators.instance_of(int), validators.ge(1)))
+
+    def __attrs_post_init__(self):
+        self.n_components, self.hours = update_dimensions(self.n_components, self.hours)
+
+    def sample_downtime(self):
+        """Return an array of 100 :py:attr:`hours`."""
+        return np.ones((1, 100), dtype=int) * self.hours * self.simulation.n_timesteps_in_hour
```
johnjasa, line 346: my own curiosity: is it always 100? could/should this be a user-set value later on?

johnjasa (changes requested):
Thanks for this, @RHammond2! I appreciate so much of it, including your good use of class abstraction, functions, and clear naming throughout. I also appreciate that you didn't do _all_ of the reliability work in one PR and are purposefully building it out.

I have small comments, nothing hugely blocking at all, but I'm more making sure I come back to this after you make some more of your planned updates!

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
RHammond2: If it's a class output or OpenMDAO output that is desired, then yes, but it is also available through `NaturalGasPerformanceModel.reliability.availability`.

h2integrate/reliability/test/test_models.py
RHammond2: Yes! I'm going to resolve this though since I'm going to split the model tests out for a little organizational sanity.

h2integrate/reliability/models.py
RHammond2: lol, no literally just `np.random.SeedSequence().entropy` as stated. Adding a comment though to explain the why a little better though.

h2integrate/reliability/utilities.py
RHammond2: This one got me a little bit too because we're either updating the array or we're updating `n_components` to match the array so we're not always expanding the array. I do agree the name is a bit lackluster though so I'll think on it a bit more, and am open to ideas.

h2integrate/reliability/utilities.py
RHammond2: Agreed. I almost wonder if a separate PR should just preemptively create a simulation class for all these definitions kept in dictionaries. I was tempted by it, but figured that was getting into a bit of scope creep given there is a lot more than `n_timesteps` and `dt`.

h2integrate/reliability/models.py
RHammond2: Good catch, `availability` vs `system_availability` was a leftover from iterating. I've just updated the base models to use `component_availability`, though the overarching model will still use `availability` until I add in system vs component level metrics.

h2integrate/reliability/models.py
RHammond2: Now reads as "fixed duration."

h2integrate/reliability/models.py
RHammond2: I chose 100 to fall somewhere in the long enough to likely not need to regenerate events later, but to have enough in case someone models highly frequent events. Basically some balance between computational and memory efficiency even though these models are all extremely negligible relative to the H2I as a whole. So, no I don't plan on giving users access, rather than just enforcing the number of events.

h2integrate/reliability/models.py
RHammond2: Yes, thanks again for flagging!

h2integrate/converters/natural_gas/natural_gas_cc_ct.py
RHammond2: I forgot to comment, but this is now included with control over the results application being provided in the performance model itself.

h2integrate/reliability/models.py
RHammond2: The downtime durations are now individual models, with `dt` and `n_timesteps` fully incorporated.

h2integrate/reliability/models.py
RHammond2: I should also add that if the events are exhausted, then new ones are automatically generated as part of the availability calculation.

PR #844, opened 2026-08-14
WIP: Heterogeneous-Commodity System-Level Control (Static and Measured Conversion Ratios)

johnjasa:
# Heterogeneous-Commodity System-Level Control (Static and Measured Conversion Ratios)

This PR teaches the system-level control (SLC) framework to handle heterogeneous commodity systems, where the demand is for one commodity (for example ammonia) but the technologies that ultimately satisfy it produce and consume other commodities upstream (for example electricity into hydrogen into ammonia). The controller now translates demand for the demand commodity into set-points for each upstream technology by walking backward across converters using per-technology conversion ratios. Ratios are measured from each converter's consumed and produced streams when available and fall back to static ratios defined in the tech config.

Summary of changes:
1. Add backward demand propagation to `SystemLevelControlBase` so demand for one commodity is translated into derived demand for upstream input commodities.
2. Detect converters robustly from the technology graph, including converter chains such as electricity to hydrogen to ammonia.
3. Prefer measured conversion ratios computed from each converter's `{commodity}_consumed` and produced streams per timestep, falling back to a static ratio from the tech config, and emit a one-time warning only when neither a measurement nor a static ratio is available.
4. Register and connect the converters' `{commodity}_consumed` outputs to the controller so measured ratios are available during dispatch.
5. Refactor the three concrete controllers onto a shared per-commodity dispatch path with strategy-specific dispatchable hooks.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Add converter detection and backward demand propagation to the base class
- [x] Read static conversion ratios from the tech config
- [x] Compute measured conversion ratios from converter consumed and produced streams
- [x] Register and connect converter `{commodity}_consumed` outputs to the controller
- [x] Keep propagation opt-in with a one-time warning when no ratio is available
- [x] Refactor the three controllers onto the shared dispatch path
- [x] Add unit tests for propagation, chaining, mixed sources, measured ratios, and fallbacks
- [x] Update CHANGELOG

**Structural feedback**:
Confirm the static-ratio location in the tech config
(`technologies.<tech>.model_inputs.control_parameters.conversion_ratios.<in>_per_<out>`)
is the right home for these values.

**Implementation feedback**:
Confirm that the opt-in, warn-and-skip fallback for missing ratios is the desired
backward-compatibility behavior versus a hard error.

- [x] Documentation
- [x] `CHANGELOG.md`

Backward propagation is opt-in. A converter whose input commodity has no controller-managed producer (for example a feedstock-supplied stream) is skipped without a ratio. A converter whose input commodity does have controllable producers but no configured ratio is also skipped, leaving upstream techs at their default dispatch (the legacy behavior), and emits a one-time warning so the missing ratio is discoverable. This keeps existing single-commodity and electricity-to-hydrogen examples unchanged.

Ratios are interpreted so that `input_rate = output_rate * ratio` in each commodity's rate units (for example a `51` electricity-per-hydrogen ratio translates a hydrogen production rate in `kg/h` into an electricity demand in `kW`). When a converter's `{commodity}_consumed` output is connected, the controller measures the ratio per timestep as consumed over produced, guards the divide where production is zero, and falls back to the static or nominal ratio there. This captures nonlinear and time-varying efficiencies without requiring a hand-tuned static ratio.

N/A


h2integrate/core/h2integrate_model.py
```diff
@@ -632,6 +634,22 @@ def _classify_slc_technologies(self):
...
+            tech
+            for tech, classifier in upstream_tech_control_classifiers.items()
+            if classifier in converter_input_classifiers
+        }
+        produced_by_tech = defaultdict(set)
+        for tech, commodity in tech_to_commodity:
+            produced_by_tech[tech].add(commodity)
+        slc_topology["converters"] = detect_commodity_converters(
+            upstream_tech_graph, converter_input_techs, produced_by_tech
+        )
```
elenya-grant: Nitpick - curious whether this should just be done in `SystemLevelControlBase`? Is there a reason you've called this within H2I rather than in the SLC baseclass? My main concern is that its already difficult to set-up tests for the SLC models given its dependence on information from H2I and that it may be better if this was just called within the SLC model.

h2integrate/core/h2integrate_model.py
```diff
@@ -632,6 +634,22 @@ def _classify_slc_technologies(self):
...
+        # Detect commodity converters (e.g. electrolyzer: electricity -> hydrogen)
+        # so the same converter set drives backward demand propagation in the
+        # controller and the consumption-signal connections made below
+        converter_input_classifiers = {"fixed", "flexible", "dispatchable", "storage"}
+        converter_input_techs = {
+            tech
+            for tech, classifier in upstream_tech_control_classifiers.items()
+            if classifier in converter_input_classifiers
+        }
+        produced_by_tech = defaultdict(set)
```
elenya-grant: I'm not familiar with `defaultdict` - how it this `defaultdict(set)` different than just initializing a dictionary as `dict()` or `{}`?

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+        # Read each converter's static input-per-output ratio from the plant
+        # config's system-level-control ``conversion_parameters`` block, keyed by
+        # tech name with ``<in_commodity>_per_<out_commodity>`` entries.
+        self.conversion_ratios = {}
+        self._missing_ratio_warned = set()
+        conversion_parameters = (
+            self.options["plant_config"]["system_level_control"]
+            .get("control_parameters", {})
+            .get("conversion_parameters", {})
+        )
```
elenya-grant: if we want to go the route of using user-input conversion parameters - then I'd rather it be defined in a configuration class (I know that may not be super easy since it depends on the system) but I think it'd be more clear than here. This would likely mean creating a base configuration class for SLC.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+            ratios = conversion_parameters.get(tech_name, {})
+            key = f"{in_commodity}_per_{out_commodity}"
+            if key in ratios:
+                self.conversion_ratios[(tech_name, in_commodity, out_commodity)] = float(
+                    ratios[key]
+                )
+
+        # Register a measured-consumption input for every converter whose input
+        # commodity has a controller-managed producer. H2IntegrateModel wires
+        # the converter's ``{in_commodity}_consumed`` output to this input so
```
elenya-grant: I think the usage of ``{in_commodity}_consumed`` is a little confusing - it makes me think that (for example) `electrolyzer.electricity_consumed` is an input to the SLC (which it isn't). I think that it'd be more clear to say "the input commodity to the converter"

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+                unit_kwargs = {"units": None}
+
+            consumed_name = f"{tech_name}_{in_commodity}_consumed"
+            self.add_input(
+                consumed_name,
+                val=np.full(self.n_timesteps, np.nan),
+                shape=self.n_timesteps,
+                desc=f"Measured {in_commodity} consumed by converter {tech_name}",
+                **unit_kwargs,
+            )
```
elenya-grant: Oh! Ignore my above comment - you actually ARE inputting the amount of some input commodity consumed. I'm not sure my overall feelings on this but here's some thoughts:

- I am not totally confident that `commodity_consumed` is a standardized output in H2I (it's not enforced in the PerformanceModelBaseClass). But - I think that it is included in most models so this should be fine
- I thought that this would be helpful for determining which input commodity is limiting some commodity production from a converter (i.e., is the ammonia production limited by nitrogen, hydrogen, or electricity) but I don't think it necessarily does by itself (maybe I'll learn more as I keep going in the code)
- I think this could be useful with splitters potentially but I'm not totally sure (that's a whole can of worms we haven't made a plan for yet)
- I think this would be paired well with the approach I had (where it calculates the total in_commodity to a converter based on its upstream technologies) - since these two approaches could help figure out what is the limiting input and which inputs are being provided in excess.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+            else:
+                unit_kwargs = {"units": None}
+
+            consumed_name = f"{tech_name}_{in_commodity}_consumed"
+            self.add_input(
+                consumed_name,
+                val=np.full(self.n_timesteps, np.nan),
+                shape=self.n_timesteps,
+                desc=f"Measured {in_commodity} consumed by converter {tech_name}",
+                **unit_kwargs,
```
elenya-grant: should this be moved to a separate method? For adding inputs we have a lot of methods named `_setup_something` - I think that perhaps that'd make sense here.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+            if commodity in self._get_commodity_for_tech(flexible_tech):
+                updated = self._subtract_flexible(flexible_tech, demand, commodity, inputs, outputs)
+                if updated is not None:
+                    demand = updated
+
+        # 3. Storage techs: split residual demand evenly and dispatch
+        n_storage = len(
+            [s for s in self.storage_techs if commodity in self._get_commodity_for_tech(s)]
+        )
+        for storage_tech in self.storage_techs:
```
elenya-grant: my biggest gripe with this method (not new in your PR) is that storage is done before dispatchable - which means like in an electrolyzer/h2 storage subsystem, the h2 storage is not dispatched to account for the hydrogen produced from the electrolyzer. Anywayyy - I kind of like that being obvious at a higher level. Not asking for any changes - just sharing some thoughts.

h2integrate/control/control_strategies/system_level/system_level_control_base.py
```diff
@@ -588,6 +649,389 @@ def _get_commodity_for_tech(self, tech_name):
...
+
+            self._dispatch_commodity(commodity, demand.copy(), inputs, outputs)
+
+            # Translate committed converter output into upstream input demand
+            for in_commodity, tech_name, out_commodity in converters_by_output.get(commodity, []):
+                self._accumulate_derived_demand(
+                    derived_demand, tech_name, in_commodity, out_commodity, inputs, outputs
+                )
+
+    def _dispatch_commodity(self, commodity, demand, inputs, outputs):
```
elenya-grant: here's an example use-case that I'm not sure this would work as-intended for - wind+solar connected to electrolyzer, electrolyzer connected to h2 fuel cell, demanded commodity is electricity from fuel cell. Would this method dispatch the wind, solar, and fuel cell all together? Although the wind+solar demand should include the conversion from the electrolyzer

PR #847, opened 2026-08-17
Added automerge enabled to CI testing

johnjasa:
# Enabled automerge trigger for CI

A reasonable assumption would be that if `automerge` is enabled for a PR, we'd want the full test suite to be run again.
This PR changes that to be true.


kbrunik (approved):
so fast (whispers: zoom zoom)

PR #849, opened 2026-08-18
Addition to GenericConverterCostModel and Schema Update

### LINKED ISSUE #841: Update all schemas
# Update all schemas

It's been a while since we've revisited the schemas and their checking abilities.

## Proposed solution

We should update all keys, checks, required and optional items, to better reflect the config files that exist in H2I today and are used.

elenya-grant:
# Addition to GenericConverterCostModel and Schema Update

Added 3 new configuration parameters to the `GenericConverterCostConfig` and 3 new inputs to `GenericConverterCostModel`. Below shows `config_parameter` -> `input_name`. These new inputs do not scale with the capacity of the technology - they just are flat cost additions. 
- `additional_capex_USD` -> `extra_capex`
- `additional_opex_USD_per_year` -> `extra_opex`
- `additional_varopex_USD_per_year` -> `extra_varopex`

This is an alternative approach to adding additional costs (original approach was shown in PR #838, which was closed without merging to try this method instead)

Also did some updating of the schemas! Please see Section 2 for requested feedback!

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [x] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR
- [x] Complete Section 8: New Model Checklist (if applicable)

### TODO:
- [ ] Update changelog

**Structural feedback**:

**Implementation feedback**:

**Schema feedback**:
- Thoughts on having defaults in the schemas for `name` and `description`? I don't think these tend to be used anywhere or are particularly useful besides in examples/personal workflow.
- What level of detail should `finance_groups`, `finance_subgroups`, and `cost_adjustment_parameters` have in the plant schema? 
   - If we include `finance_subgroups` - should I also put in the parts of that (commodity, commodity_stream, finance_groups, technologies, etc?)
      - If we include the `commodity_stream` as required in the finance subgroups - then we get a json schema ValidationError instead of our nice error message. Should we have commodity_stream be required or not? Aka - which error message do we prefer? (see the NOTE I added in `h2integrate/core/test/test_framework.py::test_finance_subgroup_missing_commodity_stream_raises`
- should I include the `dispatch_rule_set` and `dispatch_parameters` in the tech_schema?
- should we allow `control_parameters`, `shared_parameters`, `performance_parameters`, etc to be of type 'object' OR 'null'? Or should we enforce object? (brought up with test `examples/test/test_all_examples.py::test_pyomo_heuristic_dispatch_example`)
- What other things are missing from the driver_schema? 
- Should `plant` be required in the plant config?

**Other feedback**:

- Any feedback on in the new attributes of `GenericConverterCostConfig` and new inputs to `GenericConverterCostModel`?

- [x] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

Resolves #841

Created Issue #852 

### Section 5.1: New Files
N/A

- `h2integrate/converters/generic_converter_cost.py`
  - `GenericConverterCostConfig`: Added 3 new configuration parameters, they default to 0
  - `GenericConverterCostModel`: 
     - `setup()`: added 3 new inputs
     - `compute()`: updated calculation of costs to include the extra costs
- `h2integrate/converters/test/test_generic_converter_cost.py`
   - `test_generic_converter_cost_extra_costs`: new test with using the additional cost inputs

**Schema updates:**
- `h2integrate/core/inputs/driver_schema.yaml`: added section for recorder (optional). Added default values for `name` and `description`.
- `h2integrate/core/inputs/plant_schema.yaml`: added section for `resource_to_tech_connections`. Added default values for `name` and `description`.
- `h2integrate/core/inputs/tech_schema.yaml`: added section for `finance_model` and `model_inputs`. Removed `details` and `financial_model`. Added default values for `name` and `description`.

`examples/test/test_all_examples.py::test_pyomo_heuristic_dispatch_example`: the subtest "Ensure no-tech name validates" was setting `control_parameters` to None, which failed the schema validation because it has to be an object. I replaced `tech["technologies"]["battery"]["model_inputs"]["control_parameters"]=None` with `tech["technologies"]["battery"]["model_inputs"].pop("control_parameters")`. Alternative would be to update the tech schema so that the `x_parameters` under `model_inputs` are allowed to be objects or null (but I don't really think they should be able to be empty). This is also noted in the requested feedback for reviewers.

`h2integrate/core/test/test_framework.py::test_finance_subgroup_missing_commodity_stream_raises`: error changed to a json schema validation error - please see the note I added in this test and the requested feedback for reviewers to see if this is a change we want to make.


h2integrate/converters/generic_converter_cost.py
```diff
@@ -93,6 +105,26 @@ def setup(self):
             desc="Unit Variable O&M",
         )
 
+        self.add_input(
+            "extra_capex",
```
genevievestarke: I think this all makes sense, but I think calling it `additional_constant_capex`, like in the config, makes more sense to me than `extra_capex`.

h2integrate/converters/generic_converter_cost.py
```diff
@@ -112,13 +144,13 @@ def setup(self):
...
+            opex = inputs[f"rated_{self.config.commodity}_production"] * inputs["unit_opex"]
         else:
-            outputs["OpEx"] = tot_capex * inputs["fixed_opex_ratio"]
+            opex = tot_capex * inputs["fixed_opex_ratio"]
+
+        outputs["OpEx"] = opex + inputs["extra_opex"]
         outputs["VarOpEx"] = (
             inputs[f"annual_{self.config.commodity}_produced"] * inputs["unit_varopex"]
-        )
+        ) + inputs["extra_varopex"]
```
genevievestarke: Just to test my understanding here: since this is variable opex, should it still be based on how much the technology produced rather than being a flat value added on? Is variable opex handled differently than regular opex in profast? Kind of confusing for me merging these two ideas!

genevievestarke (commented):
No major comments, just some small thoughts! 
For the schema, I'd like if if we could get the cost, performance, shared, etc. dictionaries in the schema, and just have them be empty if there's nothing in them. Thoughts on that?

h2integrate/converters/generic_converter_cost.py
elenya-grant: Yes - I see where you're coming free. Variable opex is handled differently in ProFAST. Fixed OpEx is input to ProFAST as a fixed item, whereas variable OpEx is input as a feedstock item. I suppose that I figure it's nice to have as a "just in case the use-case comes up" but I agree that it doesn't totally fit the definition of "variable" opex if its a flat cost adder. I suppose, it adds some level of flexibility to this cost model but is a bit confusing. I can remove this as an input if we can't think of a good use-case for it.

elenya-grant: I think the main use-case is that the additional variable opex value can be negative. Negative VarOpEx values are input to profast as coproducts. So - I think this could be helpful for any cases where someone wants to have some co-product cost or reduce the variable opex by some flat dollar amount.

johnjasa (approved):

PR #850, opened 2026-08-18
Allow demand components as non-topology-constraining destinations in technology interconnections

johnjasa:
# Allow demand components as non-topology-constraining destinations in technology interconnections

1. Demand-classified components (subclasses of `DemandComponentBase`) are now exempt from the per-commodity output-destination count in `_validate_technology_interconnections`
2. Added a unit test to cover the valid case and confirm the two-real-consumer error is still raised.

- [x] Feature Enhancement
  - [x] Framework

N/A - not a draft PR.

- [x] Tests pass
- [x] Documentation
- [x] `CHANGELOG.md`
  - [x] After the above, a hyperlink has been provided to the PR

## Section 4: Related Issues

N/A

### Section 5.1: New Files

N/A

  - `_validate_technology_interconnections`: Skip demand-classified destinations when building `out_commodity_dests` so they do not count against the source's per-commodity output limit.
  - `test_validate_interconnections_demand_component_not_counted_as_destination`: New unit test covering valid source-to-consumer-plus-demand-reporter topology and confirming the two-real-consumer error is unchanged.

N/A

New unit test passes. Existing unit tests for `_validate_technology_interconnections` are unaffected.

### DIFF
--- a/h2integrate/core/h2integrate_model.py
+++ b/h2integrate/core/h2integrate_model.py
@@ -2396,6 +2396,66 @@ class H2IntegrateModel:
                         f"send it to at most 1. Consider using a splitter component."
                     )
 
+        # --- Check 4: prevent commodity double-counting via demand components ---
+        # A demand component may receive a commodity and pass it on to a real consumer
+        # (e.g. acting as a profile regularizer). However, if a source does this it
+        # must NOT also send the same commodity directly to another real consumer,
+        # because the flow would be counted twice.
+        #
+        # Valid:   source -> demand_comp -> real_consumer   (single path through demand)
+        # Valid:   source -> demand_comp (pure observer)
+        #          source -> real_consumer
+        # Invalid: source -> real_consumer_A                (competing direct path)
+        #          source -> demand_comp -> real_consumer_B (and also via demand)
+        #
+        # The check is transitive: a demand component "reaches a real consumer" even
+        # when the path passes through a chain of other demand components first.
+
+        def _demand_reaches_real_consumer(demand_tech: str) -> bool:
+            """Return True if demand_tech can reach a non-demand tech via L4 edges."""
+            visited: set[str] = set()
+            stack = [demand_tech]
+            while stack:
+                node = stack.pop()
+                if node in visited:
+                    continue
+                visited.add(node)
+                for _, d, c in self.technology_graph.out_edges(node, data="commodity"):
+                    if not c:
+                        continue
+                    if self.tech_control_classifiers.get(d) != "demand":
+                        return True
+                    stack.append(d)
+            return False
+
+        # Build per-(source, commodity) destination lists from L4 edges.
+        source_commodity_dests: dict[tuple[str, str], list[str]] = {}
+        for source, dest, commodity in self.technology_graph.edges(data="commodity"):
+            if not commodity:
+                continue
+            for c in commodity:
+                source_commodity_dests.setdefault((source, c), []).append(dest)
+
+        for (source, commodity), dests in source_commodity_dests.items():
+            direct_real_dests = [
+                d for d in dests if self.tech_control_classifiers.get(d) != "demand"
+            ]
+            outputting_demand_dests = [
+                d
+                for d in dests
+                if self.tech_control_classifiers.get(d) == "demand"
+                and _demand_reaches_real_consumer(d)
+            ]
+            if direct_real_dests and outputting_demand_dests:
+                raise ValueError(
+                    f"Technology {source!r} sends commodity {commodity!r} both directly "
+                    f"to {direct_real_dests} and to demand component(s) "
+                    f"{outputting_demand_dests} that re-emit the commodity to real "
+                    f"consumers. This would double-count the commodity flow. Either route "
+                    f"all {commodity!r} from {source!r} through the demand component, or "
+                    f"connect the downstream consumers directly to {source!r} instead."
+                )
+
     def _check_tech_connections(self):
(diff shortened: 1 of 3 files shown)


h2integrate/core/test/test_framework.py
```diff
@@ -462,6 +462,54 @@ def test_validate_interconnections_length3_commodity_pair_with_slices_raises(sub
...
+    """
+    interconnections_valid = [
+        ["wind", "electrolyzer", "electricity", "cable"],
+        ["wind", "demand_reporter", "electricity", "cable"],
+    ]
+    classifiers_valid = {
+        "wind": "flexible",
+        "electrolyzer": "dispatchable",
+        "demand_reporter": "demand",
+    }
```
elenya-grant, line 487: Okay - I was thinking of ways that people could use demand components to double-count electricity and it took a while but I did figure it out. Below is an example where some electricity would be double-counted from the wind. This doesn't throw an error:

```python
    interconnections_invalid = [
        ["wind", "electrolyzer", "electricity", "cable"],
        ["wind", "demand_reporter", "electricity", "cable"],
        ["demand_reporter", "grid_sell", "electricity", "cable"],
    ]
    classifiers_invalid = {
        "wind": "flexible",
        "electrolyzer": "dispatchable",
        "demand_reporter": "demand",
        "grid_sell": "dispatchable",
    }
```


I have a few thoughts on ways to catch-this. Since the length 4 tuple is enforced, we could do an extra check on the demand components and basically figure out which ones are being used for _metrics_ vs which ones are being use for "_performance_". Demand technologies that are used for _metrics_ should **not be the source** of a length-4 connection (this is how double-counting commodity can happen) **_unless_** the destination is another demand component (daisy-chained demand components).

h2integrate/core/h2integrate_model.py
```diff
@@ -2302,13 +2302,19 @@ def _validate_technology_interconnections(self):
...
-                    {c: out_commodity_dests.get(source, {}).get(c, 0) + 1}
-                )
+                # Demand-classified techs are observers/sinks (they report on a commodity
+                # stream but do not consume it in a topology sense). Exclude them from the
+                # source's output-destination count so that a source may simultaneously
+                # feed a real consumer and a reporting/demand component without triggering
+                # the multi-destination error in Check 3.
+                if dest_classifier != "demand":
+                    current = out_commodity_dests.setdefault(source, {})
+                    current[c] = current.get(c, 0) + 1
```
elenya-grant: I don't think this logic maybe works as intended for daisy-chained demand components.

elenya-grant (commented):
Unsure how much it's worth digging into the edge-case I pointed out in my one-comment. Let me know if you think its worth addressing or not!

h2integrate/core/test/test_framework.py
johnjasa: Thanks for writing this out and your clear example! I agree that this is an important case that (given the checking we're currently doing) should be caught but isn't with my initial changes.

I've updated the logic to allow for demand components to have passthrough behavior if and only if the inputted commodity stream isn't also used elsewhere. The logic is a bit complex to handle the cases correctly, but take a look at the test to see if it's behaving as you expect!

h2integrate/core/test/test_framework.py
```diff
@@ -462,6 +462,135 @@ def test_validate_interconnections_length3_commodity_pair_with_slices_raises(sub
...
+
+    with subtests.test("source to direct real consumer AND outputting demand raises error"):
+        with pytest.raises(ValueError) as excinfo:
+            H2IntegrateModel._validate_technology_interconnections(fake_double_count)
+        err = str(excinfo.value)
+        assert "wind" in err
+        assert "double-count" in err
+
+    # Daisy-chain through two demand components then to a real consumer is valid as
+    # long as wind does not also have a competing direct path to a real consumer.
```
elenya-grant, line 553: Nice edge case that I didn't think of!

elenya-grant (approved):
Thanks John! This looks great! I think your tests capture the cases I thought were necessary to test. I haven't dug into the code too much (I'm trusting the tests) - but one other case I thought _may_ be good to test is if something outputs multiple commodities (one commodity goes to a non-demand component and another output commodity goes to a demand component). 

I tested some cases locally like this and it worked properly (see below)! So nice job! Not asking for changes - mostly commenting that it does work for multi-commodity-output cases!

The cases I tested (in case anyone is curious) are below:

1. Invalid (oxygen goes to two consumers):

```python
interconnections_invalid = [
["electrolyzer","haber-bosch", "hydrogen", "pipe"],
["electrolyzer","haber-bosch", "oxygen", "pipe"],
["electrolyzer","demand_reporter", "oxygen", "pipe"],
["demand_reporter", "oxygen_sell", "oxygen", "cable"],
]
classifiers_invalid = {
    "electrolyzer": "dispatchable",
    "demand_reporter": "demand",
    "oxygen_sell": "dispatchable",
}
```

2. Valid (oxygen and hydrogen each go to one consumer):

```python

interconnections_valid = [
    ["electrolyzer","haber-bosch", "hydrogen", "pipe"],
    ["electrolyzer","demand_reporter", "oxygen", "pipe"],
    ["demand_reporter", "oxygen_sell", "oxygen", "cable"],
]
classifiers_valid = {
    "electrolyzer": "dispatchable",
    "demand_reporter": "demand",
    "oxygen_sell": "dispatchable",
}
```


All this is to say - thanks for accommodating these edge-cases and making it robust to very complex architectures/connections! Approved!

Issue #819, opened 2026-08-03
Concurrent simulation framework to enable feedback in system-level control


tully9000:
# Concurrent simulation framework to enable feedback in system-level control

Change the H2Integrate framework so that the system-level model can be run in both sequential-type simulations and concurrent-type simulations. A sequential-type simulation is where the inputs and outputs are computed for the whole simulation period (8760 steps) on a subsystem-by subsystem basis:

<img width="594" height="190" alt="Image" src="https://github.com/user-attachments/assets/88606d81-4050-42ea-a470-4db96c61b35e" />

A concurrent-type simulation is where the inputs and outputs are computed for all subsystems on a step-by-step basis:

<img width="594" height="190" alt="Image" src="https://github.com/user-attachments/assets/b0ba5dd7-cc9b-4b87-9146-478cbfa21c53" />

Currently, only the sequential-type simulation is possible in H2Integrate, which makes it impractical to include feedback in system-level control decisions. 

## Proposed solution

This is a running code mock up of a simulation paradigm that could be used to run both sequential-type simulations and concurrent-type simulations in H2Integrate. This example relies on the definition of `N_sim`, the total number of steps in the simulation and `N_step`, the number of steps simulated in each computation step, both of which are made available throughout the model. 

To run a sequential-type simulation set `N_sim = 8760` and `N_step = 8760` and call the openmdao `prob.run_model()`. Openmdao will run the model, computing the 8760-length inputs and outputs on a subsystem-by-subsystem basis as it traverses the system graph.

To run a concurrent-type simulation set `N_sim = 8760` and `N_step = 1` and call `prob.run_model()` in a loop while updating the time step index (see example). Openmdao will run the model, computing the 1-length inputs and outputs for each subsystem. When called in a loop, all subsystem models inputs and outputs will be computed concurrently for each step of the simulation.

```python
import numpy as np
import matplotlib.pyplot as plt

import openmdao.api as om


class ModelBaseClass(om.ExplicitComponent):

    def initialize(self):

        # Number of steps in overall simulation, often 8760.
        self.options.declare(
            "N_sim", default=1, types=int, desc="number of steps in simulation"
        )

        # Number of steps to simulate for each compute call. This will be 8760 if running an annual
        # simulation and 1 if running a feedback steppable simulation.
        self.options.declare(
            "N_step", default=1, types=int, desc="number of steps per compute call"
        )

    def setup(self):

        # Add time step index as an input to all models
[...]

PR #857, opened 2026-08-27
Adding concurrent simulation example

tully9000:
# Concurrent simulation example

This pull request adds the capability to run a concurrent simulation where all subsystems are simulated concurrently, time step by time step in a loop rather than sequentially as individual technologies. This capability relies on two details:

1. All performance models are given `timestep_index` as a discrete input and `n_steps_per_compute` as an attribute. When the `compute` method is called, the performance model will only populate the `timestep_index : timestep_index + n_steps_per_compute` slice of the relevant output vector.
2. A custom openmdao nonlinear solver is implemented to manage the concurrent simulation. The custom solver is set as the nonlinear solver for the "plant" openmdao group and it handles the concurrent simulation loop and updating the `timestep_index` input for the subsystem performance models.

If you want to run a normal H2Integrate analysis, omit `n_steps_per_compute` from the plant config and H2Integrate will behave as it did before.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues

Builds on #819.

### Section 5.1: New Files
- `h2integrate/core/concurrent_nl_solver.py`
- `examples/37_concurrent_simulation/run_solar_battery_grid.py` run script for the example with associated config files.

- `examples/test/test_all_examples.py` - Added concurrent simulation example test.
- `h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py` - Enabled calculation of storage setpoints for just a range of the simulation rather than the full 8760-length vector.
- `h2integrate/control/control_strategies/openloop_control_base.py` - Added `n_steps_per_compute` attribute, `timestep_index` input, and `_get_compute_time_range` method.
- `h2integrate/converters/grid/grid.py` - Enabled calculation of grid operation for just a range of the simulation rather than the full 8760-length vector.
- `h2integrate/converters/solar/solar_pysam.py` - Added logic so the pysam solar model is only run once at the start of the simulation then solar generation information is regurgitated throughout the rest of the simulation as needed.
- `h2integrate/core/dict_utils.py` - Added 3 functions for comparing H2Integrate inputs and outputs between similar runs.
- `h2integrate/core/h2integrate_model.py` - Added logic that runs when the H2I model is being setup to determine whether a concurrent simulation is being run and if so, to apply the custom nonlinear solver to the plant group.
- `h2integrate/core/model_baseclasses.py`- Added `n_steps_per_compute` attribute, `timestep_index` input, `skip_compute` discrete_input, and `_get_compute_time_range` method.
- `h2integrate/demand/demand_base.pEy` - Enabled calculation of demand operation for just a range of the simulation rather than the full 8760-length vector.
- `h2integrate/finances/profast_base.py` - Added `skip_compute` discrete_input so that profast models can be skipped during all but the last time step of a concurrent simulation.
- `h2integrate/storage/storage_baseclass.py`- Enabled calculation of storage operation for just a range of the simulation rather than the full 8760-length vector.

This pull request demonstrates the concurrent simulation capability through an example with `PySAMSolarPerformanceModel`, `StoragePerformanceModel`, `GenericDemandComponent`, and `GridPerformanceModel`. Those technology performance models were modified to varying degrees to work in the concurrent simulation paradigm. To use other technology performance models in a concurrent simulation, they will have to be modified in a similar fashion.

- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:


examples/test/test_all_examples.py
```diff
@@ -3202,3 +3209,66 @@ def test_nuclear_reactor_htse_example(subtests, temp_copy_of_example):
...
+)
+def test_concurrent_simulation_example(subtests, temp_copy_of_example):
+    example_folder = temp_copy_of_example
+
+    # Load config files into dict
+    config_root = example_folder
+    config_path = config_root / "solar_battery_grid.yaml"
+
+    # Load top level config
+    config = load_yaml(config_path)
```
elenya-grant, line 3226: I think this could be removed and you could just initialize `config` as an empty dictionary.
```suggestion
    config = {}
```

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -49,6 +49,8 @@ class DemandOpenLoopStorageController(OpenLoopControlBase):
         3600,
     )  # (min, max) time step lengths (in seconds) compatible with this model
 
+    _soc_timeseries = np.zeros(8760)  # state of charge storage array
```
elenya-grant, line 52: I don't think this is needed here since its done in `setup()`?

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -57,7 +59,17 @@ def setup(self):
         )
         super().setup()
 
-        self.n_timesteps = self.options["plant_config"]["plant"]["simulation"]["n_timesteps"]
```
elenya-grant, line 60: nice!

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -57,7 +59,17 @@ def setup(self):
...
+        # Ensure _soc_timeseries is consistent length with simulation
+        self._soc_timeseries = np.zeros(self.n_timesteps)
+
+        # Initialize soc to the value in the config, if given or halfway between
+        # the minimum and maximum limits otherwise.
+        if hasattr(self.config, "init_soc_fraction"):
+            soc_init = self.config.init_soc_fraction
+        else:
+            soc_init = (1 / 2) * (self.config.min_soc_fraction + self.config.max_soc_fraction)
```
elenya-grant, line 70: I think that the config always have `init_soc_fraction` as an input - at least the logic that previously existed in `compute()` makes it seem that way

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -57,7 +59,17 @@ def setup(self):
...
+
+        self._soc_timeseries[0] = soc_init
```
elenya-grant, line 72: I think that `soc_init` is technically the soc at timestep -1, so I think this logic could be removed. I do think (in a separate PR), we need to better document the relationship between SOC and the charge/discharge profiles.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -131,23 +148,29 @@ def compute(self, inputs, outputs):
...
 
         charge_eff = float(self.config.charge_efficiency)
         discharge_eff = float(self.config.discharge_efficiency)
 
         # Initialize time-step state of charge prior to loop so the loop starts with
         # the previous time step's value
-        soc = deepcopy(init_soc_fraction)
+        if simulation_range.start == 0:
+            if hasattr(self.config, "init_soc_fraction"):
+                soc = self.config.init_soc_fraction
```
elenya-grant, line 159: may still be good to use `deepcopy` here. Also similar comment about `init_soc_fraction` always be a configuration attribute and maybe its not necessary to check it here?

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
```diff
@@ -131,23 +148,29 @@ def compute(self, inputs, outputs):
...
 
         demand_profile = inputs[f"{commodity}_set_point"]
 
         # initialize outputs
-        soc_array = np.zeros(self.n_timesteps)
         set_point_array = np.zeros(self.n_timesteps)
+
         combined_output_array = np.zeros(self.n_timesteps)
         # Loop through each time step
-        for t, demand_t in enumerate(demand_profile):
```
elenya-grant, line 150: kinda think that this looping could be moved to a separate method (similar to the storage model baseclass) so that it can be easily looped over.

h2integrate/converters/solar/solar_pysam.py
```diff
@@ -153,6 +153,9 @@ class PYSAMSolarPlantPerformanceModel(SolarPerformanceBaseClass):
         3600,
     )  # (min, max) time step lengths (in seconds) compatible with this model
 
+    # Flag to avoid unnecessary re-calculation of PySAM model
+    _PYSAM_model_has_been_executed = False
```
elenya-grant, line 157: can this attribute be set in the `setup()` method rather than here?

h2integrate/core/dict_utils.py
```diff
@@ -357,3 +357,79 @@ def check_inputs(prob, tech: str, tech_info: dict, tech_config_path: str):
...
+
+    Returns:
+        (number | array-like): Percent differences
+    """
+
+    # Compute percent difference
+    pd = np.nan_to_num((v2 - v1) / (0.5 * (v1 + v2)))
+
+    # Clamp to zero if either value is smaller than 1e-8
+    pd = np.where(np.max(np.abs(np.stack([v1, v2])), axis=0) <= 1e-8, 0, pd)
```
elenya-grant, line 379: I think the 1e-8 should be an optional input so that it can be changed, i.e.,

```python
def percent_diff(v1, v2, rel=1e-8):
```

h2integrate/core/dict_utils.py
```diff
@@ -357,3 +357,79 @@ def check_inputs(prob, tech: str, tech_info: dict, tech_config_path: str):
...
+        ref_dict (dict): H2I inputs or outputs dict
+
+    Returns:
+        tuple[dict, dict]: First dict is only the keys and items that have percent difference larger
+            than 1e-8. Second dict is only the keys and items that have percent difference larger
+            than 1e-8 and a norm of the original value larger than 1e-8.
+    """
+
+    # Return only the dict items that are non-zero
+    abs_pd = {k: v for k, v in pd_dict.items() if np.abs(v) > 1e-8}
```
elenya-grant, line 433: similar comment, I don't think the 1e-8 should be hard-coded. Could be nice to have as an optional input.

h2integrate/core/dict_utils.py
```diff
@@ -357,3 +357,79 @@ def check_inputs(prob, tech: str, tech_info: dict, tech_config_path: str):
...
+
+    """
+
+    d_out = {}
+    for k1, v1 in d1.items():
+        assert k1 in d2.keys(), "Dicts do not have the same set of keys."
+
+        v2 = d2[k1]
+
+        if isinstance(v1["val"], dict | bool):
```
elenya-grant, line 406: maybe also check if its a string?

h2integrate/core/dict_utils.py
```diff
@@ -357,3 +357,79 @@ def check_inputs(prob, tech: str, tech_info: dict, tech_config_path: str):
...
+        d1 (dict): First dict to compare
+        d2 (dict): Second dict to compare
+
+    Returns:
+        (dict): Dict of percent differences between d1 and d2
+
+    """
+
+    d_out = {}
+    for k1, v1 in d1.items():
```
elenya-grant, line 401: I think you could do this check before starting the dictionary comparison. Like

```python
if bool(set(d1) ^ set(d2)): # if this is True, there are elements that are not in both dictionaries
   raise ValueError("dictionaries do not have the same set of keys")
```

Also - this could be updated to only do a comparison of the keys that are shared, rather than throwing an error if theres non-shared keys. Like `shared_keys = set(d1) & set(d2)`

h2integrate/core/dict_utils.py
```diff
@@ -357,3 +357,79 @@ def check_inputs(prob, tech: str, tech_info: dict, tech_config_path: str):
...
+    """
+    Calculate the percent difference between two numbers or arrays. Return 0
+    percent difference if either value is smaller than 1e-8.
+
+    Args:
+        v1 (number | array-like): First value to compare
+        v2 (number | array-like): Second value to compare
+
+    Returns:
+        (number | array-like): Percent differences
```
elenya-grant, line 372: is this outputting percent difference as a percent or a fraction? Shouldn't it be multiplied by 100 to be percent?

h2integrate/storage/storage_baseclass.py
```diff
@@ -51,6 +51,8 @@ class StoragePerformanceBase(PerformanceModelBaseClass):
     )  # (min, max) time step lengths (in seconds) compatible with this model
     _control_classifier = "storage"
 
+    _soc_timeseries = np.zeros(8760)  # state of charge storage array
```
elenya-grant, line 54: similar comments here as I had with the controllers (and the rest of the similar stuff - like `_soc_timeseries` and `soc_init` and `_soc_timeseries[0]`, etc.

h2integrate/storage/storage_baseclass.py
```diff
@@ -388,7 +424,7 @@ def simulate(
                     (0-100).
         """
 
-        n = len(storage_dispatch_commands)
+        n = sim_end_index - sim_start_index
```
elenya-grant, line 427: can you explain why the `sim_end_index` is now an input and why the previous logic no longer works?

h2integrate/storage/storage_baseclass.py
```diff
@@ -334,6 +366,7 @@ def simulate(
         storage_capacity: float,
         commodity_available: list | np.ndarray,
         sim_start_index: int = 0,
+        sim_end_index: int = 8760,
```
elenya-grant, line 369: don't love the 8760 default value but I can't think of an alternative way of doing this.

h2integrate/finances/profast_base.py
elenya-grant, line 1: I wonder if theres a way we can put the concurrent stuff just around the technology models and not the finance models - so that this kind of logic doesn't have to be included in the finance models. Maybe @johnjasa has an idea on that? Like - what if we put the concurrent solver around `plant` and moved the finance models to a separate group thats at the same level as the `plant` group? That would maybe require more framework changes that its worth ...

elenya-grant (commented):
Hi Zack! Wanted to give you some preliminary feedback - but would love to do another review! I think that this looks good and its very cool functionality! My main comments are related to the initial soc stuff in the controllers/storage models (like - I think `init_soc` is the soc at end of timestep -1). Besides that, my overall thoughts are:

1. I think that some of the class attributes you've added should be created in `setup()` rather than at the class-level
2. I think that for this to come in, it may require adding a flag attribute to all models that says whether it can be run in the concurrent way. Similar to the usage and checking of `_time_step_bounds` - but maybe something named `is_steppable`. Then have a check in H2I to make sure that if someone is running something with this solver, that all the models are steppable.

Thanks again! I'm stoked to see what other folks have to say and I think this is a great start at including feedback control in H2I! Let me know if you have any questions or want to discuss anything on a call!

examples/test/test_all_examples.py
tully9000: I'm not sure. The driver, technology, and plant config information is loaded from the file paths in `solar_battery_grid.yaml` on lines 3227-3229. I think I would need to define all of those separately if config was just an empty dict.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
tully9000: In the testing for `StorageAutoSizingModel` I ran into a case where `init_soc_fraction` was not available in the `self.config` object, which is why I had to include this logic.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
tully9000: removed

h2integrate/converters/solar/solar_pysam.py
tully9000: moved

h2integrate/core/dict_utils.py
tully9000: It was outputting fractional difference (between 0 and 1) but now it returns percent difference (between 0 and 100).

h2integrate/core/dict_utils.py
tully9000: Done.

h2integrate/core/dict_utils.py
tully9000: I like the key similarity logic.

I'm on the fence about comparing keys only the common keys. I was mostly using this function to confirm that two cases are giving the same outputs and in that use case, I would want the function to complain when it gets two dicts with inconsistent sets of keys. But I could also see it being a useful comparison tool for two cases where the outputs aren't supposed to match.

h2integrate/core/dict_utils.py
tully9000: done.

h2integrate/core/dict_utils.py
tully9000: done.

examples/test/test_all_examples.py
elenya-grant: I think you are populating the config on line 3227-3229. I think that other tests in test_all_examples do something similar, check out `test_splitter_wind_doc_h2_example` and `test_electrolyzer_demand`.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
elenya-grant: ooooh - I see. I think if the `init_soc_fraction` isn't in the config, maybe `soc_init` should be set to `min_soc_fraction` instead?

h2integrate/core/dict_utils.py
elenya-grant: for sure! That makes sense. 

If you _do_ want to allow it to only compare the shared keys (even if theres some non-shared keys) you could just do a warnings.warn(dictionaries do not have the same set of keys, UserWarning) instead of a ValueError. Up to you though!

h2integrate/storage/storage_baseclass.py
johnjasa: Agreed that this size hardcoding is a big limitation; looks like it's overwritten in `setup()` but because this size is set at the class level I could see risk of shared-state bugs before `setup()` runs, or for anyone who forgets to reset it.

johnjasa (commented):
Many comments! Not connected but a smattering:

- in the solver, why have multiple possible `skip_compute` keys?
- every cost/finance component now needs an `if not discrete_inputs["skip_compute"]:` check or it silently breaks the skip. This would be easy-to-forget and affects every existing and future cost/finance model. I'd want to ideate on different ways to effectively skip these calcs than having the logic needing to be added to every component
- important: both `add_system_level_controller` and `connect_technologies` currently have logic to add a nonlinear solver based on system coupling. This might overwrite the custom concurrent solver previously set in `create_plant_model`. I'd suggest a more robust check-and-set system or another solution to avoid confusion, maybe something where we store the type of solver needed based on the problem definition, then right before the `setup()` call we actually set the nonlinear solver
- agreed with Elenya that we should find a better solution than `skip_compute` here; there's a lot of unnecessary overhead associated with the current implementation. Putting anything that needs computing into a group might make sense, but it could break some of the performance+cost+controls grouping for technologies we already have
- is there a good general way to handle models that always run the full time amount, e.g. the pysam solar? like, is there a name we could use and have checks to generalize categorizing these and to run them before any concurrent groups?
- we'll need some way to define if a component is steppable, i.e. `is_steppable` as I think Elenya suggested
- small: in general, I'd want to avoid requiring discrete inputs and outputs (which this PR does by the inclusion of `skip_compute`) in the performance models
- state handling should be formalized; all I mean is that SOC is currently stored as a mutable component attribute, and the initial SOC represents the state before timestep zero, but some setup logic stores it at index zero. Repeated-run behavior may vary based on the order of methods called or how setup is called
- I think we could use a different and more clear name than "concurrent simulation", but if that's what's commonly used in the control literature, I'll get on board with it. My main issue is that "concurrent" somewhat implies a parallelism which isn't what's going on here; it's more "time-marching" or "chunked simulation", do any of these (or something else) resonate?
- I think we could benefit by looking at this concurrent framework change at the same time as examining the existing rolling horizon controllers. The current example uses the `DemandOpenLoopStorageController`, but I'd be interested in understanding how the (existing) optimized pyomo controller could be reworked in this new setup or understood differently with the potential concurrent simulation capabilities. Have you already given this much thought, or did the existing optimized storage-level controllers not necessarily provide much fodder?

I have more digging and thought about potential solution pathways here, let me keep digging today and I'll come back to this PR.

h2integrate/storage/storage_baseclass.py
tully9000: From what I can tell, this `simulate` method was being called in two ways: 
1. With an 8760-length `storage_dispatch_commands` input, when the storage is being simulated for the whole year. 
2. With a 24-length `storage_dispatch_commands` input, when the storage is being simulated one day at a time with a storage dispatch controller. 
In both cases, the for-loop runs for the whole length of the `storage_dispatch_commands` input. To simulate just a subset of the `storage_dispatch_commands`, the simulation end index is needed.

It may be possible to achieve the same behavior by pre-processing the `storage_dispatch_commands` array input just before the `simulate` method is called but that would require some extra logic and it seemed more appropriate to handle it within the `simulate` method.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
tully9000: done

examples/test/test_all_examples.py
tully9000: removed unneeded yaml loads.

h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
tully9000: Logic removed.

h2integrate/storage/storage_baseclass.py
tully9000: Removed class level attribute declaration, initialized in `setup`.

tully9000:
> Many comments! Not connected but a smattering:
> 
>     * in the solver, why have multiple possible `skip_compute` keys?
> 
>     * every cost/finance component now needs an `if not discrete_inputs["skip_compute"]:` check or it silently breaks the skip. This would be easy-to-forget and affects every existing and future cost/finance model. I'd want to ideate on different ways to effectively skip these calcs than having the logic needing to be added to every component
> 
>     * important: both `add_system_level_controller` and `connect_technologies` currently have logic to add a nonlinear solver based on system coupling. This might overwrite the custom concurrent solver previously set in `create_plant_model`. I'd suggest a more robust check-and-set system or another solution to avoid confusion, maybe something where we store the type of solver needed based on the problem definition, then right before the `setup()` call we actually set the nonlinear solver
> 
>     * agreed with Elenya that we should find a better solution than `skip_compute` here; there's a lot of unnecessary overhead associated with the current implementation. Putting anything that needs computing into a group might make sense, but it could break some of the performance+cost+controls grouping for technologies we already have
> 
>     * is there a good general way to handle models that always run the full time amount, e.g. the pysam solar? like, is there a name we could use and have checks to generalize categorizing these and to run them before any concurrent groups?
> 
>     * we'll need some way to define if a component is steppable, i.e. `is_steppable` as I think Elenya suggested
> 
>     * small: in general, I'd want to avoid requiring discrete inputs and outputs (which this PR does by the inclusion of `skip_compute`) in the performance models
> 
>     * state handling should be formalized; all I mean is that SOC is currently stored as a mutable component attribute, and the initial SOC represents the state before timestep zero, but some setup logic stores it at index zero. Repeated-run behavior may vary based on the order of methods called or how setup is called
> 
>     * I think we could use a different and more clear name than "concurrent simulation", but if that's what's commonly used in the control literature, I'll get on board with it. My main issue is that "concurrent" somewhat implies a parallelism which isn't what's going on here; it's more "time-marching" or "chunked simulation", do any of these (or something else) resonate?
> 
>     * I think we could benefit by looking at this concurrent framework change at the same time as examining the existing rolling horizon controllers. The current example uses the `DemandOpenLoopStorageController`, but I'd be interested in understanding how the (existing) optimized pyomo controller could be reworked in this new setup or understood differently with the potential concurrent simulation capabilities. Have you already given this much thought, or did the existing optimized storage-level controllers not necessarily provide much fodder?
> 
> 
> I have more digging and thought about potential solution pathways here, let me keep digging today and I'll come back to this PR.

- RE setting the plant solver: I moved the relevant logic from `add_system_level_controller` and `connect_technologies` to a new method that checks whether there is a system level controller, loops in the graph, or a steppable simulation, then sets the appropriate plant solver.
- RE skip compute: Your `SkippableComputeMixin` class handles the skip compute logic nicely
- RE `_is_steppable` label: I added a `_is_steppable` label attribute to a number of base classes with the following behavior:
	- In control strategies and performance models:
		- By default, the flag is set to False in `PerformanceModelBaseClass`, `OpenLoopControlBase`, and `SystemLevelControlBase`. When a subclass that inherits one of these base classes has been modified to be capable of steppable simulation, the `_is_steppable` attribute should be set as true for that subclass only. 
		- If a user tries to run a steppable simulation with a model that isn't capable of steppable simulation, an error will be raised when initializing the `H2IntegrateModel`.
	- In cost models and finance models:
		- By default, the flag is set to False in `CostModelBaseClass` and `ProFastBase`. These base classes inherit John's `SkippableComputeMixin` class so they and their subclasses are only run once in the final time period of a steppable simulation.
		- No error is raised if the user tries to run a steppable simulation with cost or finance models that have `_is_steppable = False` because the default behavior is defined by `SkippableComputeMixin`.
		- If the flag is set to true in a subclass, skip-compute logic in `SkippableComputeMixin` is ignored and the subclass' `compute` method is called at every simulation step just like the other steppable models. This option is relevant for certain system level control scenarios that may need to use cost models outputs such as VarOpEx as inputs in a steppable simulation.
- RE SOC state handling: I removed the old logic that stores the initial soc in the soc array and made some other changes to how the initial soc is set.
- RE naming: I like time-marching simulation with steppable components.

Issue #817, opened 2026-07-30
Connecting Site Lat/Lon to other technology models


elenya-grant:
# Connecting Site Lat/Lon to other technology models

Some models that don't need resource data may have some calculations done based on the site location (EIA natural gas feedstock costs, iron transport model). This was somewhat brought up in Issue #48. Since the site location may be changed in a design of experiments and the site names can be anything a user wants, it'd be nice for models that need the site latitude and longitude to be able to have inputs of latitude and longitude that is connected from some site. I think that one way to do this would be in `resource_to_tech_connections`:
```yaml
[site, eia_ng_feedstock, latitude]
[site, eia_ng_feedstock, longitude]
```

## Proposed solution

The only thing to enable this connection is to update the logic following this if statement in connect_technologies():
```python
if len(resource_source_connections) > len(resource_models):
   # below is existing code
   missing_resource = [k for k in resource_source_connections if k not in resource_models]
   # below is to make sure error isn't thrown with the above connection type
   missing_resource = list(set(missing_resource) - set(self.plant_config["sites"]))
```

For the above solution, it would be good to make sure that someone doesn't connect the latitude of site A to tech and longitude of site B to tech. Aka - we want to prevent the below from running successfully:
```yaml
[siteA, eia_ng_feedstock, latitude]
[siteB, eia_ng_feedstock, longitude]
```

There are to potential ways to prevent this situation from happening:
1. (preferred) Throw an error, add a check it in `connect_technologies()` that raises an error if that happens. Raise a user warning if only latitude or longitude is connected (not both)
2. Add an output named `location` to the site component thats a length 2 array formatted as (latitude, longitude). This was, the user can connect the latitude and longitude with one line. This may get weird though with site sweeps because the location output would be different than the latitude/longitude output. Now - I don't think we should do it this way.

## Alternatives considered

In order to enable this for a possible **transport** component that needs the site info of perhaps two locations to calculate distance, one option is to allow for length 3 connections in `resource_to_tech_connections` such as:
```yaml
[siteA, transport_tech, (latitude, source_latitude)]
[siteA, transport_tech, (latitude, source_latitude)]
[siteB, transport_tech, (latitude, destination_latitude)]
[siteB, transport_tech, (latitude, destination_latitude)]
```

which would connect `site_A.latitude` to `transport_tech.source_latitude` and etc.

[...]

PR #865, opened 2026-09-01
Connecting Site Location to Technologies

elenya-grant:
# Connecting Site Location to Technologies

Update to be able to connect a site to different technologies. This is intended to resolve Issue #817. Basically, this enables the ability to connect site lat/lon to other technology models (transport or otherwise). This required updating some of the logic that handled the `resource_to_tech_connections`, which can now have lists as the 3rd element. For example, to connect a site lat/lon to a transport model:

```yaml
sites:
  source_site:
    latitude: 34.22
    longitude: -102.75
  interconnection_site:
    latitude: 34.126
    longitude: -99.144

resource_to_tech_connections:
  # connect the source site to the transport cost model
  - [source_site, electricity_transport, [latitude, source_latitude]]
  - [source_site, electricity_transport, [longitude, source_longitude]]
  # connect the interconnection site to the transport cost model
  - [interconnection_site, electricity_transport, [latitude, dest_latitude]]
  - [interconnection_site, electricity_transport, [longitude, dest_longitude]]
```

Or, if the destination model just has 'latitude' and 'longitude' as the input names, the connection can look like this:
```yaml
resource_to_tech_connections:
  # connect the source site to the transport cost model
  - [source_site, technology, latitude]
  - [source_site, technology, longitude]
```

This new feature is partially tested with the addition of a new model, the `LinearDistanceCostModel`. The integration of this addition in a site DOE is tested with the changes to Example 22.

- [ ] Feature Enhancement
  - [x] Framework
  - [x] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR
- [x] Complete Section 8: New Model Checklist (if applicable)

### TODO:
- [x] Add check.error message for normal length 3 connections to make sure lat is connected to lat and so on
- [x] Add tests that error messages are thrown as expected
- [x] Add unit tests for the new cost model
- [x] Make updates to the EIA feedstock model as needed?
- [x] Update the doc pages stuff as outlined in Section 8

**Structural feedback**:

**Implementation feedback**:

Should the part of the `LinearDistanceCostModel` that calculates distance be split out to a separate performance model? The reason I didn't do it that way is because the performance is pretty basically covered with the generic transport model.

**Other feedback**:

- [x] Documentation
- [x] `CHANGELOG.md`

## Section 4: Related Issues

Intended resolve Issue #817

Issue(s) created due to this PR are #868 

### Section 5.1: New Files
- `h2integrate/transporters/linear_transport_cost.py`: file containing new transport cost model

  - `connect_technologies()`: Expanded logic for `resource_to_tech_connections` to handle new connection types and check for invalid connections. Also updated logic for when skipping a technology-to-finance model connection.
  - `create_technology_models()`: Updated logic for throwing error for missing technology interconnection. Only throws error if multiple technologies are in the tech config and are not transport models
- `h2integrate/transporters/generic_transporter.py`: added control classifier
- `examples/22_site_doe/`: updated plant and tech config to use the `LinearTransportCostModel` instead of cable.
- `h2integrate/core/test/test_framework.py::test_invalid_resource_to_tech_connections`: new test for error messages 
- `examples/test/test_all_examples.py::test_sweeping_solar_sites_doe`: added subtests for integration testing

- [x] **Model Structure**:
  - [x] Add the new model to the appropriate `__init__.py` file to ensure it is properly imported and used in `supported_models.py`
  - [x] If a new commodity_type is added, update `create_financial_model` in `h2integrate_model.py`
  - [x] If applicable add integration tests
- [x] **Example**: If applicable, a working example demonstrating the new model has been created
  - [x] Input file comments
  - [x] Run file comments
  - [x] Example has been tested and runs successfully in `test_all_examples.py`
- [x] **Documentation**:


examples/22_site_doe/plant_config.yaml
```diff
@@ -9,9 +9,18 @@ sites:
         resource_model: GOESAggregatedSolarAPI
         resource_parameters:
           resource_year: 2012
+  interconnection_site:
+    latitude: 34.126
+    longitude: -99.144
 resource_to_tech_connections:
```
kbrunik, line 15: I know this would be a major change but I wonder if this should be renamed to `site_to_tech_connections`, I don't think that needs to happen in this PR but just musing

h2integrate/feedstocks/eia_ng_price.py
```diff
@@ -162,8 +162,51 @@ def setup(self):
         self.config.price = price.price.to_numpy()
         super().setup()
 
+        if self.config.latitude is not None and self.config.longitude is not None:
```
kbrunik: This is so you can include this in a DOE?

kbrunik (commented):
This looks good to me. I don't feel strongly about moving the "performance" of straightline distance calculation into a performance model. It seems more relevant for the costs than any actual performance modeling.

h2integrate/feedstocks/eia_ng_price.py
elenya-grant: its because theres a few different ways to specify how to define the site for this EIA data. You can define it as a state or with lat/lon directly or with lat/lon pulled from the site. So - the lat/lon is only added as an input if thats how its set in the EIA data.

examples/22_site_doe/plant_config.yaml
```diff
@@ -9,9 +9,18 @@ sites:
...
   # connect the solar resource to the solar technology
   - [site.solar_resource, solar, solar_resource_data]
+  # connect the solar resource site to the transport cost model
+  - [site, electricity_transport, [latitude, source_latitude]]
+  - [site, electricity_transport, [longitude, source_longitude]]
```
RHammond2, line 20: This likely a naive question, but is there a reason why the below isn't an option? The repetition to get two separate variables moved from the `site` to `electricity_transport` feels excessive (same for the below set of lines).

```yaml
  - [site, electricity_transport, [[longitude, source_longitude], [latitude, source_latitude]]]
```

h2integrate/core/h2integrate_model.py
```diff
@@ -1056,6 +1045,20 @@ def create_technology_models(self):
                 )
                 self._check_time_step(tech_name, comp)
                 self.plant.add_subsystem(tech_name, comp)
+        n_non_transport_techs = sum(
+            1 for _, v in self.tech_control_classifiers.items() if v != "transport"
```
RHammond2:
```suggestion
            1 for v in self.tech_control_classifiers.values() if v != "transport"
```

h2integrate/core/h2integrate_model.py
```diff
@@ -1812,8 +1817,61 @@ def connect_technologies(self):
...
+                                f"{tech_parameter.replace(other_loc_var, site_parameter)}."
+                            )
+                            raise ValueError(msg)
+
+                        # If latitude is connected, make sure longitude is also connected
+                        other_connection = [
+                            resource_name,
+                            tech_name,
+                            [other_loc_var, tech_parameter.replace(site_parameter, other_loc_var)],
+                        ]
```
RHammond2, line 1849: I think this actually takes care of my other comment because the connection of latitude and longitude is done automatically if only one is defined. Is there a reason why we need to define both in the plant configuration? If not, I might leave it as a small comment that longitude is automatically connected if latitude is provided, and vice versa, so long as that's necessary for the configuration.

h2integrate/feedstocks/eia_ng_price.py
```diff
@@ -162,8 +162,51 @@ def setup(self):
...
+        self.config.price = price.price.to_numpy()
+
+        return self.config.price
+
     def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
-        if not np.isclose(inputs["price"], self.config.price, rtol=1e-6).all():
+        if "latitude" in inputs:
+            price = self.get_price(inputs["latitude"][0], inputs["longitude"][0])
+        else:
+            price = self.config.price
```
RHammond2: The logic here and in the method call itself make me think that we should be updating the configuration itself with the updated site information rather than potentially overriding the state value and not setting it to the value that's actually used.

Basically having an `update_config()` for if there is site info that updates `state`, `latitude`, and `longitude`, then calling `get_price()`.

Does that make sense or is there a reason why wouldn't want to do that I'm glossing over/missing?

RHammond2 (commented):
I primarily focused on the EIA pieces, so my substantive commentary is there with a couple small bits in the configuration. In general the EIA modifications are ok, but I do wonder if we should just be fully updating the configuration if there is new site information to override what the user might have originally passed.

examples/22_site_doe/plant_config.yaml
elenya-grant: I was trying to keep the connection logic the same as `technology_interconnections`. I agree it seems a bit excessive but ... I would argue that it would allow for more flexibility in the future, where perhaps other site information could be passed to technology model thats not just latitude and longitude.

h2integrate/core/h2integrate_model.py
elenya-grant: This is not an auto-connection - it's more of an error-check. It checks that if latitude is connected, that longitude is also connected (but does not make that connection automatically). It also checks that latitude is connected to latitude (don't want someone to connect latitude to longitude). But we could enable an auto-connection but I think I'm leaning towards having the user be more explicit rather than handling these connections under the hood. I'm curious what other folks like @johnjasa and @kbrunik think.

h2integrate/feedstocks/eia_ng_price.py
elenya-grant: I think I get what you're saying. would it look like this:
```python
if "latitude" in inputs:
   self.config.latitude = inputs["latitude"][0]
   self.config.longitude = inputs["longitude"][0]
   self.config.state = get_state_from_coords(latitude = inputs["latitude"][0], longitude=inputs["longitude"][0])
   self.get_price(inputs["latitude"][0], inputs["longitude"][0])
```

examples/22_site_doe/plant_config.yaml
elenya-grant: I agree - could be nice to change in the future!

h2integrate/core/h2integrate_model.py
elenya-grant: done!

h2integrate/feedstocks/eia_ng_price.py
elenya-grant: I just undid the changes to the EIA model since I think there may be a bigger discussion there for the future! I will comment the issue number once its made.

h2integrate/core/h2integrate_model.py
kbrunik: I feel fine about leaving it as something the user has to explicitly define. I'm not against having some of this auto-connection but don't really see it as necessary either

kbrunik (approved):
Looks good to me! thanks for making issue #868

h2integrate/core/h2integrate_model.py
```diff
@@ -1812,8 +1817,61 @@ def connect_technologies(self):
...
+                        # If latitude is connected, make sure longitude is also connected
+                        other_connection = [resource_name, tech_name, other_loc_var]
+                        if other_connection not in resource_to_tech_connections:
+                            msg = (
+                                f"{variable} is connected between {resource_name} and "
+                                f"{tech_name}, but {other_loc_var} is not. Please ensure that "
+                                f"both latitude and longitude are connected from "
+                                f"'{resource_name}' to technology '{tech_name}'"
+                            )
+                            raise ValueError(msg)
```
elenya-grant: @johnjasa - theres a bit of almost repeated logic here around the 3rd element of the connection being a string vs a list - curious if you have any thoughts on if theres a slick way to reduce the nearly-identical "checking" for these two cases while maintaining the ability to understand the individual checks that are happening

h2integrate/core/h2integrate_model.py
johnjasa: Agreed with your view, @elenya-grant and @kbrunik! I think we could generalize potential auto-connections in a helpful way later down the line, and probably for more things than just lat/lon pairs. For now, let's stay explicit, especially because this is the first time in a while that the site stuff has been touched and I expect more future work here with the metals and other projects; it'll pay to have behavior be simple, explicit, and clear here for now.

h2integrate/core/h2integrate_model.py
johnjasa: I pushed up a simplification directly!

PR #846, opened 2026-08-14
Improve error message checking throughout; bugfix in `check_inputs`

elenya-grant:
# Improve error message checking throughout; bugfix in `check_inputs`

This was inspired by Issue #845. While resolving the error message, I wanted to make sure that all error messages were being properly tested in `h2integrate/core/test/test_utilities.py::test_check_inputs`. I temporarily modified the error messages being thrown in `check_inputs` to have numbers at the start (to see if they were being tested) and the tests still passed (which they shouldn't have - because I changed the error string). So - I realized that the assert statement paired with the `pytest.raises` command were indented one too many times and that the tests were not actually being ran. Anyway - I made it so that the tests do run (they all failed initially because of small differences in the error message strings). 

I'm going to delete the test `test_check_inputs_example01` that I added - since now `test_check_inputs` is actually working. I also fixed the error message described in Issue #845 so that it works properly.

John note: I've also updated a few other files where we had similar issues with the assert nesting structure to remove those problems.

## Section 1: Type of Contribution

- [ ] Feature Enhancement
  - [ ] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [ ] Tools/Utilities
  - [ ] Other (please describe):
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes
- [ ] Other (please describe):

## Section 2: Draft PR Checklist

- [x] Open draft PR
- [x] Describe the feature that will be added
- [x] Fill out TODO list steps
- [x] Describe requested feedback from reviewers on draft PR
- [ ] Complete Section 8: New Model Checklist (if applicable)

### TODO:
- [x] Remove `test_check_inputs_example01` and `h2integrate/core/test/inputs/buggy_battery_v0.yaml`
- [x] Update changelog

### Type of Reviewer Feedback Requested (on Draft PR)

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

Should the `msg` in `check_inputs` be added onto and have an error thrown at the end of the method or just have tests be raised immediately when the situation arises? (see my comment on Issue #845 for more info)

## Section 3: General PR Checklist

- [x] PR description thoroughly describes the new feature, bug fix, etc.
- [x] Added tests for new functionality or bug fixes
- [x] Tests pass (If not, and this is expected, please elaborate in the Section 6: Test Results)
- [x] Documentation
  - [x] Docstrings are up-to-date
  - [x] Related `docs/` files are up-to-date, or added when necessary
  - [x] Documentation has been rebuilt successfully
[...]

PR #882, opened 2026-09-14
Check for extraneous and missing `tech_to_dispatch_connections`

### LINKED ISSUE #848: Check for extraneous `tech_to_dispatch_connections` in plant config if not using a pyomo storage controller
# Check for extraneous `tech_to_dispatch_connections` in plant config if not using a pyomo storage controller

If someone is swapping out the storage controller from a Pyomo controller to an openloop controller in their tech config, they have to do a few things:
- remove `dispatch_rule_set` from the storage technology and the corresponding inputs/outputs
- remove `dispatch_rule_set` from the technology connected to the storage model (and - if that technology is a combiner, then rename `shared_parameters` to `performance_parameters`)

Once these two things are done, if they don't remove `tech_to_dispatch_connections` in the plant config - then they get a hard-to-debug error. This check should NOT be done in the storage performance models or controllers, but instead should be done in H2I when the configuration classes are either loaded (or after they're checked)

Check for extraneous `tech_to_dispatch_connections`:
Check if `tech_to_dispatch_connections` is in the plant config. If so - then look at the technologies in those connections in the tech config and see if those techs include a `dispatch_rule_set`. If they don't - then throw an error. 

Check for missing `tech_to_dispatch_connections` (if needed):
Check if any technologies in the tech config have a `dispatch_rule_set` then check that those technologies are included in the `tech_to_dispatch_connections`. If `tech_to_dispatch_connections` is missing or incorrect - throw an error that tells the user what `tech_to_dispatch_connections` should look like (we can figure this out based on the technology connections).

johnjasa:
# Check for extraneous and missing `tech_to_dispatch_connections`

This PR adds configuration validation to check for extraneous or missing `tech_to_dispatch_connections` between the plant and technology configurations:
1. When `tech_to_dispatch_connections` is specified in `plant_config`, verifies that connected technologies include `dispatch_rule_set` or utilize a Pyomo storage controller, raising a helpful error if open-loop controllers are used without dispatch rules.
2. When technologies define `dispatch_rule_set`, verifies that `tech_to_dispatch_connections` is present and correctly structured, inferring expected connections from `technology_interconnections` when reporting errors.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [x] Tools/Utilities
- [x] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [x] Add dispatch connection checker utility function
- [x] Integrate checker into H2IntegrateModel configuration loading
- [x] Add comprehensive unit tests for validation cases

**Structural feedback**:

**Implementation feedback**:

**Other feedback**:

- [x] Documentation
  - [ ] Related docs/ files are up-to-date, or added when necessary
- [x] CHANGELOG.md

## Section 4: Related Issues
- Closes #848
- Made #883
### Section 5.1: New Files
- N/A

  - Added `_check_dispatch_connections`, invoked in `__init__` right after `create_technology_models()` (so it runs before any dispatch-related OpenMDAO connections are made).
  - Fixed a latent bug in `connect_technologies()` where `dispatch_rule_set` lookups used `self.technology_config.get(tech_name, ...)` instead of `self.technology_config["technologies"].get(tech_name, ...)`, which meant dispatch rule connections were never actually wired.
  - Added lightweight stub-based unit tests for `_check_dispatch_connections` covering valid heuristic-style and standalone-Pyomo-controller configs, a no-op case, an extraneous-connection error, and two missing-connection error cases.
- `examples/09_co2/direct_ocean_capture/tech_config.yaml`
  - Removed the unused `dispatch_rule_set`/`dispatch_rule_parameters` on the `wave` technology (it fed `combiner`, not the storage tech directly, so it was never wired).
- `examples/11_hybrid_energy_plant/tech_config.yaml`
  - Removed the unused `dispatch_rule_set`/`dispatch_rule_parameters` on the `wind` and `solar` technologies (same reason as above).

N/A

New and existing tests pass

N/A

### DIFF
--- a/h2integrate/core/h2integrate_model.py
+++ b/h2integrate/core/h2integrate_model.py
@@ -75,6 +75,10 @@ class H2IntegrateModel:
         # they will need tech_config but not driver or plant config
         self.create_technology_models()
 
+        # validate `tech_to_dispatch_connections` against `dispatch_rule_set`/
+        # `control_strategy` declarations before building any further OpenMDAO connections
+        self._check_dispatch_connections()
+
         self.create_finance_model()
 
         # add system-level controller if configured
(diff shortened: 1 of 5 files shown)


h2integrate/core/h2integrate_model.py
```diff
@@ -1111,6 +1115,135 @@ def _check_control_classifier(self, model_name, model_object):
...
+        """
+        from h2integrate.control.control_strategies.pyomo_storage_controller_baseclass import (
+            PyomoStorageControllerBaseClass,
+        )
+
+        technologies = self.technology_config.get("technologies", {})
+        dispatch_connections = self.plant_config.get("tech_to_dispatch_connections") or []
+
+        def _has_pyomo_storage_controller(tech_name):
+            control_model_name = (
```
elenya-grant, line 1164: some inline comments on `_has_pyomo_storage_controller` and `_is_dispatch_controlled` would be nice. Like "returns True if X and false if X" - cause the `and` and `or` logic makes it a bit harder to follow

h2integrate/core/h2integrate_model.py
```diff
@@ -1111,6 +1115,135 @@ def _check_control_classifier(self, model_name, model_object):
...
+                tech_name, {}
+            ) or _has_pyomo_storage_controller(tech_name)
+
+        # --- Check 1: extraneous connections -------------------------------
+        if dispatch_connections:
+            invalid_dispatching_techs = sorted(
+                {
+                    connection[1]
+                    for connection in dispatch_connections
+                    if len(connection) == 2 and not _is_dispatch_controlled(connection[1])
```
elenya-grant: when would `len(connection)` ever not equal 2? I think its fine to leave as-is or else this error-check would end-up throwing a hard-to-figure-out error if connection[1] was used but connection was length 1

h2integrate/core/h2integrate_model.py
```diff
@@ -1111,6 +1115,135 @@ def _check_control_classifier(self, model_name, model_object):
...
+                control_cls, PyomoStorageControllerBaseClass
+            )
+
+        def _is_dispatch_controlled(tech_name):
+            return "dispatch_rule_set" in technologies.get(
+                tech_name, {}
+            ) or _has_pyomo_storage_controller(tech_name)
+
+        # --- Check 1: extraneous connections -------------------------------
+        if dispatch_connections:
```
elenya-grant, line 1183: what if we first start by checking that all connections is dispatch_connections are length 2 then throw an error if any are not length 2 before we get into this method? That way, the two-length connection checks in the rest of this method can be removed.

h2integrate/core/test/test_framework.py
```diff
@@ -1376,3 +1376,165 @@ def test_create_xdsm_propagates_file_not_found_error():
...
+    fake = _make_dispatch_fake_model(
+        technologies,
+        [["combiner", "battery"], ["battery", "battery"]],
+        interconnections,
+    )
+    with pytest.raises(ValueError) as excinfo:
+        H2IntegrateModel._check_dispatch_connections(fake)
+    err = str(excinfo.value)
+    assert "battery" in err
+    assert "tech_to_dispatch_connections" in err
```
elenya-grant: what are your thoughts on these tests having more detailed sub-sentences in the error-message asserts? I think my rule-of-thumb is starting to be that any part of an error message that has curly-braces should be tested with some pre-curly brace and some post-curly brace words for context. I can't tell which of the two errors are being thrown on first-glance at these tests.

I think it'd be nice to test (for the first error), something more like below:

```python
assert "plant config references ['battery'] but this technology" in err
assert "corresponding entries for ['battery'] from `tech_to_dispatch_connections`"
```

And for the second error, something like (for example):

```python
assert "Technology ['battery'] declare a `dispatch_rule_set` but" in err
assert "(at least): [['combiner','battery'],['battery','battery']]." in err
```

I think that PR #846 made me think that it's good to really check that the information that the error message should raise is communicated well. Some of the error messages that were fixed in PR #846 were not properly tested and the error messages were actually confusing and not communicating the problem well.

elenya-grant (commented):
thanks for adding this check in and making issue #883! I left some small comments, which are probably because I'm still waking up and wanted some more comments to tell me what was happening in the checks! My main comment is about error message checking in the tests and adding more detail to those if possible :)

h2integrate/core/h2integrate_model.py
johnjasa: Good call, added

h2integrate/core/h2integrate_model.py
johnjasa: Bad connection lengths are handled later in `connect_technologies()` with the clearer `ValueError("Invalid tech to dispatching_tech_name connection: ...")` so I'll leave it as-is.

h2integrate/core/h2integrate_model.py
johnjasa: Good call, added this at the top of the method:

```
        invalid_connections = [
            connection for connection in dispatch_connections if len(connection) != 2
        ]
        if invalid_connections:
            raise ValueError(
                "Invalid tech to dispatching_tech_name connection(s): "
                f"{invalid_connections}. Each connection must contain exactly two technology names."
            )
```

h2integrate/core/test/test_framework.py
johnjasa: Thanks for the very clear comment, with reasoning, suggestions, and a call back to other PRs! Really helpful context here to help me understand what changes you want.

I updated all three tests to be more verbose in their checking, here's one example:

```
assert "plant config references ['battery'], but this technology does not" in err
assert (
    "corresponding entries for ['battery'] from `tech_to_dispatch_connections`" in err
)
```

elenya-grant (approved):
Thanks for making those changes so quickly! This is good to come in from my perspective! Thanks for updating the tests and adding in this extra check! :)

Issue #651, opened 2026-04-03
Create up sampling and down sampling capabilities


jaredthomas68:
# Create upsampling and downsampling capabilities
Some models will not be capable of shifting the time step duration and we may need to up sample or down sample in order to accommodate variable dt.

Issue #250, opened 2025-09-17
Generalize dispatch timesteps (allow different than hourly)


johnjasa:
would the units for `pyomo_model.time_duration` being hard-code to `hr` create problems for sub-hourly time steps?

_Originally posted by @elenya-grant in https://github.com/NREL/H2Integrate/pull/211#discussion_r2334899411_
            

Dispatch currently assumes hourly timesteps; to make H2I more flexible for different timesteps, we can change dispatch to handle other timestep durations.

PR #887, opened 2026-09-17
Allow for non-annual simulation.

jaredthomas68:
# Allow for non-annual simulation

This PR adds capabilities to run H2I for non-annual simulations. The approach taken allows for data and simulation on nearly any time step and number of steps. Annual simulation data is then reduced to lists of capacity factors, varopex, and a replacement schedule, using some extrapolation if needed to reach plant life. The resulting lists are then used in the ProFAST cash flow analysis to estimate life-time financial metrics. 

The following is an incomplete list of changes

 1. General
     - n_timesteps*dt no longer has to equal 1 year
     - Example 24 has been updated to demonstrate non-annual and non-hourly simulation within a single run script.
 3. Performance model changes
     - Output capacity factor and replacement schedule are determined based on all simulation time completed
     - Extrapolation to years not simulated can be done using tiling of simulated years, repeating the final year, or using the mean CF.
     - Similar extrapolation approaches can be applied to state-of-health (SOH)/degradation when supplied and will help determine the replacement schedule
     - Adjusted to handle non-hourly simulation more generally
 5. Cost models
     - Adjusted to handle non-hourly simulation and use multiple simulation years for varopex
 6. Finance models
     - ProFAST wrapper has been adjusted to not assume hourly timesteps or annual simulation. 
 7. Resource models/api calls
     - Data for multiple years will be downloaded automatically to accommodate the simulation length requested
     - Data for multiple years can be supplied by providing a list of resource filenames where each file is assumed to contain a single year of data and the files are assumed to be given in chronological order
     - Leap can be included or excluded from the data if available in the downloaded or provided data
     - Data may be upsampled or downsampled using pandas resampling methods to meet the requested timestep. The sampling method can be changed by the user.

- [x] Feature Enhancement
  - [x] Framework
  - [ ] New Model
  - [ ] Updated Model
  - [x] Tools/Utilities
- [ ] Bug Fix
- [ ] Documentation Update
- [ ] CI Changes

- [x] Open draft PR

### TODO:
- [ ] Make sure user is aware if resampling is done
- [ ] Add documentation

**Structural feedback**:
- Suggestions regarding use of time_tools.py functions vs class methods
- Code reuse or other ideas for reducing the number of changes

**Implementation feedback**:
- I have used a couple `try` statements and while loops. I'd love any ideas you have for avoiding those as they can be error prone and hard to read.

**Other feedback**:
- Suggestions on how/if to split this PR into smaller PRs
- Anything else you want to comment on

- [ ] Documentation
- [ ] `CHANGELOG.md`

## Section 4: Related Issues

This PR relates to several issues, I may have missed some:
- #651 - upsampling and downsampling
- #204 - variable timestep number and length
- #250 - allow non-hourly dispatch timesteps 
    - This PR only addresses this for `DemandOpenLoopStorageController`
- #638 - non-hourly grid model
 
### Section 5.1: New Files
- test_model_baseclasses.py
  - `calculate_annual_cf_and_replacement_schedule`: Adds unit tests for annualization and replacement-schedule projection.
- test_resource_base.py
  - `_acquire_resource_data`: Adds tests for multi-year resource acquisition and ordered filename-list handling.
- test_time_tools.py
  - `process_leap_day`, `conform_resource_data_to_n_timesteps`, `concatenate_resource_years`, `resample_resource_data_to_dt`: Adds tests for leap-day handling, horizon conformance, multi-year concatenation, and timestep resampling.
- 30.6617_-101.7096_psmv3_60_2014.csv
  - Resource data: Adds a second solar weather year so the example can run a multi-year case without API calls.

- examples/24_solar_battery_grid/plant_config.yaml
  - Finance/resource config: Points the example at multi-year GOES resource inputs for non-annual runs.
- run_solar_battery_grid.py
  - Example driver: Runs half-year, annual, and two-year variants programmatically with different time steps and prints a comparison table.
- demand_openloop_storage_controller.py
  - Controller timestep math: Reworks charge/discharge and SOC updates to use `dt` consistently for non-hourly simulations.
- grid.py
  - `compute`: Annualizes grid production and variable costs by simulated year so sub-annual and multi-year cases scale correctly.
- atb_utility_pv_cost.py
  - `_time_step_bounds`: Widens accepted timesteps so the solar cost model can participate in non-hourly runs.
- solar_pysam.py
  - `compute`: Uses helper-based annualization for non-annual runs while retaining PySAM annual outputs for exact annual cases.
- test_pysam_solar.py
  - Solar tests: Adds non-annual regression coverage and updates grouped assertions for the new behavior.
- validation.py
  - `load_plant_yaml`: Allows positive non-annual simulation horizons up to plant life instead of assuming a single year.
- model_baseclasses.py
  - `calculate_annual_cf_and_replacement_schedule`, `calculate_annual_varopex`: Centralizes annual projection logic for arbitrary simulation lengths and fixes variable OpEx annualization across timesteps.
- test_framework.py
  - Framework tests: Updates validation coverage for finance subgroup commodity-stream timeseries requirements.
- demand_base.py
  - Demand annualization: Uses helper-based annual production and capacity-factor logic for arbitrary simulation lengths.
- profast_base.py
  - `populate_profast`: Reconstructs ProFAST capacity and annual production from commodity amount/rate units instead of timestep-specific assumptions.
- resource_base.py
  - Resource acquisition: Supports slicing sub-annual data, loading multiple years, and consuming lists of per-year resource files.
- resource_base_hpc.py
  - Leap-day handling: Aligns the HPC resource acquisition with the updated leap-day processing interface.
- openmeteo_solar.py
  - Resource preprocessing: Adopts the updated leap-day utility without fixed annual-length assumptions.
- time_tools.py
  - Time utilities: Adds leap-day-aware slicing, multi-year concatenation, and pandas-based resampling for arbitrary timesteps.
- nlr_developer_wtk_api_models.py
  - Resource loading: Adapts wind API resource handling to the updated time-tool and multi-year acquisition flow.
- openmeteo_wind.py
  - Resource preprocessing: Adopts the updated leap-day utility without fixed annual-length assumptions.
- atb_battery_cost.py
  - `_time_step_bounds`: Removes hourly-only restrictions because the battery cost model is timestep-independent.
- storage_baseclass.py
  - Storage annualization: Separates discharge-based annual production from net-flow capacity factor for non-annual runs.
- storage_performance_model.py
  - `_time_step_bounds`: Widens accepted timesteps so storage performance can run outside hourly resolution.
 
- [ ] **Model Structure**:
  - [ ] Input file comments
  - [ ] Run file comments
- [ ] **Documentation**:

### DIFF
--- a/h2integrate/core/inputs/validation.py
+++ b/h2integrate/core/inputs/validation.py
@@ -152,15 +152,33 @@ def load_tech_yaml(finput):
 def load_plant_yaml(finput):
     plant_config = _validate(finput, fschema_plant)
 
-    n_timesteps = plant_config["plant"]["simulation"]["n_timesteps"]
-    dt = plant_config["plant"]["simulation"]["dt"]
+    n_timesteps = int(plant_config["plant"]["simulation"]["n_timesteps"])
+    dt = int(plant_config["plant"]["simulation"]["dt"])
+    plant_life = int(plant_config["plant"]["plant_life"])
 
-    if int(n_timesteps) * int(dt) != 31536000:  # seconds in simulation must be seconds/year
+    seconds_per_year = 31_536_000  # 8760 h/year * 3600 s/h
+    seconds_simulated = n_timesteps * dt
+    years_simulated = seconds_simulated / seconds_per_year
+
+    # The simulation may cover any positive duration (a fraction of a year, a single
+    # year, or multiple years). Performance/cost/finance models annualize their results
+    # using ``fraction_of_year_simulated`` (see the model base classes), so an arbitrary
+    # horizon is supported as long as it is positive and does not exceed the plant life.
+    if seconds_simulated <= 0:
         msg = (
-            "H2Integrate does not currently support simulations that are less than or "
-            "greater than 1-year. Please ensure that "
+            "The simulation horizon must be positive. Please ensure that "
             "plant_config['plant']['simulation']['n_timesteps'] times "
-            "plant_config['plant']['simulation']['dt'] equals 31536000 (s)."
+            "plant_config['plant']['simulation']['dt'] is greater than 0 (s)."
+        )
+        raise ValueError(msg)
+
+    if years_simulated > plant_life:
+        msg = (
+            "H2Integrate does not support simulations that are longer than the plant "
+            f"life. The configured simulation covers {years_simulated:.4g} years "
+            f"(n_timesteps={n_timesteps} * dt={dt} s = {seconds_simulated} s), but "
+            f"plant_config['plant']['plant_life'] is {plant_life} years. Either shorten "
+            "the simulation horizon or increase plant_life."
         )
         raise ValueError(msg)
 
--- a/h2integrate/core/test/test_framework.py
+++ b/h2integrate/core/test/test_framework.py
@@ -798,20 +798,32 @@ def test_unsupported_simulation_parameters(temp_dir):
     plant_config_data_dt = load_plant_yaml(temp_plant_config_dt)
     # docs fencepost end: DO NOT REMOVE
 
-    # Modify the n_timesteps entry for the temp_plant_config_ntimesteps
-    plant_config_data_ntimesteps["plant"]["simulation"]["n_timesteps"] = 8759
-    # Modify the dt entry for the temp_plant_config_dt
-    plant_config_data_dt["plant"]["simulation"]["dt"] = 3601
+    plant_life = int(plant_config_data_ntimesteps["plant"]["plant_life"])
 
-    # Save the modified plant_configs YAML back
+    # Sub-year and multi-year horizons are now supported as long as they are positive
+    # and do not exceed the plant life. A ~half-year horizon should load without error.
+    plant_config_data_ntimesteps["plant"]["simulation"]["n_timesteps"] = 4380  # 0.5 year at 1 h
     with temp_plant_config_ntimesteps.open("w") as f:
         yaml.safe_dump(plant_config_data_ntimesteps, f)
+    load_plant_yaml(temp_plant_config_ntimesteps)
+
+    # A multi-year horizon (2 years of hourly data) should also load without error.
+    plant_config_data_dt["plant"]["simulation"]["n_timesteps"] = 2 * 8760
     with temp_plant_config_dt.open("w") as f:
         yaml.safe_dump(plant_config_data_dt, f)
+    load_plant_yaml(temp_plant_config_dt)
 
-    # check that error is thrown when loading config with invalid number of timesteps
-    with pytest.raises(ValueError, match="greater than 1-year"):
-        load_plant_yaml(plant_config_data_ntimesteps)
+    # A horizon longer than the plant life is not supported and must raise.
+    over_life = deepcopy(plant_config_data_dt)
+    over_life["plant"]["simulation"]["n_timesteps"] = (plant_life + 1) * 8760
+    with pytest.raises(ValueError, match="longer than the plant"):
+        load_plant_yaml(over_life)
+
+    # A non-positive horizon is invalid and must raise.
+    non_positive = deepcopy(plant_config_data_dt)
+    non_positive["plant"]["simulation"]["dt"] = 0
+    with pytest.raises(ValueError, match="must be positive"):
+        load_plant_yaml(non_positive)
 
 
 @pytest.mark.unit
--- a/h2integrate/storage/battery/atb_battery_cost.py
+++ b/h2integrate/storage/battery/atb_battery_cost.py
@@ -1,4 +1,5 @@
 from attrs import field, define, validators
+from numpy import inf
 from openmdao.utils import units
 
 from h2integrate.core.utilities import merge_shared_inputs
@@ -61,9 +62,11 @@ class ATBBatteryCostModel(CostModelBaseClass):
     """
 
     _time_step_bounds = (
-        3600,
-        3600,
-    )  # (min, max) time step lengths (in seconds) compatible with this model
+        1e-6,
+        inf,
+    )  # (min, max) time step lengths (in seconds) compatible with this model. The ATB
+    # cost model is time-step independent (it only uses storage capacity and charge
+    # rate), so it accepts any sub-hourly-to-hourly time step.
 
     def setup(self):
         self.config = ATBBatteryCostConfig.from_dict(
@@ -104,8 +107,7 @@ class ATBBatteryCostModel(CostModelBaseClass):
             storage_duration_hrs = max_capacity_kWh / max_charge_rate_kW
         if max_charge_rate_kW < 0:
             msg = (
-                f"max_charge_rate cannot be less than zero and has value of "
-                f"{max_charge_rate_kW} kW"
+                f"max_charge_rate cannot be less than zero and has value of {max_charge_rate_kW} kW"
             )
             raise UserWarning(msg)
         # CapEx equation from Cell E29


h2integrate/control/control_strategies/storage/demand_openloop_storage_controller.py
elenya-grant, line 1: this change does not seem relevant to non-annual simulations but instead seems more relevant to enabling different timesteps and should perhaps be part of a different PR (given the size of this one)

h2integrate/converters/solar/solar_pysam.py
```diff
@@ -173,8 +173,8 @@ class PYSAMSolarPlantPerformanceModel(SolarPerformanceBaseClass):
     """
 
     _time_step_bounds = (
-        3600,
-        3600,
+        900,
+        14400,
```
elenya-grant, line 177: are you sure this is a valid upper-bound? I thought PvWatts could only for 5min, 15 min, 30 min or hourly?

h2integrate/core/model_baseclasses.py
```diff
@@ -140,6 +141,257 @@ def apply_curtailment(self, outputs):
             command_value = self._inputs[command_value_key]
             outputs[commodity_out_key] = np.minimum(uncurtailed, command_value)
 
+    # TODO rewrite and reorganize this method
+    def calculate_annual_cf_and_replacement_schedule(
```
elenya-grant, line 147: I agree with this TODO statement. This method is really large, I think it should last least be split up so that the state of health stuff is separate (not many models have degradation included - this is a new concept).

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+                year_data = self._load_single_year_data(
+                    latitude, longitude, site_changed, year_filename
+                )
+                yearly_data.append(year_data)
+                total_timesteps += self._resource_length(year_data)
+                offset += 1
+        finally:
+            # Restore the configured resource year, which is advanced above for multi-year loads
+            if isinstance(base_year, int):
+                self.config.resource_year = base_year
```
elenya-grant: I don't think we should be changing the config value for resource_year - as noted in Issue #870

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+        filename_list = (
+            list(resource_filename) if isinstance(resource_filename, list | tuple) else None
+        )
+        base_year = self.config.resource_year
+
+        yearly_data = []
+        total_timesteps = 0
+        offset = 0
+        try:
+            while total_timesteps < self.n_timesteps:
```
elenya-grant: I think the `while` could be replaced with instead calculating the number of resource years needed to get enough resource data to achieve n_timesteps.

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+                            f"{self.n_timesteps} timesteps required by the simulation horizon. "
+                            "Provide additional resource files or shorten the horizon."
+                        )
+                        raise ValueError(msg)
+                    year_filename = filename_list[offset]
+                elif offset == 0:
+                    year_filename = resource_filename or None
+                else:
+                    # A single provided filename cannot supply additional years
+                    if resource_filename:
```
elenya-grant: I think this type of logic should be discussed with the rest of the group - I think that theres a few different routes we could go with resource_filenames that I mentioned in Issue #888

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+        )
+        base_year = self.config.resource_year
+
+        yearly_data = []
+        total_timesteps = 0
+        offset = 0
+        try:
+            while total_timesteps < self.n_timesteps:
+                # Resolve the filename to use for this year, if any
+                if filename_list is not None:
```
elenya-grant: I think a lot of these checks within the `while` statement could be moved outside of it.

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+
+                # Advance the resource year for years after the first
+                if offset > 0:
+                    if not isinstance(base_year, int):
+                        msg = f"Resource year must be an integer, {base_year} was given."
+                        raise ValueError(msg)
+                    try:
+                        # Setting resource_year runs the config validator, which raises
+                        # if the year is outside the range supported by this dataset.
+                        self.config.resource_year = base_year + offset
```
elenya-grant: this logic will not work for the three TMY solar resource models. I also don't think we should be changing the config value here. I think that resource_year needs to made as an input to the corresponding methods.

h2integrate/resource/resource_base.py
```diff
@@ -298,14 +354,114 @@ def get_data(self, latitude, longitude, first_call=True):
...
+
+    @staticmethod
+    def _resource_length(data):
+        """Return the number of timesteps in a resource data dictionary."""
+        for key in ("year", "month", "day", "hour", "minute"):
+            if key in data:
+                return len(np.asarray(data[key]))
+        # Fall back to the first array-like value
+        for value in data.values():
+            if isinstance(value, np.ndarray | list | tuple) and not isinstance(value, str | bytes):
```
elenya-grant, line 482: I think that the check could be as simple as making sure value isn't a string, float, int, dictionary or boolean.

h2integrate/converters/solar/solar_pysam.py
```diff
@@ -476,12 +476,32 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
...
-
-        outputs["capacity_factor"] = outputs["total_electricity_produced"] / max_production
-        outputs["annual_electricity_produced"] = self.system_model.value("ac_annual")
+        # For a full-year (annual) simulation, use PySAM's ac_annual and the simple scalar
+        # capacity factor. Pvwattsv8 does not assign ac_annual for non-annual horizons
+        # (sub-annual or multi-year), so use the base-class projection to compute per-year
+        # capacity factors and annual production across the plant life.
+        seconds_per_year = 31_536_000  # 8760 h/year * 3600 s/h
+        seconds_simulated = self.n_timesteps * self.dt
+        is_annual = abs(seconds_simulated - seconds_per_year) < self.dt / 2
```
elenya-grant: I thin that we could just check `self.fraction_of_year_simulated` instead of `is_annual`

h2integrate/converters/solar/solar_pysam.py
```diff
@@ -476,12 +476,32 @@ def compute(self, inputs, outputs, discrete_inputs, discrete_outputs):
...
+
+        rated_production = outputs["rated_electricity_production"][0]
+        if is_annual:
```
elenya-grant: I'm nervous about this logic if using lifetime performance results in PySAM

elenya-grant (changes requested):
Hi Jared! Thanks for this PR! I think that this is a neat functionality add. Thanks for putting all this work! 

Responding to your requested feedback from reviewers, I think that this should be split into separate PRs so that each change can get the attention and feedback it deserves. I left some small comments for now but would want to do an in-depth review if this is going to come-in as-is. Since this touches so many baseclasses, I think that splitting out PRs based on base-class changes would be a good approach. I think that the sub-PRs would be:
1. updates to the finance models (if they're needed?)
2. update to resource models (I think that theres a whole lot of different implementation questions that I'd want the group's ideas on)
3. update to performance model baseclass and cost model baseclass and maybe changes to 1 performance model and 1 cost model to showcase usage
4. updates to other performance and cost models

I haven't dug into the changes in the model baseclasses much, I think those methods would benefit from some inline comments to make it easier to follow the logic.
I also think that this PR has changes that are broader than just making models compatible to different simulation lengths, it seems like a lot of updates are based on time-step changes which, to me, is a different category. I also think that all performance and cost models should get an attribute like `_time_step_bounds` that is something like `_simulation_length_bounds` so we can better keep track of which models work with different simulation lengths

h2integrate/resource/resource_base.py
jaredthomas68: Resource year is coming from the config and just for the base year. The number of years needed and their values are determined from the dt and n_timesteps. I don't think we should require the user to put in every year. In the future I think we can have a string start and stop date/time format, but for now I lean towards keeping the single year spec.

I have made it not overwrite the base_year. Instead, resource year is determined in the loop and threaded through all the validators when loading multiple years. This created many small changes, but allows us to still reuse the single year loader code.

I also updated the logic slightly so that representative years can be handled (e.g. TMY) and when a representative year is specified, that same data is repeated for all requested simulation time. I thought about throwing an error, but I decided there may be other models or resources in the simulation that we do want to simulate and so I did not restrict the simulation length to match representative year data.

h2integrate/converters/solar/solar_pysam.py
jaredthomas68: Done

h2integrate/resource/resource_base.py
jaredthomas68: Fixed in https://github.com/NatLabRockies/H2Integrate/pull/887/commits/708d26802bf6522ce74593f9bf93a0713bcfb5ef

examples/24_solar_battery_grid/plant_config.yaml
```diff
@@ -2,14 +2,18 @@ name: plant_config
...
       solar_resource:
         resource_model: GOESAggregatedSolarAPI
         resource_parameters:
           resource_year: 2013
-          resource_filename: 30.6617_-101.7096_psmv3_60_2013.csv
+          resource_filename:
+            - 30.6617_-101.7096_psmv3_60_2013.csv
+            - 30.6617_-101.7096_psmv3_60_2013.csv
+            # - 30.6617_-101.7096_psmv3_60_2014.csv
+            # - 30.6617_-101.7096_psmv3_60_2014.csv
```
johnjasa, line 16: Intentional inclusion of the commented-out years? I'd suggest removing the addition of the 2014 .csv file in this PR if possible

h2integrate/core/model_baseclasses.py
johnjasa: Agreed on fully splitting out the SOH contributions outside of this PR. I also think we could separate the CF and replacement schedule computations since they have meaningfully different purposes.

johnjasa (changes requested):
## Top-level thoughts

Overall, this PR is moving in a useful direction and I'm glad we're finally crossing this bridge!
Thanks for this, @jaredthomas68.

This PR bundles several fairly large changes: annualization, multi-year resource handling, timestep definition expansion, model updates, and finance changes.
I think it would be easier to review and make robust as a sequence of smaller PRs.
I know @elenya-grant feels similarly, though I might have a slightly different suggested sequence for the sub-PRs.

I think the biggest decision points about how to subdivide these efforts revolve around project timeline needs; this will help set the order we tackle annualization, resource handling, model updates, finance changes, and lifecycle logic.
If you said "I absolutely need this before the end of the FY," then I vote you use the battery degradation branch for your project needs, but take more time and stepwise discussion with the team to ensure we're moving in the same decided direction.

## Suggested potential broken-out PR order

1. Define annualization behavior at a baseclass model and framework level

	Keep the basic annualization helper and validation changes together. Add tests for partial years, multi-year simulations, leap years, and non-dividing timesteps. We should also document what `tile`, `final_sim_value`, and `average_sim_value` mean for a partial-year run, making that behavior very clear for users.

2. Add resource loading and resampling (could be the first or second sub-PR)

	Heavily influenced by #888 and input from others. Keep the multi-year resource acquisition and time-tools changes together. Add at least one end-to-end test covering load, resample, concatenate, and consumption by a real performance model.

3. Enable models one at a time

	Some of these might need to happen with the sub-PRs above, but otherwise: Update grid/demand first, then solar, storage, and the other technologies. Each model should get focused tests for the timestep ranges we actually support. Finance changes should be handled here, but with special care: verify that annual production, utilization, and variable OpEx match independently calculated timeseries totals for more than just electricity. Once the underlying models are proven, update examples for the supported combinations.

4. Separate SOH and degradation lifecycle logic

	I think this can be split out relatively easily; this PR currently adds generic SOH-based degradation, EOL detection, replacement-cycle extrapolation, and replacement schedules. The initial non-annual work could use the existing no-degradation path and continue returning a zero replacement schedule.


## Some open questions worth discussing

I expect that by tackling these as sub-PRs, it'll be easier for folks to identify questions like these and provide higher quality feedback:

- Is the default sub-annual run meant to represent a typical slice that gets scaled to a year, or an actual seasonal slice?
- Are all resource values rates, or do some need different resampling rules? i.e. would they ever be totals or integrated quantities?
- Which models and timestep ranges are intended to be supported in the first version?
- Do we enforce that a model coming into H2I has certain allowable timestep or annualization behavior? 
- For the separate SOH follow-up, is replacement always recorded at plant-year boundaries if SOH reaches EOL mid-year?

examples/24_solar_battery_grid/plant_config.yaml
jaredthomas68: Good catch, I meant to have the 2013 and 2014 only, but was playing around with running 4 years so I doubled each. I'm inclined to keep the addition of the 2014 csv do clearly demo multi-year data, not just giving one year twice. I'll revert to the intended two years.

